PR #119, opened 2019-09-03
Minor null pointer fix

arcturus2:
### Description of the Change

A null pointer fix for when there is no type defined for a node.


arcturus2:
Thank you.

Issue #21, opened 2019-07-26
Basic support for Mac OS


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

Provide basic support for Macs. Constellation currently does not run/isn't usable on a Mac

### Steps to Reproduce

1. Unzip Constellation on a Mac
2. Start Constellation

**Expected behaviour:** Constellation should render a graph and the varies views as expected (like on a Windows machine)

**Actual behaviour:** Constellation does not render a graph due to unsupported hardware

**Reproduces how often:** 100%

arcturus2:
I've gotten hold of a Mac to try and progress with effort and with basic testing, Constellation does run, however there are some issues.

To get it to run I did the following:
* Downloaded the Linux version of Constellation v1.3.1 from https://constellation-app.com
* Unzip
* Deleted the jre folder
* Downloaded the MacOS JRE (with JFX) from [azul](https://www.azul.com/downloads/zulu-community/?version=java-8-lts&os=macos&architecture=x86-64-bit&package=jre-fx)
* Unzip the JRE and put that as the `jre` folder
* Ran `bin\constellation` from a terminal

The issues I've found are:
* Selecting on the graph actually makes the clicks at an offset location on the graph. This may be linked to #387 but I've tested this on NetBeans 8 with Java 8 and NetBeans 11 with Java 11 and they both have the problem.
* Opening up different views (from the Views menu) work though there seem to be graphics issues with some parts of the view not rendering and it takes a bit of fiddling to get them to show (by minimising and maximising them a few times)

arcturus2:
Thanks for contributions from @hydra41 and @TheTimurid I'm going to say that we've achieved basic support for MacOS. The fixes are in https://github.com/constellation-app/constellation/releases/tag/v2.0.0-beta3.

PR #120, opened 2019-09-09
Changed texture2D to texture as the new method works on both MacOS an…

GCHQDeveloper601:
…d Windows/Linux

### Description of the Change
Changed two files to use another method from OpenGL, this method is valid for both Windows and MacOS OpenGL libraries. 

This change will enable further development for the Mac platform

### Benefits
Constellation will open on a Mac

### Possible Drawbacks

None

### Verification Process
I have tested this on a Mac, Windows and Linux (CentOS and Ubuntu) platform. The software works as before on Windows and Linux. 

### Applicable Issues

On Mac there is an issue where the Graph interaction is offset from where the mouse is.

### DIFF
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/shaders/Graph.fs
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/shaders/Graph.fs
@@ -6,5 +6,5 @@ uniform sampler2D graphTexture;
 out vec4 fragColor;
 
 void main(void) {
-    fragColor = texture2D(graphTexture, textureCoordinate) /*vec4(0.0, 1.0, 1.0, 1.0)*/;
+    fragColor = texture(graphTexture, textureCoordinate) /*vec4(0.0, 1.0, 1.0, 1.0)*/;
 }
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/shaders/Graph.vs
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/shaders/Graph.vs
@@ -9,5 +9,5 @@ uniform sampler2D depthTexture;
 
 void main(void) {
     textureCoordinate = inputTextureCoordinates.xy;
-    gl_Position = position; /*vec4(position.xy, texture2D(graphTexture, textureCoordinate));*/
+    gl_Position = position; /*vec4(position.xy, texture(graphTexture, textureCoordinate));*/
 }


arcturus2:
> there is no difference between texture2d and texture. Considering the fact that texture is the modern way of sampling a texture. Like @arcturus2 mentioned, MacOS is not in scope. What is the main motivation of updating from texture2d and texture for windows/Linux versions? Can you add some more justification on this change?

Hi @rsabhi. I will let @GCHQDeveloper601 respond but my understanding is that this change will enable Constellation to open a graph on a Mac which is a welcomed change to head towards resolving https://github.com/constellation-app/constellation/issues/21.

GCHQDeveloper601:
> > there is no difference between texture2d and texture. Considering the fact that texture is the modern way of sampling a texture. Like @arcturus2 mentioned, MacOS is not in scope. What is the main motivation of updating from texture2d and texture for windows/Linux versions? Can you add some more justification on this change?
> 
> Hi @rsabhi. I will let @GCHQDeveloper601 respond but my understanding is that this change will enable Constellation to open a graph on a Mac which is a welcomed change to head towards resolving #21.

Hi @rsabhi, as @arcturus2 says, this enabled Constellation to run on a Mac. This isn't a complete fix due to the offset so doesn't make it production ready for Mac's it does get over the Null pointer error that #21 shows.

rsabhi:
> > > there is no difference between texture2d and texture. Considering the fact that texture is the modern way of sampling a texture. Like @arcturus2 mentioned, MacOS is not in scope. What is the main motivation of updating from texture2d and texture for windows/Linux versions? Can you add some more justification on this change?
> > 
> > 
> > Hi @rsabhi. I will let @GCHQDeveloper601 respond but my understanding is that this change will enable Constellation to open a graph on a Mac which is a welcomed change to head towards resolving #21.
> 
> Hi @rsabhi, as @arcturus2 says, this enabled Constellation to run on a Mac. This isn't a complete fix due to the offset so doesn't make it production ready for Mac's it does get over the Null pointer error that #21 shows.

@GCHQDeveloper601, First of all your suggestion and change request are awesome. Really appreciated your effort to put forward. 
I tested in Mac Mojave, initial startup issue is gone. Only DNA graph works, None of the other graphs works in my testing. 

Just for the record, Mac configuration
Model Name:	MacBook Pro
  Model Identifier:	MacBookPro15,1
  Processor Name:	Intel Core i7
  Processor Speed:	2.6 GHz
  Number of Processors:	1
  Total Number of Cores:	6
Radeon Pro 555X 4 GB
Intel UHD Graphics 630 1536 MB

Approved your PR. Btw, there are 2 other places the same changes required and a PR has been created #121

PR #122, opened 2019-09-14
Cleanup Ivy dependencies

arcturus2:
### Description of the Change

* Upgraded Ivy to the latest version and use some new features like the ability to check if there is a newer version of a dependency
* Added an ant task to generate a html report of dependencies to get a better idea how all of the dependencies fit together
* Enabled sync during an Ivy retrieve command which will ensure that only the files retrieved remain, i.e. it removes files no longer relevant
* Remove JOGL solaris files as we are not supporting Solaris
* Remove `gt-epsg-hsql` as it is not used
* Remove `junit` as we are using TestNG
* Reviewed the "To be removed" section after reviewing it
* Include a version number to the Ivy jar to make sure the latest version is downloaded


CoreDependencies/src/ivy.xml
```diff
@@ -47,7 +41,7 @@
         <dependency org="org.geotools" name="gt-referencing" rev="21.2"/>
         <!--<dependency org="org.geotools" name="gt-render" rev="21.2"/>-->
         <dependency org="org.geotools" name="gt-xml" rev="21.2"/>
-        <dependency org="org.geotools" name="gt-epsg-hsql" rev="21.2"/>
+        <!--<dependency org="org.geotools" name="gt-epsg-hsql" rev="21.2"/>-->
```
rsabhi: this commented line can be removed. Or add an inline comment for the reason we need to have this line here.

build.xml
```diff
@@ -192,4 +190,21 @@
...
+        <ivy:resolve type="${ivy.resolve.types}"/>
+        <ivy:report todir="${suite.build.dir}/ivy-reports" graph="false" xml="false"/>
+    </target>
+    
+    <target name="check-latest-version" depends="-load-ivy-tasks" description="Check whether there is a new release of a dependency.">
+        <ivy:resolve/>
+        <ivy:checkdepsupdate showTransitive="false" revisionToCheck="latest.release"/>
+    </target>
+    
+    <target name="dependency-tree" depends="-load-ivy-tasks" description="Output the dependencies ina tree layout to the console.">
```
rsabhi: typo "ina" change to "in a"

CoreDependencies/src/ivy.xml
arcturus2: @rsabhi I think the approach used with the `org.geotools` section is to list all of the libraries `org.geotools` supports and comment the libraries we are not using. Given `gt-epsg-hsql` is not used based on my investigation I have commented it but left it there so that should someone want to include it they can. Ditto for `gt-coverage` and `gt-render`.

build.xml
arcturus2: Thank you @rsabhi. I've corrected it, please check.

CoreDependencies/src/ivy.xml
rsabhi: My recommended approach here is to remove all the commented unused dependencies.

arcturus2: Understood @rsabhi and this is what I would normally follow as well. However the `org.geotools` is a bit special and I've added a comment block to explain it (see https://github.com/constellation-app/constellation/pull/122/commits/fd9b0f2c611513a53e675c04931092c88e0bcaa7).

Issue #139, opened 2019-10-08
Can't add custom blazes with specified vertex ids from the Python client


algol60:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues

### Description

Calling `cc.run_plugin('addcustomblaze')` with specified vertex ids causes an exception.

### Steps to Reproduce
```
import constellation_client
cc = constellation_client.Constellation()
df = cc.get_dataframe(vx=True, attrs=['source.[id]'])
ids = list(df['source.[id]'])
cc.run_plugin(
    'addcustomblaze',
    {'BlazeUtilities.color':'#ffc828', 'BlazeUtilities.vertex_ids':ids}
)
```

**Expected behaviour:** Blazes are added to all of the vertices.

**Actual behaviour:** An exception is thrown.

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.
```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
	at au.gov.asd.tac.constellation.functionality.blaze.BlazePluginUtilities.verticesParam(BlazePluginUtilities.java:57)
	at au.gov.asd.tac.constellation.functionality.blaze.AddCustomBlazePlugin.edit(AddCustomBlazePlugin.java:87)
	at au.gov.asd.tac.constellation.pluginframework.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:204)
etc...
```

Issue #133, opened 2019-09-30
Column types are incorrectly inferred in constellation_client.get_dataframe()


algol60:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues

### Description

When reading string values that look like numbers, the column types are being incorrectly inferred in `constellation_client.get_dataframe()`.

### Steps to Reproduce

Create a new analytic graph. Manually add three nodes with the following identifiers:
* "1.0"
* "2"
* "12345678901234"

Create a transaction from node "2" to node "12345678901234".

Start the CONSTELLATION REST server using Tools -> Start REST Server.

In a Jupyter session (notebook or console), execute the following code.

```
import constellation_client

cc = constellation_client.Constellation()

df = cc.get_dataframe(attrs=['source.Identifier', 'destination.Identifier'])
print(df.dtypes)
print(df)
```

The result is:

```
source.Identifier           int64
destination.Identifier    float64
dtype: object
   source.Identifier  destination.Identifier
0                  1                     NaN
1                  2            1.234568e+13
```

**Expected behaviour:** The Identifier types should be string (object), not int or float.

**Actual behaviour:** `pandas.read_json()` is inferring ints and floats because the values look like ints and floats.

**Reproduces how often:** 100%

arcturus2:
Thank you @algol60. I've merged your PR so I'll close the issue.

PR #140, opened 2019-10-08
:bug: Allow vertex ids to be strings.

algol60:
### Description of the Change

Issue #133 caused vertex ids to be strings in a dataframe. (Previously they were inferred as integers.) Using the vertex ids as-is to call the addcustomblaze plugin causes the plugin to fail because it wsa incorrectly expecting integer ids rather than strings.

This fix allows the addblaze and addcustomblaze plugins to accept string ids (and integer ids for backwards compatibility).

### Verification Process

Re-ran the steps to reproduce in issue #139.

### Applicable Issues

#139

### DIFF
--- a/CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/blaze/BlazePluginUtilities.java
+++ b/CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/blaze/BlazePluginUtilities.java
@@ -27,19 +27,34 @@ import java.util.List;
 public class BlazePluginUtilities {
 
     /**
+     * Convert a String or Integer to an int.
+     * <p>
+     * The vertex ids should be strings, but we need to accept integers
+     * for backwards compatibility.
+     *
+     * @param o A String or Integer.
+     *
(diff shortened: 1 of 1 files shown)


CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/blaze/BlazePluginUtilities.java
```diff
@@ -48,12 +63,11 @@ static BitSet verticesParam(final PluginParameters parameters) {
         } else if (vParam.getClass() == BitSet.class) {
             vertices = (BitSet) vParam;
         } else {
-            final List<Integer> vertexList = (List<Integer>) vParam;
+            final List<Object> vertexList = (List<Object>) vParam;
```
arcturus2: @algol60 was the reason you changed this to `Object` because `toInt()` expects an Object?

arcturus2: I answered my own question...going to merge this. Thank you again @algol60.

Issue #146, opened 2019-10-18
Return relevant error messages to REST clients


algol60:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues

### Description

When an exception is thrown during a REST operation, the error is not consistently returned to the client.

### Steps to Reproduce

After creating a graph and starting the REST server:

```
import constellation_client
cc = constellation_client.Constellation()
df = cc.get_dataframe(attrs=['nonexistent-attribute'])
```
**Expected behaviour:** The Python client should throw an exception containing a relevant error message.

**Actual behaviour:** The Python client throws an exception containing a generic "server error" message.

**Reproduces how often:** 100%

PR #147, opened 2019-10-18
:bug: Correctly return error messages to REST clients.

algol60:
(Note: this PR should be done in sync with updated jars.)

### Description of the Change

The Python requests client expects the first line of an HTTP response
message to be of the form

Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF

as per RFC 2616 section 6.1. However, there is nothing in the servlet
code to properly set the reason phrase. This commit catches exceptions
in the called code and correctly sets the response using
response.sendError().

However, the call to response.SendError sort of works: it sets the
status and status phrase, but then errors with
"java.lang.NoSuchMethodError: javax.servlet.http.HttpServletResponse.getStatus()I".
This indicates mismatched jars: see for example:
https://stackoverflow.com/questions/44730096/javax-servlet-http-httpservletresponse-getstatus-not-found .
Some work-around code catches the exception and continues.
When the jars are sorted out, the extra code can be removed.

The clash appears to be between `javax.servlet-api-3.1.0.jar` (the one we want) and `javaee-api-5.0-2.jar` (the incompatible one).

### Benefits

Informative error messages for REST clients.

### Possible Drawbacks

None. The code path for calls that don't throw an exception is unchanged.

### Verification Process

Run the reproducer code in issue #146 and see a relevant error message in the exception.

### Applicable Issues

#146


CoreWebServer/src/au/gov/asd/tac/constellation/webserver/api/ConstellationApiServlet.java
```diff
@@ -25,11 +27,21 @@
 /**
  * A web service which makes up part of the Constellation REST API. The use of a
  * secret is enforced for this type of web service.
+ * <p>
+ * Any exceptions thrown while executing a servlet call are caught and converted
+ * to an HttpServletResponse.sendError() response, as well as being logged
+ * at Level.INFO (to avoid an error dialog box being displayed).
```
arcturus2: @algol60 I'm going to suggest we log at to `Logger.ERROR` instead. It will still appear on the output console but will be prefixed with `[ERROR]`. I'm going to merge your PR and send through a follow up PR with this minor suggestion rather than hold up your PR any longer.

Issue #98, opened 2019-08-19
Delimited File Importer 'Row' column cannot be used


cygnus-x-1:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

The auto-generated 'Row' column in the file preview pane cannot be assigned an attribute (this worked at one point in the past).

### Steps to Reproduce

1. Open the Delimited File Importer tool and choose a file.
2. Drag the source.Identifier attribute onto the 'Row' column.
3. Click the Import button.

**Expected behaviour:** [What you expect to happen]

There should be n nodes on your graph where n is the number of rows in your delimited file.

**Actual behaviour:** [What actually happens]

There is one node on your graph, as the 'Row' column was ignored and so all nodes were collapsed into one due to the missing primary key values.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100% of the time

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

serpens24:
The issue here is the **setValue** function in the class **ImportAttributeDefinition**.

this function has an if else block testing the member variable **columnIndex**.
The value of **columnIndex** for attributes tied to columns in the source data run from 0-x where x is the total columns -1.
The value of **columnIndex** for attributes not tied to columns is ATTRIBUTE_NOT_ASSIGNED_TO_COLUMN.
However if the user selected the **Row** column in the importer, this is a generated column and doesn't exist in source data, it returns a **columnIndex** value of -1. As can be seen, this value falls through the ifs and hence a null value is returned as the column value.

To fix, modify the **setValue** function to take a fourth parameter **rowIndex** which is passed the index of the row. An else condition can then be added to use this value. Note this else condition will test for -1, hence if by some means a value other than -1, ATTRIBUTE_NOT_ASSIGNED_TO_COLUMN, or the value of a real column in the importer, then a null would be returned.

This function is called 4 times, all in the **processVertices** or **processTransactions** methods within **ImportDelimitedPlugin**. These are all sitting in loops running through the rows of data, so its easy to determine the row index.

[...]

PR #161, opened 2019-12-13
Added rowIndex as 4th parameter to ImportAttributeDefinition.setValue().

serpens24:
Use rowIndex as attribute value if attributes selected column = -1.
Update all calls to ImportAttributeDefinition.setValue() to use rowIndex.

### Description of the Change
- Update ImportAttributeDefinition.setValue() to accept a rowIndex parameter. This parameter will be used in cases where an import uses the **Row** column, as this column has index -1 and is not currently handled by logic. We need to pass this index in as the supplied parameters do not provide enough information to determne it.
- The 4 existing calls to ImportAttributeDefinition.setValue() are all made from loops iterating over rows and creating attributes, as such, modify the calls to also send the row index.

### Alternate Designs
Considered passing in value as extra columns of the row parameter, but this required a little extra refactoring and would be less efficient performance wise.

Bug fix of existing core code.

### Benefits

Ability to map attributes to the row index of source csv data.

### Possible Drawbacks

N/A

### Verification Process

Performed a series of imports of new CSV data, focussed on cases where we use the Row column as either:
a) the identifier
b) a value other than identifier
In each case confirm that the correct attribute is updated in attribute editor and visually on grapg.

In terms of verifying nothing has been broken, the change was made in code that was only executed in error condition, so both the above tests, and visual inspectionw ere used.

### Applicable Issues

N/A

### DIFF
--- a/CoreImportExport/src/au/gov/asd/tac/constellation/importexport/delimited/ImportAttributeDefinition.java
+++ b/CoreImportExport/src/au/gov/asd/tac/constellation/importexport/delimited/ImportAttributeDefinition.java
@@ -37,6 +37,7 @@ public class ImportAttributeDefinition {
     private final PluginParameters parameters;
 
     private static int ATTRIBUTE_NOT_DEFINED = -93459;
+    private static int ROWID_COLUMN_INDEX = -1;
 
     /**
      * Immutable attributes are defined in {@code ImportController} as mockup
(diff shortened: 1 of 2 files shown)


arcturus2:
@rsabhi how are you?

Have a read of #98 if you haven't seen it. Having the row of the excel as a property on the graph is an easy way to cross check which row produced that part of the graph. Actually we can achieve this by adding a column to the excel and increment the numbers but as @cygnus-x-1 mentioned, this used to work at one point and saves time by not having to manually edit the Excel file; it's there so why not use it!

serpens24:
Apologies - the reference to performance was referring to an alternate design that I discounted.

arcturus2:
@rsabhi I'm going to go ahead and merge this in because I feel that @serpens24 and myself have clarified your questions. If not then I apologies and happy to discuss further.

PR #168, opened 2019-12-16
DAV graph creation issue#158

aldebaran30701:
### Description of the Change
The change handles the case where a user opens constellation, and launches the Data Access View (DAV). The Execute button (GO) is disabled. The change allows the user to select a plugin to run, and a default graph be launched upon selection. this allows the user to still utilise Constellation to its full effect, as well as add significant UX improvement. This is more effective than showing the user an error message and them have to manually correct the issue.
The code checks if there is a graph currently open, if a plugin is selected and if that plugin's parameters are valid. Once a graph is not present, but all other conditions are met, a new default graph is launched, allowing the user to proceed to click the go button and run the plugin.

### Alternate Designs
Add checking to the update method within DataAccessPane, where the execute button is enabled. This was not suffice as it was called upon multiple times during execution. 
The proposed solution makes the most sence as it is handled within the same section as other checks for graph status are made.

This enhancement allows the user to navigate the detailed plugin menu (DAV) more efficiently and user friendly. It avoids error states being presented to the user as none are entered or introduced.

### Benefits
A more intuitive UI is presented to the user, with the Execute button being enabled at almost all times when a graph plugin can be ran. The user will not run into an error state that will cause confusion during use.

### Possible Drawbacks
Slight delay caused by creating a graph on selection of a plugin. This can be eventually sped up by increasing the performance of graph launching in the future.

### Verification Process
- Ran all test suites provided
- Multiple runs of Constellation to ensure no precondition occurs that will error the introduced code
- Clicked DAV, clicked each single plugin, clicked run = runs the plugin
- Clicked DAV, clicked multiple plugins, clicked run = runs the plugin
- Clicked DAV, clicked single plugin, clicked new graph, clicked run = runs the plugin
- Clicked DAV, clicked single plugin, clicked new graph, clicked close graph, clicked extra single plugin, clicked run = runs the plugin

### Applicable Issues
N/A

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
@@ -17,6 +17,7 @@ package au.gov.asd.tac.constellation.views.dataaccess.panes;
 
 import au.gov.asd.tac.constellation.graph.Graph;
 import au.gov.asd.tac.constellation.graph.manager.GraphManager;
+import au.gov.asd.tac.constellation.graph.node.NewDefaultSchemaGraphAction;
 import au.gov.asd.tac.constellation.pluginframework.Plugin;
 import au.gov.asd.tac.constellation.pluginframework.PluginException;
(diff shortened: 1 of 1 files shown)


arcturus2:
Thanks for this @aldebaran30701

Issue #26, opened 2019-07-26
Setting a choice in updateParameters causes the DAV plugin to be enabled


arcturus2:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

Setting a choice in `updateParameters` in a Data Access View (DAV) plugin will fire an event and cause the plugin to be enable. For example refer to `ExtractFromContent` or `SplitNodesPlugin` which have this issue and so for now, the `setChoice` logic has been commented out.

### Steps to Reproduce

1. Open the Data Access View
2. Open a graph

**Expected behaviour:** No plugin is active

**Actual behaviour:** The Extract from Content plugin or Split Nodes plugin is enabled in the DAV

**Reproduces how often:** 90%

### Additional Information

The issue currently does not occur because the relevant code is commented.

aldebaran30701:
Does this issue relate to the following state?

1. Open Constellation
2. Open Data Access View
3. Open a graph
4. Edit a parameter from within any plugin
The above leads to the plugin becoming selected. My question is, is this the behaviour described from the issue?

Uncommenting the relevant code and testing reveals that no odd behaviour occurs following the steps to reduce listed in the issue.
If they are the same issue, it seems that all plugins have this behaviour by default; not just the ones listed within the issue.

arcturus2:
Hey @aldebaran30701, the plugin being enabled when changing a parameter is the expected behaviour.

The issue was that with the mentioned lines uncommented, when you open a new graph, the plugin is enabled. Can you try uncommenting the lines in the SplitNodesPlugin and try to see if the issue occurs?

https://github.com/constellation-app/constellation/blob/7286c5bfa6685cd6cb5e1fdcdf8dafb13c4a5501/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java#L125-L128

In theory what would happen is the split nodes plugin becomes enabled when you open a new graph.

aldebaran30701:
I did manage to get this event to occur, and as a side note, it did not seem to occur often.
[...]

PR #174, opened 2019-12-19
Add methods to Plugin classes fix issue #26

aldebaran30701:
- PluginParameter setPropertyWithoutEvent
 - SingleChoiceParameterType setOptionsWithoutEvent
 - usage added within plugins
   - ExtractWordsFromText
   - SplitNodesPlugin
Implementation should fix issue #26

### Description of the Change
This change adds methods to assist in the management and configuration of plugins in the DAV. The methods allow the plugins to set parameters and options which may depend on the type of graph opened and within focus. this enables the plugin to configure it's options without enabling itself through a changeEvent triggering the plugin to activate.
The method name has _WithoutEvent concatenated onto it which allows for ease of readability and less confusion of an overloaded method. the method uses the same logic from the original method to ensure consistency in it's operation.
The methods in this PR have been implemented within the following plugins which had issues relating to #26. 

### Alternate Designs
Overloaded method signatures with a boolean flag left the same amount of code in each file, but reduced readability. The original idea was to tackle the issue within setChoice which was commented out previously. This was a two part issue, which both appear to be solved with this fix.

This should be in core as it defines the "expected" behaviour of loading a graph when the DAV is opened. It solves issue #26 and allows for future expansion with any other plugin which may need to configure settings dependent on the graph loaded.

### Benefits
This fix as mentioned before, allows for future plugin development and also enables the user to create a graph without having to deselect the automatically enabled plugins.

### Possible Drawbacks
None realized thus far.

### Verification Process
Run all tests in suite
Open Constellation
 - Open DAV -> open graph = no plugin automatically activated.
- Open DAV -> open graph -> open SplitNodesPlugin -> select an option from dropdown -> deselect enabled checkbox -> close graph -> open new graph = no plugin automatically activated.
- Open DAV -> open graph -> draw two nodes -> change identifiers -> split by identifier -> close graph -> deselect plugin -> new graph = no plugin automatically activated.
- Open DAV ->open graph -> open both SplitNodesPlugin and ExtractWordsFromTextPlugin and select parameters -> deselect both plugins ->close graph -> new graph = no plugin automatically activated.
- Open DAV ->open graph -> close graph -> close DAV -> open DAV -> open graph = no plugin automatically activated.

### Applicable Issues
N/A


arcturus2:
@aldebaran30701 this works and I'm happy with it.

@cygnus-x-1 if you have time, I'd like your feedback on this.

cygnus-x-1:
This might work for the two plugins identified, however there are several other methods in SingleChoiceParameterType and MultiChoiceParameterType (and maybe other ParameterType classes) which fire change events and do have not been provided an equivalent fix. I think a proper solution to this should be built into the framework, not just this SingleChoiceParameterType.

arcturus2:
Thanks @cygnus-x-1, your feedback is always valued.

@aldebaran30701 would you be able to think about this a bit more?

I've been trying to think about a viable solution and I'm wondering if something like this is a good idea? It would be to suppress the event firing by introducing a `suppressEvent()` method to `PluginParameter`? The `fireChangeEvent()` is part of `PluginParameter` so it will be easy to not fire.

Then, you can use it like this in the case of `updateParameters()` and add some docs to `Plugin.updateParameters()` to suggest it would be useful to use `PluginParameter.suppressEvent()` and explain the reason.

```java
transactionType.suppressEvent(true);
if (types.contains("Correlation")) {
    SingleChoiceParameterType.setChoice(transactionType, "Correlation");
}
transactionType.suppressEvent(false);
```

aldebaran30701:
I really like that idea @arcturus2 
Allows for consistent practice throughout the whole code base if needed for other items firing events.
I can start working on something today, following along the lines of what you have outlined above.
@cygnus-x-1 do you for see any issues with this implementation?

arcturus2:
Good thinking @aldebaran30701. Can we have both method signatures? 😃 
If you don't pass anything then it suppresses everything, otherwise you can specify your list.

aldebaran30701:
done and done. 
I've updated the changelog, added documentation to plugin.updateparameters().
used a collection to hold the currently "filtered" pluginparameters. 
used a boolean flag to check whether or not to filter a result. 
I have one other improvement I would like to make, that would be to refactor the checking of the events, into the fire change event itself.

aldebaran30701:
Fixed. Much neater.

CorePluginFramework/src/au/gov/asd/tac/constellation/pluginframework/parameters/PluginParameter.java
```diff
@@ -70,6 +71,11 @@
     private boolean visible = true;
     private boolean enabled = true;
     private String helpID;
+    private boolean isSuppressed = false;
+    
+    private ArrayList<ParameterChange> suppressedEvents = new ArrayList<>();
+    
+    private static final ArrayList<ParameterChange> PARAMETER_EVENTS = new ArrayList<>(Arrays.asList(ParameterChange.values()));
```
arcturus2: Ditto, can you please change to `private static final List`. It doesn't make too much difference, its more my OCD 😸 .

CorePluginFramework/src/au/gov/asd/tac/constellation/pluginframework/parameters/types/SingleChoiceParameterType.java
```diff
@@ -170,6 +170,7 @@ public static ParameterValue getChoiceData(final PluginParameter<SingleChoicePar
 
     /**
      * Set the current selected choice from a string.
+     * This will fire a {@link ParameterChange} event if successful.
```
arcturus2: If you are going to add this to `SingleChoiceParameterType` can you add it to the other *ParameterType classes for consistency? Otherwise removing it is also OK.

CorePluginFramework/src/au/gov/asd/tac/constellation/pluginframework/parameters/PluginParameter.java
arcturus2: Do you think renaming it to `ALL_PARAMETER_EVENTS` makes for better understanding?

CorePluginFramework/src/au/gov/asd/tac/constellation/pluginframework/parameters/PluginParameter.java
```diff
@@ -150,6 +156,38 @@ public void setVisible(boolean visible) {
...
+     * or disengage them. 
+     * NOTE: passing suppress as false will enable all events.
+     * 
+     * @param suppress if true, the listed events will not fire. if no events are listed, 
+     * all events will not fire. If false, all events are enabled.
+     * @param eventsToSuppress the events to suppress. pass an empty {@link ArrayList}
+     * when specifying all events or enabling events. Pass {@link ParameterChange} 
+     * enumerated types when specifying certain events to suppress.
+     */
+    public void suppressEvent(final boolean suppress, final ArrayList<ParameterChange> eventsToSuppress){
```
arcturus2: Please change the parameter type to `List`.

I might be alone in thinking that it feels a little weird to specify an empty list to suppress everything but passing a populated list to ignore specific ones. Perhaps we could have another method being `public void suppressEvent(final boolean suppress)` which will just call this `suppressEvent` with the empty list.

Happy to be told otherwise.

aldebaran30701: passing a populated list should suppress only the ParameterChange events listed within it.
I've done the other changes and pushed them, but yet to change the suppressEvent(...)

arcturus2: > passing a populated list should suppress only the ParameterChange events listed within it.

Yep, so if we go with that logic then passing an empty list should suppress nothing right? But it suppresses everything. I'm OK with this though because there is javadoc explaining it. My only minor suggestion is that we could have another method signature of `public void suppressEvent(final boolean suppress)` which will hide the detail of needing to pass the empty list.

Issue #176, opened 2019-12-19
Problem with dependencies not found


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Starting to see file not found exceptions appearing in master. Note sure what's caused this or how I've missed this. I'll work on a proper solution next year!

```
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\imageio-ext-geocore-1.2.2.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\imageio-ext-utilities-1.2.2.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\imageio-ext-streams-1.2.2.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\jai_codec-1.1.3.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\gluegen-rt-main-2.3.2.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\jaxb-runtime-2.4.0-b180830.0438.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\FastInfoset-1.2.15.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\javax.activation-api-1.2.0.jar (The system cannot find the file specified)
java.io.FileNotFoundException: D:\Users\arcturus\code\constellation\build\cluster\modules\ext\imageio-ext-tiff-1.2.2.jar (The system cannot find the file specified)
[...]

PR #177, opened 2019-12-19
Fix broken dependencies #176

arcturus2:
:bug: Fix broken dependencies
* Not sure why they were added and no longer downloading from maven. The short answer is that it works without them so removing them.
* The long term solution is for the project.properties and project.xml file to update when dependencies change and there is a solution to fix that (next year)/

### Description of the Change

Refer to #176 for the background. The immediate fix is to remove these unused jars and work on a solution to update the `project.xml` and `project.properties` files when dependencies change. This will be worked on separately in the new year.

### Benefits

Fix startup errors showing the `FileNotFoundException`.

### Verification Process

I've tested the application opening different views and running plugins and things seem to work so the problem jars don't seem to be used.


cygnus-x-1:
Oh cool - this was getting annoying.

PR #182, opened 2019-12-25
Restore the Flights.csv sample file

arcturus2:
### Description of the Change

:bug: Restore the Flights.csv sample file
* The sample Flights.csv file that can be used for importing was empty in master. The file has been restored using the file from the constellation-training repository.


cygnus-x-1:
Why is this in the main project?

arcturus2:
> Why is this in the main project?

A great question. Based on feedback from the initial Data61 UX review, it was identified that it will be really useful to step someone through how to import data via the CSV importer and having a sample file they can play with. The steps to import the CSV was going to be put in the help section of the Welcome tab. This is still a WIP so happy to get your thoughts on this too.

Perhaps I can hold off on this PR until the Welcome tab is done...perhaps maybe delete this CSV file for now given its an empty size anyway!

Issue #164, opened 2019-12-16
The 'Split Nodes' plugin should copy all attribute values


cygnus-x-1:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The 'Split Nodes' plugin (available through the Data Access View) allows you to break a node into multiple nodes by splitting on its 'Identifier' attribute. Currently, any new nodes created are created only with an 'Identifier' attribute value (all other attributes take their respective default values). I propose we change this behaviour so that any new nodes created instead copy all attribute values of the node they are spawning from. This makes much more sense for cases where you are cleaning up the graph, for example you might import the CC field from an email giving you a node such as 'person_one@company.org person_two@company.org' and want to split that into two separate email addresses. It makes less sense for certain other cases, for example you might import a URL such as 'https://abc.company.org/awesome_app:8888' and want to break off the port as a separate node. In these cases, however, the behaviour is no worse that what currently happens as manual edits to attribute values will still be required.

### Steps to Reproduce

1. Create a node.
2. Set its 'Identifier' attribute value to 'person_one@company.org person_two@company.org'.
3. Set its 'Type' attribute value to 'Email Address'.
4. Select the node and run the 'Split Nodes' plugin, splitting on a single space.

**Expected behaviour:** [What you expect to happen]
Two nodes now exist on the graph, 'person_one@company.org' and 'person_two@company.org', and both are of type 'Email Address'.

**Actual behaviour:** [What actually happens]
Two nodes now exist on the graph, 'person_one@company.org' which has type 'Email Address' and 'person_two@company.org', which has type 'Unknown'.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%.

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

PR #185, opened 2020-01-06
The 'split nodes' plugin should copy all attribute values #164

formalhaut69:
### Description of the Change
This change solves issue #164 by setting the new node's Type to be the same as the source node. In doing this, the new node copies the attribute values from the source as well. 

Provides an enhancement for the split nodes plugin.

### Benefits
Enhances the split node plugin's functionality to enable it to set the copy all of the original node's attributes to the new node. 

### Possible Drawbacks
Whilst the solution copies over custom attributes to the new node, it does not copy the values of the custom attribute. I am unsure if this behavior is preferred. 

### Verification Process
Followed the steps outlined in the issue description to ensure that the expected outcome is met.


arcturus2:
@formalhaut69 thanks for the PR and apologies for the delay to get back to you.

I've done some testing and I can confirm that it works for the expected example of making sure the type is copied for the 2 emails 'person_one@company.org person_two@company.org' but I believe the idea behind this issue was to copy all attributes of the node being split.

Are you able to have another go at trying to fulfil this? I am happy to merge this PR as it is a step forward but issue #164 will have to remain open.

arcturus2:
@formalhaut69 i'll add that if you want a hand with making this more generic just let me know, otherwise I'll leave it with you.

formalhaut69:
@arcturus2 yep, I can take another look at this but just to confirm,

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -152,6 +152,25 @@ private void editResultStore(final RecordStore result, final String left, final
             result.set(GraphRecordStoreUtilities.SOURCE + AnalyticConcept.VertexAttribute.TYPE, leftVertexTypesMatches.get(0));
         }
         result.set(GraphRecordStoreUtilities.DESTINATION + VisualConcept.VertexAttribute.IDENTIFIER, right);
+        
+        //Loops through all of the Node attributes and copies them to the new node
+        query.reset();
+        while(query.next()){
+            for (final String key : query.keys()){
+                if(SOURCE_ID.replace("<string>", "").equals(key) || SOURCE_IDENTIFIER.replace("<string>", "").equals(key)) {
```
arcturus2: @formalhaut69 thanks for this.

@cygnus-x-1 I wanted to ask if you have any ideas about how we could format this a bit better so that we don't have to be so verbose and replace the type `<string>` to get this to work.

We can replace `query.keys()` with `((GraphRecordStore) query).keysWithType()` which will work for line 160 but then the if block (162-164) comparing x, y and z will not work because the key is now `x<float>`.

To me something looks wrong with the design of the `GraphRecordStore`. Keen for feedback from anyone.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -152,6 +152,25 @@ private void editResultStore(final RecordStore result, final String left, final
...
+                     //Skips the id and Identifier to make the new node unique
+                } else if ((GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.X).equals(key) 
+                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Y).equals(key) 
+                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Z).equals(key)){ 
+                    //The coordinates are also skipped so that the second node is not created in the exact same location
+                    //as the first node
+                } else {
+                     result.set(GraphRecordStoreUtilities.DESTINATION + key.replace(GraphRecordStoreUtilities.SOURCE, ""),
```
arcturus2: A very minor comment but I would put this in 1 line because it is a little confusing seeing the `query.get` on its own.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
cygnus-x-1: You shouldn't need to worry about the field type, as the default behaviour of GraphRecordStore.keys is to explicitly **not** return field types.

formalhaut69: https://github.com/constellation-app/constellation/blob/28443a95682f6240cfca4d4792ca13000f3fdbf3/CoreGraph/src/au/gov/asd/tac/constellation/graph/processing/GraphRecordStoreUtilities.java#L59
@cygnus-x-1 @arcturus2 I've made a mistake here sorry, the SOURCE_IDENTIFIER never contains ```<string>``` so when comparing it to the key, we don't need to replace ```<string>```. The SOURCE_ID does still contain it's type so we still need to replace ```<string>``` for it. You can see in GraphRecordStoreUtilities it's declared with ```<string>``` in its value . The X,Y and Z values are declared without their type.

cygnus-x-1: While `GraphRecordStoreUtilites.ID` does explicitly declare a type (because it's designed to work with `GraphRecordStore` which requires typed fields), this type will only be recorded in the `typedRecords` object. `GraphRecordStore.keys` will return the keys of its `records` object (which has no types as it comes from `TabularRecordStore`) whereas `GraphRecordStore.keysWithType` will return keys from `typedRecords`. Since you are using `GraphRecordStore.keys` in this case, you shouldn't have to worry about types.

formalhaut69:
https://github.com/constellation-app/constellation/blob/f311947644f0bdc3fd3d82740c4c041aedb0d5fc/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java#L160
Thanks for your advice @cygnus-x-1 but I can't seem to find a cleaner way to reference the ID. I've slightly modified it so that we don't have to replace the ```<string>``` but it still doesn't seem like the best way to do this.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
arcturus2: Thanks @formalhaut69 for the fix.

Thanks @cygnus-x-1 for the explanation.

It looks like we are using features from the `TabularRecordStore` and `GraphRecordStore` causing issues. Though I'm not a fan of magic strings (i.e ".[id]"), I don't know how best to solve this one without going deep into the framework...and I don't feel its worth it right now given what @formalhaut69 has proposed works and I'm worrying about semantics.

cygnus-x-1: A simple fix might be to create ```TabularRecordStoreUtilities.ID``` as an analogue to ```GraphRecordStoreUtilities.ID``` and to use that instead.

Issue #170, opened 2019-12-17
Constellation freezes when opening DAV or CQV on a newly built sphere graph


formalhaut69:
### Prerequisites
[SupportPackage-klewis47-2019-12-18-09-52-52.zip](https://github.com/constellation-app/constellation/files/3975885/SupportPackage-klewis47-2019-12-18-09-52-52.zip)

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
The application freezes when opening the data access view or the quality control view on a newly built sphere graph. 

### Steps to Reproduce

1. Create a new analytic graph
2. Experimental > Build Graph > Sphere Graph Builder
3. Open the data access view or the quality control view

**Expected behaviour:**  The view should open properly

**Actual behaviour:** Constellation freezes

**Reproduces how often:** 100% 

### Additional Information

This bug doesn't occur if the view is already open in memory when you launch constellation. It also doesn't seem to occur if constellation has been running for a few minutes.

formalhaut69:
Issue also seems to occur if you follow the same steps but instead of building a new sphere graph you just create a new node.

arcturus2:
I remember this happening to me too but I wasn't able to reproduce it and it seemed to have "gone away".

One theory is that there is a race condition occuring which effects the `CountdownLatch` not working as intended.

https://github.com/constellation-app/constellation/blob/27c9d4091cc84f7918d32239d3c14ae2c9e93490/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java#L503

The latch not being released can explain why Constellation freezes (because that code is running on the graphics thread until `final Thread thread = new Thread("Data Access View: Load Plugins") {}` finishes.

Ironically, I think I moved the bulk of the code in `lookupPlugins()` to run on its own thread to improve performance.

If this issue can be reproduced reliably, I'd like my theory to be tested by setting a long timeout to see if it comes back after say 30 seconds.

```java
countDownLatch.await(30, TimeUnit.SECONDS);
```

Also interested if anyone else has ideas.

[...]

PR #188, opened 2020-01-07
Dav freeze issue#170

aldebaran30701:
### Description of the Change
The change adds a 1 second delay to the CountdownLatch which is used to avoid deadlock situations in the DAV and QCV.

### Alternate Designs
An alternate option is to dig deep into the code to find what other thread may be executing and accessing information at the same time to solve the issue.

This allows the user to actually open the DAV and QCV when a new graph was built.

### Benefits
As above

### Possible Drawbacks
Await call may be blocking another thread in the background. I cannot detect if there is any additional changes to the DAV or QCV because the button still appears to function as usual.

### Verification Process
Open multiple instances of Constellation to verify that the DAV and QCV still opens after a graph is created.
Cross check changes before and after to the QCV button.

### Applicable Issues
issue #170


CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/daemon/QualityControlAutoVetter.java
```diff
@@ -192,7 +193,7 @@ public void run() {
             thread.start();
 
             try {
-                countDownLatch.await();
+                countDownLatch.await(1, TimeUnit.SECONDS);
```
arcturus2: @aldebaran30701 apologies for the delay getting back to you on this PR.

I'd like to see if we can dig deeper into what the root cause might be following on from your findings from https://github.com/constellation-app/constellation/issues/170#issuecomment-570992326

I think I've introduced this bug so happy to look at assisting with a fix. I'm just having trouble getting reproducing this so I might have to introduce a sleep in one of the classes to force the freeze and maybe use that to see what other classes are getting blocked.

If you get to this before I do then that is also fine and welcomed 😄

aldebaran30701: #214 is now the most relevant Pull Request for this issue. Close this if necessary.

arcturus2: Thanks @aldebaran30701

PR #199, opened 2020-01-14
Handle K-Truss interactive state changes (Feature/issue#62)

serpens24:
# Description of the Change

1. Capture the cases where an open K-Truss panel is closed, If this panel was actively displaying interactive K-Truss visuals, then disable interactive mode on close (ensuring graph is redrawn without K-truss colours/shading).
2. If interactive is toggled off, ensure K-Truss colours are cleared. 

### Alternate Designs

nil

This addresses an issue described in description by which toggling K-truss interactive flag (or closing K-truss component) was leaving K-Truss visualisations on the graph.

### Benefits

See above, basically, if K-truss interactive is turned off then graph is not cluttered with K-truss information without context.

### Possible Drawbacks

Not aware of any.

### Verification Process

To verify toggling interactive doesn't leave dangling colours:
1. Open a graph with suitable clustering data for display.
2. Select Tools->Cluster->K-Truss.
3. Press the Cluster button. (Interactive is on by default)
4. Check Colour Nested K-Trusses.
5. Turn off interactive by pressing the Interactive button.
Confirm that colours are only shown if interactive is turned on.

To verify closing K-Truss form removes coliuring/shading:
1. Open a graph with suitable clustering data for display.
2. Select Tools->Cluster->K-Truss.
3. Press the Cluster button. (Interactive is on by default)
4. Close the K-Truss component (hitting X in tab header).
Confirm colours and shading are only shown if K-truss component displayed, and interactive is on.

### Applicable Issues

N/A


serpens24:
Based on @arcturus2 comments above, I have opted to make some further change. Currently the form is somewhat un-intuitive in that when the form is opened initially, the **Interactive** button is immeidately availabe - but doesn't respond to presses, additonally this button is marked as selected. This selected state means that as soon as the **Cluster** button is pressed to run the clustering that the interactive mode is treated as on.

Having spoken to @arcturus2 we beleive a better approach is only enable the **Interactive** button when it can be run, and to have it deslected when the form is initially opened.

This would mean when opening the form for the first time a user needs to run a cluster, at which point the **Interactive** button is enabled and can be pressed to toggle interactive mode. upon closing the form - interactive mode is turned off, and not turned back on until the form is re-opened and it is pressed again.

One thing to check for in all this is the button states are set correctly based on graph state, and this also needs to occur after a graph is saved and reopened. Ie if a graph was capable of displaying interactive data and then saved - if re-opened, this should be available.

serpens24:
@arcturus2 - this should be ready for you to take another look.

Issue #202, opened 2020-01-16
Empty new graph gets created even though graphs exist


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

A change was made recently to open a new graph if the Data Access View was open and a plugin was selected AND there was no graphs open. See https://github.com/constellation-app/constellation/issues/158.

I think this has introduced a side effect bug of creating new graphs when graphs are open. See the reproducer for more information.

### Steps to Reproduce

1. Open Constellation
2. Open a saved graph
3. Open the Data Access View
4. File -> New Graph (i.e analyticgraph1)
5. Select a Data Access plugin
6. Close the analyticgraph1

**Expected behaviour:**
Nothing should happen

**Actual behaviour:**
A new graph is opened, i.e analyticgraph2

**Reproduces how often:**
100%

### Additional Information

This one is starting to annoy users.

PR #206, opened 2020-01-16
Dav graph creation fix #158

### LINKED ISSUE #158: Data Access View - Green button greyed out without graph open
### Prerequisites

### Description

When I open the Data Access View, I can check some plugins to run but I am not able to click on Go. It took some time to figure out that a new graph has to be opened first.

This is not intuitive so some thought to improve the UX is required.

### Steps to Reproduce

1. Open the Data Access View
2. Select a plugin

**Expected behaviour:** The Go button should be green

**Actual behaviour:** The Go button is disabled because there is no open graph.

### Additional Information

Some suggestions:
* Can the Go button be enabled and if there is no graph opened then can the default graph be opened before the Data Access Plugin runs?

aldebaran30701:
### Description of the Change
Create a new graph only when the user selects the run button in the current state:
 - No graph open
 - At least one valid plugin selected (With correct parameters)
 - Run button has to be active

This fixes the issue #158 in light of the mishap with #202.
Change is now reflective of user running the plugin, as opposed to the previous selection of a plugin.

Additional null checks added for defensive programming

### Alternate Designs
Add graph creation onto selection event handler of plugin, this could have worked but may be annoying for users wanting to select a plugin without having a graph be created
Add graph creation to Execute button (implemented)

This fixes the issue #158

### Benefits
No extra graph created on close event of graph.

### Possible Drawbacks
None for seen

### Verification Process
Opening new graph - selecting (one, multiple and no plugins) and clicking run button
Opening saved graph - selecting (one, multiple, and no plugins) and clicking run button
Closing DAV with (graph open, closed) (Plugins selected, and deselected)
Opening DAV with graphs opened or closed, running plugins.
DAV opened with graph closed, multiple fast clicks on run button to try open multiple graphs

### Applicable Issues
#158


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
```diff
@@ -208,9 +209,32 @@ public DataAccessPane(DataAccessViewTopComponent topComponent) {
...
+                        selectedPluginsValid = false;
+                    }
+                } 
+            }
+            // when no graph present, create new graph
+            if(graphId == null){
+                if (pluginSelected && selectedPluginsValid) {
+                    NewDefaultSchemaGraphAction graphAction = new NewDefaultSchemaGraphAction();
+                    graphAction.actionPerformed(null);
+                    while(GraphManager.getDefault().getActiveGraph() == null){
```
arcturus2: Thanks for looking into this so quickly @aldebaran30701.

Just with this, given we are running this inside setOnAction() something to be mindful of is whether this is running on the JavaFX graphics thread. If so it could lock up the app.

I will do some testing and get back to you but I like the idea.

arcturus2: @aldebaran30701 I had a look at this and yes this is running on the FX thread and it does lock up the application. I don't think we can do anything in this case unless we have a progress circle saying that something is happening. I'll say that this is out of scope for this ticket though.

During testing I noticed that when I pressed go, the plugin I had set to run didn't run. I had to click Go again to make the plugin run. Have you seen this? This was happening every single time to me so it didn't work as expected. The test I did was add "www.google.com" to the "Extract Types from Text" plugin. I expected a node on the graph after it created the graph but it's empty.

Can you have a look when you get the chance @aldebaran30701?

aldebaran30701: I've just got around to checking this out. 
I notice what you mean about hanging, it appears frozen but then loads the graph. I am currently unsure how I could go about loading the graph without the delay. If I load it on another thread, the user still waits, only the UI will be usable. 

Great catch! The Graph State was null in the first call to Update(), which then initialized it. I added some initialization code just after the graph ID assigning so it will work on one click. this shouldn't effect load times.

arcturus2: Thanks @aldebaran30701 its working great now and very intuitive.

With regards to the delay I tried running some of the work in a different thread and it took longer lol so I think it's current behaviour is something we have to live with.

On web apps now that follow Material UI then have a small loading bar at the top. Something like that would be pretty cool I reckon.

PR #207, opened 2020-01-20
Add checks for null graph prior to refreshing graphs timezone. (feature/issue#193)

serpens24:
### Description of the Change

When a graph is closed with an open timeline view, calls are made to update the graph with timezone details - these calls result in null pointer exceptions once the graph has closed.
This change considers these cases and add robustness to the persistStateToGraph() and populatefromGraph() methods to ensure they check for a valid graph. If the graph is null, methods return gracefully without update.

### Alternate Designs

Nil

Addresses bug that occurs if graph is closed while timeline view is open (and a new timezone selected).

### Benefits

No exception raised anymore.

### Possible Drawbacks

Nil.

### Verification Process

Repeat **Steps to Reproduce** from parent issue, ie:
1. Make a new analytic graph
2. Open the timeline view
3. Change the timezone
4. Close the graph

### Applicable Issues

Nil.


serpens24:
Sorry, silly mistake on my part, deleted a change to updateTimeZone on commiting diff checks... a null check is needed on the line in question - however looking deeper, populateFromGraph is called from aheap of places and NONE of these places are performing the check.
So to deal with this added function populateFromGraphNode which just wraps populateFromGraph but first checks that graphNode is non null before extracting graph from it. Almost every call to populateFromGraph followedf the pattern populateFromGraph(graphNode.getGraph(), isFullRefresh); - so my new wrapper function validates graphNode can be accessed.

Issue #54, opened 2019-07-26
The Analytic View should read from its state attribute


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

The Analytic View 2.0 has a state attribute which it writes parameter information to, and saves to the graph. However it currently does not read from that state to restore that parameter information, but rather resets itself whenever graphs change.

### Steps to Reproduce

1. Make a new analytic graph
2. Open the Analytic View
3. Expand the Categories
4. Check the "Betweenness Centrality Analytic"
5. Open a new graph
6. Expand the Categories
7. Check the "Closeness Centrality Analytic"
8. Click on the first graph

**Expected behaviour:** 
The analytic view should show the "Betweenness Centrality Analytic" checked.

**Actual behaviour:**
The analytic view shows the first question selected instead

**Reproduces how often:** 
100%

aldebaran30701:
Am I correct in saying that below, reads a JNode, and obtains the stored parameter information from the graph? 
https://github.com/constellation-app/constellation/blob/c968184e866a76049d50df69b27b6304f7b739b5/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewStateIoProvider.java#L53-L98
I created a graph, selected and ran some AnalyticView plugins, saved the file and subsequently reloaded them with what appeared to be the correct state.

Could you provide steps to replicate this, or expand on your point regarding the resetting of parameter information and when it occurs. 

One other thing; what is Analytic View 2.0? just the current core Analytic View?

arcturus2:
@aldebaran30701 yep this is the code that looks after the state. Apologies for not providing a decent reproducer. I've just updated the ticket description if you want to have another read?

aldebaran30701:
At this stage, this is my understanding of the issue and how a working implementation would look like.
1. Having a graph opened, and launching the **Analytic View** should pre-select the plugins and parameters checked.  
[...]

PR #215, opened 2020-01-28
:bug: Analytic state attribute 54

aldebaran30701:
### Description of the Change
Implement state attribute usage for a currently opened graph.
Reads from the state to display saved graph plugins and parameters
Writes changes to the state attribute on selection or change of plugin

### Alternate Designs
Read from state as single plugin
Write to state as another plugin
This faced concurrency issues when selection was performed fast

It solves the bug of not keeping memory of plugins selected between graphs

### Benefits
Manages the state of plugins for the graph. Automatically saves the selections temporarily until complete graph save.

### Possible Drawbacks
when you open a file and change some values within plugin parameters of the Analytic view and then close that graph. The Analytic view shows the previously entered parameter values instead of a default set. The parameters get refreshed to default once the application restarts.

### Verification Process
Open graph -> open analytic view -> select plugins
-> select categories (maintained selection)
-> select questions (maintained selection)
-> deselect plugins (maintained selection)
-> save graph (maintained selection)
-> load same graph to verify state is maintained (maintained selection)
-> Open new graph
-> switch between graphs to ensure state is refreshed on switching (maintained selection)

There seems to be no issue in the operation as far as I have tested.

### Applicable Issues
#54


CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
```diff
@@ -490,15 +510,29 @@ public static final SelectableAnalyticPlugin lookupSelectablePlugin(final String
...
 
         private final CheckBox checkbox;
+        private ListCell<SelectableAnalyticPlugin> parent;
         private final AnalyticPlugin plugin;
         private PluginParameters parameters;
         private PluginParameters updatedParameters;
 
         public SelectableAnalyticPlugin(final AnalyticPlugin plugin) {
             this.checkbox = new CheckBox();
+            // Allows triggering of selection listener when a checkbox is changed
```
cygnus-x-1: What is happening here?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
```diff
@@ -582,4 +620,71 @@ public String getName() {
...
+     * Update the display by reading and writing to/from the state attribute. 
+     */
+    private static final class AnalyticViewStateUpdater extends SimpleEditPlugin {
+
+        private final AnalyticConfigurationPane analyticConfigurationPane;
+        private final boolean pluginWasSelected;
+        
+        public AnalyticViewStateUpdater(final AnalyticConfigurationPane analyticConfigurationPane, final boolean categoryWasSelected) {
+            this.analyticConfigurationPane = analyticConfigurationPane;
+            this.pluginWasSelected = categoryWasSelected;
```
cygnus-x-1: Should this be categoryWasSelected?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
```diff
@@ -109,16 +109,6 @@ public AnalyticViewTopComponent() {
                 prerequisiteAttributes.get(attribute).add(plugin);
             });
         });
-        prerequisiteAttributes.forEach((attribute, plugins) -> {
```
cygnus-x-1: Why have you removed all the attribute listeners? And is there a reason you haven't implemented the structure and state listeners above?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
```diff
@@ -582,4 +620,71 @@ public String getName() {
...
+                        checkedPlugins.add(selectablePlugin);
+                    }
+                });
+                if (!checkedPlugins.isEmpty()) {
+                    currentState.addAnalyticQuestion(analyticConfigurationPane.currentQuestion, checkedPlugins);
+                }
+                graph.setObjectValue(stateAttributeId, 0, currentState);
+            }
+            
+            // get question only if one exists
```
cygnus-x-1: Shouldn't this all be done in the state listener?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
```diff
@@ -65,7 +66,11 @@ public void setCurrentAnalyticQuestionIndex(final int currentAnalyticQuestionInd
 
     public void addAnalyticQuestion(final AnalyticQuestionDescription<?> question, final List<SelectableAnalyticPlugin> selectablePlugins) {
         activeAnalyticQuestions.add(question);
-        activeSelectablePlugins.add(selectablePlugins);
+        for (SelectableAnalyticPlugin plugin : selectablePlugins) {
```
cygnus-x-1: What's the advantage of this, isn't it functionally equivalent to what was there before?

cygnus-x-1: WHy are you doing this? In fact why is activeSelectablePlugins a list of lists?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
cygnus-x-1: Why have you removed all the attribute listeners? Also, is there a reason you haven't implemented the structure and state listeners above?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
cygnus-x-1: Shouldn't all this be in the state listener on the top component?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
cygnus-x-1: Should this be categoryWasSelected?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
```diff
@@ -490,15 +510,29 @@ public static final SelectableAnalyticPlugin lookupSelectablePlugin(final String
...
         private final CheckBox checkbox;
+        private ListCell<SelectableAnalyticPlugin> parent;
         private final AnalyticPlugin plugin;
         private PluginParameters parameters;
         private PluginParameters updatedParameters;
 
         public SelectableAnalyticPlugin(final AnalyticPlugin plugin) {
             this.checkbox = new CheckBox();
+            // Allows triggering of selection listener when a checkbox is changed
+            this.checkbox.selectedProperty().addListener((observable, oldValue, newValue) -> {
```
cygnus-x-1: What is this trying to achieve?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
aldebaran30701: https://github.com/constellation-app/constellation/blob/fbce6de64cf65397f7d6065cb58cd2b85b3c8e2c/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java#L527-L535
So, for an event to fire on the selection model of the pluginList, the plugin has to change state to highlighted. If the plugin was already highlighted but did not have it's checkbox "checked", a user could select the checkbox and no update event would occur to store the state.
By adding a listener to the checkbox, I could trigger an event in it's parent (Listviewcell)'s selection model event. I felt that this would be the cleanest way rather than creating another public method to call once the checkbox event was called.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
aldebaran30701: I did some refactoring of names to make it more readable, but forgot to change that. I'll fix that.

aldebaran30701: A missed variable when refactoring for ease of reading. Ill get that fixed.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
aldebaran30701: I had the impression it was a design decision for something that may be implemented in the future possibly to do with saving/loading of questions/plugins.

aldebaran30701: Yes, I overlooked the change when working on another solution. This should work fine without the change.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
aldebaran30701: Regarding the attribute listeners, 
They seemed to be running over most if not all plugin parameters and updating their parameters to the default parameters. Do you have a suggestion on how this would be possible to work around, whilst still allowing them to be added for the plugins not on this repo?
I can take a look at refactoring the code to include these state listeners if you feel they would benefit as opposed to the current solution.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
aldebaran30701: See https://github.com/constellation-app/constellation/pull/215#discussion_r372100518

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
aldebaran30701: On testing, I now recall that the reason for this was to pass a list of selected plugins to the state class. This state class then grabbed each one and singly added them to a list and added that list as an entry to the activeSelectablePlugins list.
Regarding the double list structure, and your comment in the original issue post; Would you like this refactored, or the implementation changed to match the design decision of including multiple plugins in a single list to suit the question and tabbed design?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
aldebaran30701: _I can take a look at refactoring the code to include these state listeners if you feel they would benefit as opposed to the current solution._

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
cygnus-x-1: Yes I think that would be a better design.

cygnus-x-1: Sorry, I recalled why it was structured this way after I wrote this comment - please disregard.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
cygnus-x-1: Sorry about the duplicate comments, I accidentally left the page while reviewing and didn't realise GitHub must have submitted my comments automatically on exit.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
cygnus-x-1: So what _should_ be happening here is that individual analytics can specify prerequisite attributes (ie. attributes they rely on for their parameters) using ```AnalyticPlugin.getPrerequisiteAttributes``` along with a way of handling changes to those attributes using ```AnalyticPlugin.onPrerequisiteAttributeChange```. This block of code is generically attaching listeners against all specified prerequisite attributes and handling them with the provided handler functions. I would expect many analytics to be affected by this as ```ScoreAnalyticPlugin``` uses this to listen for changes to the ```AnalyticConcept.TransactionAttribute.TYPE``` attribute in order to update its "Transaction Types" parameter when types are added to or removed from the graph. If it's not quite behaving on an update, then it might be worth taking a look at ```ScoreAnalyticPlugin.onPrerequisiteAttributeChange``` and ```FactAnalyticPlugin.onPrerequisiteAttributeChange```.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
cygnus-x-1: Cool - that makes sense.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
aldebaran30701: Hi @cygnus-x-1 I've noticed that within the section I removed for listeners and prerequisite attributes, there is a call to createParameters().

https://github.com/constellation-app/constellation/blob/b9389f271320f33bb94ee8115ce96376d5e344da/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/analytics/AnalyticPlugin.java#L80-L83

I believe that when it hits line 115, the plugins that do not implement any special behaviour just act out the superclass method above. Which just returns the default parameters from a plugin.
https://github.com/constellation-app/constellation/blob/b9389f271320f33bb94ee8115ce96376d5e344da/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java#L112-L122

These are then set at line 117 which overrides the different parameter values at runtime. 

As createParameters() is utilized by many other classes, changing the default behaviour may come with unwanted bugs. How could I handle this to keep it from breaking other code.

I'll also push the rest of the code, this is the last of the pull request fixes currently in review.

cygnus-x-1: That sounds right - if an ```AnalyticPlugin``` does not have any parameters, then it'll simply use an empty ```PluginParameters```  inherited from the abstract base class. 

The listeners in the ```AnalyticViewTopComponent``` will then react to changes to a "prerequisite attribute" by updating that plugins parameters (which defaults to no parameters) using that plugins ```onPrerequisiteAttributeChange``` implementation (which defaults to a no-op). Note that this code also takes a copy of the parameters and updates ```SelectableAnalyticPlugin.updatedParameters``` instead of ```SelectableAnalyticPlugin.parameters``` in order to avoid events being triggered by changes to these parameters until the Analytic View is ready for them (```AnalyticConfigurationPane.updateSelectablePluginsParameters```).

```Plugin.createParameters``` is simply designed to set-up and return the default parameters for a plugin, generally to be used at the beginning of a plugin lifecycle. It is designed to be overridden to make it specific to a plugin implementation, and I believe that's exactly how it's being used here. Are you seeing any specific issues manifesting with this code?

aldebaran30701: > Plugin.createParameters is simply designed to set-up and return the default parameters for a plugin, generally to be used at the beginning of a plugin lifecycle. It is designed to be overridden to make it specific to a plugin implementation, and I believe that's exactly how it's being used here.

You are not wrong in saying this, It currently gets the default `PluginParameters` during the `createParameters()` call. Then, obviously it `.copy()` 's the default parameters values.
The main issue I see is that within `.manualUpdate()`, all listeners are triggered to update things. Including the PrerequisiteAttributeListeners. The graph is read from and state attribute written to first. Then when you launch the Analytic View, it does a `.manualUpdate()`.

As you can see in this step through, the first image is line 117 getting the `SelectableAnalyticPlugin`. Note the red circle showing the boolean value read from the state attribute. 
before updating parameters

In this second screenshot, It displays the value of the `updatedParameters` which are going to replace the current values. Note the highlighted boolean value down the bottom showing the `SelectableAnalyticPlugin` parameters value which will now be set to true after execution of this line. 
image

All of this occurs when the following preconditions are met:
 - Graph is currently opened
 - Analytic View is being opened

Line 115 and 116 seem to work perfect for the use case when the prerequisite attribute changes, but on triggering of all handlers, it writes defaults to them which override the loaded parameters from state.

cygnus-x-1: Okay I think I understand the problem you're describing - essentially any parameters which are not updated by ```AnalyticPlugin.onPrerequisiteAttributeChange``` explicitly will instead be updated to the default value for that parameter when ```SelectableAnalyticPlugin.updatedParameters``` is copied across to ```SelectableAnalyticPlugin.parameters```, overwriting any other changes, such as those set by the state. Yeah that's a tough one - i'll have to think about that.

aldebaran30701:
I've fixed the issue of the parameter state being refreshed by default values by adding a suppression flag. This should now be fully implemented.

PR #213, opened 2020-01-24
Font improvements

cygnus-x-1:
### Description of the Change

Added ConstellationLabelFonts interface and integrated it with LabelFontsOptionsPanel to allow developers to programmatically set defaults for fonts rendered on a graph.

Updated the "Show in New Table View" context menu to actually show the selection in the Table View, rather than just open it.

Added a default arrangement to the Icon Graph builder to avoid all the nodes being rendered on top of each other.

### Alternate Designs

N/A

### Why Should This Be In Core?

These  changes all provide for a better user experience.

### Benefits

N/A

### Possible Drawbacks

N/A

### Verification Process

Exploratory testing.

### Applicable Issues

N/A

PR #222, opened 2020-02-03
Fix font issues - Issue #163

### LINKED ISSUE #163: Graph renderer doesn't support different language fonts
### Prerequisites

### Description

Constellation had great support for different languages because it relied on the [Arial Unicode MS Font](https://en.wikipedia.org/wiki/Arial_Unicode_MS) however MS Office 2016 dropped support for the font due to licensing costs I think, Constellation needs to find a font we can use to support different languages again.

A work around was to default to the `Malgun Glothic` which does a good job with Chinese characters but fails with Arabic, Hindi etc  (i.e any character not supported by `Malgun Glothic`.).

The attribute editor or table view seems to figure out which font to use and render it but on the graph, OpenGL doesn't have that fallback logic and displays blanks.

### Steps to Reproduce

1. Open a new graph
2. Change to draw mode
3. Click on the graph to create a new node
4. Change to select mode
5. Select the node
6. Bring up the attribute editor by going to Views -> Attribute Editor or Ctrl + Shift + E
7. Click on the `Edit` button besides the `Identifier` attribute
8. Paste some Arabic text
9. Click OK
10. Press F5

**Expected behaviour:** The text should be displayed on the graph

**Actual behaviour:** The label of the node is blank

### Additional Information

The `Malgun Glothic` font is currently hard coded so as a start, we need to make it visible as a preferences option in Constellation so that a user can select which font they want to use.

arcturus2:
### Description of the Change

Fix to Issue #163

* Consolidation of https://github.com/constellation-app/constellation/pull/213 and https://github.com/constellation-app/constellation/pull/220
* Added support to the renderer to display multiple fonts
* Added `ConstellationLabelFonts` interface and integrated it with `LabelFontsOptionsPanel` to allow developers to programmatically set defaults for fonts rendered on a graph.
* Minor refactor to Glyph package structure
* Added help
* Change the order of the text when the final run ends with a LTR type.
* Fix a bug when more than one font texture page is used. The texture coordinate array was being reset on a new page.

Unfortunately some good changes where made by @cygnus-x-1 however are unrelated to issue #163 and have sneaked in. They are good changes so I'd like to let them through this time to avoid @cygnus-x-1 from having to redo this in a separate PR. The requirement for 1 change per PR still stands. These changes include:

* Moved icons in modules to a `resources` sub package
* Fixed HTML warnings
* Renamed CSS files with a dash

The fonts functionality is domain agnostic and already part of Core

### Benefits

* Fixes #163 
* Code cleanup and consistency

### Verification Process

* Follow steps in https://github.com/constellation-app/constellation/blob/new-fonts/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/docs/label-fonts.html
* Create a sphere graph with 1000 nodes and check the "All displayable chars" box. Exception before, works after.

### Applicable Issues

* #163


formalhaut69:
The updates to the UI look good. A small comment, when the user adds stuff to the label fonts and then checks the use default settings, their changes are lost. Perhaps the checkbox should just grey out the textbox but not delete it's contents.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/NumberedTextArea.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.awt.Color;
+import javax.swing.JTextArea;
+import javax.swing.text.Element;
+
+/**
+ *
+ * @author algol
+ */
+public final class NumberedTextArea extends JTextArea {
```
cygnus-x-1: This should probably be moved to the CoreUtilities gui package.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/glyphs/DefaultLabelFonts.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author cygnus_x-1
+ */
+@ServiceProvider(service = ConstellationLabelFonts.class, position = Integer.MAX_VALUE)
+public class DefaultLabelFonts implements ConstellationLabelFonts {
+
+    @Override
+    public List<String> getFontList() {
+        final List<String> fontList = new ArrayList<>();
+//        fontList.add("SansSerif");
```
cygnus-x-1: Even though the font will fall back to Sans Serif, I think there's value in making it explicit here, in particular so that it shows up in the UI.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/glyphs/DirectionRun.java
```diff
@@ -0,0 +1,197 @@
...
+            final char dir;
+            final int cptype = Character.getType(codepoint);
+            /*codepoint==32*/
+            if ((cptype == Character.SPACE_SEPARATOR || cptype == Character.NON_SPACING_MARK)) {
+                dir = currDir;
+            } else {
+                final char d = codepointDirection(codepoint);
+                dir = d != 'R' ? 'L' : 'R';
+            }
+//            System.out.printf("codepoint %d %d %c %c %d %s\n", offset, codepoint, currDir, dir, Character.getType(codepoint), Character.UnicodeBlock.of(codepoint));
```
cygnus-x-1: Can we get rid of all these commented out print statements?

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/glyphs/GlyphManager.java
```diff
@@ -0,0 +1,159 @@
...
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+/**
+ *
```
cygnus-x-1: New classes should at minimum have class-level javadoc describing what they do.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/glyphs/GlyphManagerFX.java
```diff
@@ -57,9 +59,9 @@
  *
  * @author sirius
  */
-public class GlyphManager {
+public class GlyphManagerFX implements GlyphManager {
```
cygnus-x-1: Sooo... we can just get rid of this now.... right?

sol695510:
Font style only changing non-letter characters
Font style options

Found a bug where only non-letter characters in the label use the set font (notice periods and tildes). Tried both with the Bauhaus 93 and Chiller fonts.

cygnus-x-1:
> Found a bug where only non-letter characters in the label use the set font (notice periods and tildes). Tried both with the Bauhaus 93 and Chiller fonts. I've included the JOGL version here [Fix font issues - JOGL version.txt](https://github.com/constellation-app/constellation/files/4157695/Fix.font.issues.-.JOGL.version.txt).

You've set the font script to COMMON here, which only includes punctuation and other extended characters. You'll want LATIN for the English language characters.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/glyphs/GlyphManagerFX.java
cygnus-x-1: After speaking with @algol60 it sounds like we should keep this around for a few months as a fallback.

arcturus2:
I've fixed some warnings from SonarQube. Going to merge this and fix the rest with separate PR's per issue.

Issue #196, opened 2020-01-13
Analytic view Exceptions on creation / switching of graphs


aldebaran30701:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Illegal Argument Exception or Concurrent Modification Exception when switching between graphs with Analytic view open.
```
java.lang.IllegalArgumentException: Parameter 'BetweennessCentralityPlugin.include_connections_in' does not exist
	at au.gov.asd.tac.constellation.pluginframework.parameters.PluginParameters.getParameter(PluginParameters.java:504)
	at au.gov.asd.tac.constellation.pluginframework.parameters.PluginParameters.setBooleanValue(PluginParameters.java:528)
	at au.gov.asd.tac.constellation.views.analyticview.questions.BestConnectsNetworkQuestion.initialiseParameters(BestConnectsNetworkQuestion.java:75)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.lambda$updateSelectablePluginsParameters$28(AnalyticConfigurationPane.java:476)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.updateSelectablePluginsParameters(AnalyticConfigurationPane.java:474)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticViewTopComponent.lambda$null$7(AnalyticViewTopComponent.java:119)
	at au.gov.asd.tac.constellation.functionality.views.ListeningTopComponent.lambda$manualUpdate$16(ListeningTopComponent.java:330)
	at java.util.HashMap.forEach(HashMap.java:1289)
	at au.gov.asd.tac.constellation.functionality.views.ListeningTopComponent.manualUpdate(ListeningTopComponent.java:325)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticViewTopComponent.handleNewGraph(AnalyticViewTopComponent.java:162)
[...]

PR #227, opened 2020-02-04
Analytic view Exceptions on creation / switching of graphs #196 

formalhaut69:
### Description of the Change
The issue was that the JavaFX thread and the Event queue thread were accessing the currentQuestion in an unsafe way. I've added reentrant locks to the lines of code that cause the exceptions. 

### Alternate Designs
Using a synchronised block to perform the same function. This alternative wasn't used because this type of lock didn't seem to resolve the issue. 

Resolves exceptions when using the analytic view. 

### Benefits
Resolves exceptions when using the analytic view. 

### Verification Process
The steps to reproduce the errors were continuously repeated without issue. 

### Applicable Issues
This only resolves the Illegal argument exception and the first concurrent modification issue that was documented in the ticket. The other concurrent modification exception is still unresolved. 

I only added locks on the lines of code that modified the currentQuestion and caused the exception. I decided not to put a lock on every line of code that accesses the currentQuestion because they don't cause issues and might be unnecessarily taxing to constellation. Would you prefer that I lock currentQuestion every time it is accessed?

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
@@ -43,6 +43,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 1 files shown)


arcturus2:
> Would you prefer that I lock currentQuestion every time it is accessed?

My default preference would be to do it for all calls to `currentQuestion` because it may potentially future proof the code but I am OK to start with what you've done and take it a step at a time.

Issue #225, opened 2020-02-03
Deselect plugins after pressing Go


arcturus2:
In the Data Access View (DAV) I often find that after I run a plugin, I have to un-tick it before I run a different plugin. I'd like the DAV to un-tick all of the plugins I had selected. This should be a preference option (i.e Options menu in the DAV) and it should be turned off by default to not impact the current behaviour.

PR #233, opened 2020-02-05
Deselect plugins after pressing go #225

formalhaut69:
### Description of the Change
Added a check selection in the options menu of the DAV to deselect plugins after pressing Go.

Allows users to decide if they want the plugins in the data access view to be deselected when the Go button is clicked
### Benefits
See above

### Verification Process
The option was used multiple times to ensure it's functionality is an appropriate solution for this enhancement.


arcturus2:
Thank you for this @formalhaut69

PR #239, opened 2020-02-06
Fix to the Support Package class name

arcturus2:
Fix a spelling error with the SupportPackageAction and minor code refactor.

### Description of the Change

* Fix a typo with the spelling to `SupportPackageAction`
* Exposed the log directory so that it can be used in other modules


arcturus2:
I'm also adding everyone into this simple change to provide an example of how the `CHANGELOG.md` is used.

Issue #217, opened 2020-01-29
Raw Exceptions thrown for invalid Date/times in CSV importer


serpens24:
## Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

While experimenting with importing date-time data using the CSV importer while analysing issue #34, i noted that it is possible to enter (invalid) custom formats that result in rawe exeptions being thrown - rather than a more graceful validation.

2 examples of this are entering the following custom formats:
- 'yyyyyyyyyyyyyyyyyyyy'    (note this is 20 characters, 19 or less works fine, the following exception text is seen: java.lang.IllegalArgumentException: The minimum width must be from 1 to 19)
- 'aa'    (note this is 2 characters, one character worked. the following exception text is seen: java.lang.IllegalArgumentException: Too many pattern letters: a)

### Steps to Reproduce

1. Open Delimited Importer
2. Right Click on the DateTime attribute
3. Select Formatter -> Datetime
4. Select Custom
5. Enter a custom date format in line with examples above

**Expected behaviour:** 
no raw exception thrown but ERRROR shown for each entry in the timestamp column - as is the case for other inproperly formatted date times

**Actual behaviour:** 
Raw exception dialogs pop up, which need to be manually closed by pressing cancel.

**Reproduces how often:** 100%

### Additional Information

note - both exceptions are thrown with a stack trace shown. The first element in the stack trace belonging to this project is the same for both exceptions .... I suspect catching the invalid return value and handling it gracefully is the way forward:
au.gov.asd.tac.constellation.importexport.delimited.translator.DatetimeAttributeTranslator.translate(DatetimeAttributeTranslator.java:107)

serpens24:
Note: i intend to fix in conjunction with linked issue #34

arcturus2:
Thanks @serpens24

serpens24:
This issue is being addressed at the same time as #34.

Issue #34, opened 2019-07-26
Import from delimited date-time CUSTOM input issues


arcturus2:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

If you import from Delimited file, and select "custom" for your time stamp to input the correct format, I it doesn't always allow you to input stuff into the user-input field. This occurs specifically if you have already tried to apply a Date-time template to your column.

A work around that seems to work is to close and open the custom date dialog and it works the 2nd time. This has been tricky to debug.

### Steps to Reproduce

1. Open Delimited Importer
2. Right Click on the `DateTime` attribute
3. Select Formatter -> Datetime
4. Select Custom
5. Enter a custom date format

**Expected behaviour:** User can enter a custom date format

**Actual behaviour:** The field is disabled, preventing anything being typed in

**Reproduces how often:** 50%

formalhaut69:
Spent some time on this issue but I can't seem to get to the root of the bug. The issue seems to occur 100% of the time for me whenever any date format other than CUSTOM is selected and then the Datetime Parameters window is reopened. The CUSTOM textbox overall seems to be behaving very strangely. I've observed it's setEnabled() method to have two different effects; changing line 82 to disable the CUSTOM textbox or removing the line entirely will have the same effect of the bug itself (the textbox appears to be enabled but nothing can be typed in). Whilst on line 89, the setEnabled() method functions as expected before the bug occurs so the textbox, along with the small drop down list next to it are disabled and greyed out. 
https://github.com/constellation-app/constellation/blob/36f1c51aa421095e9771f0d539aa0a6741d84b19/CoreImportExport/src/au/gov/asd/tac/constellation/importexport/delimited/translator/DatetimeAttributeTranslator.java#L77-L91

serpens24:
I have experimented with a couple minor changes to allow this to work. ssentially there are 2 settings on the widget that are in play (found in thew code for ValueInputPane) these are its editable state and its disabled state.

The existing code constructor in ValueInputPane.java has the following blocks of code of interest:

Lines 187-194
`
[...]

PR #242, opened 2020-02-09
Feature/issue#34

serpens24:
### Description of the Change

Modifed handling of date time import dialog to ensure CUSTOM format text field is enabled and editable ONLY when CUSTOM format is selected.
This change ensures both editable and enabled values of the widget(s) are kept in synch so that the text field and dropdown are only available/editable when required. Previously they were falling out of synch.

### Alternate Designs

No - this is addressing a bug with limited scope.

Address identified bug.

### Benefits

Ability to set custom format as required. Removal of some raw exceptions being thron for incorrect datetime format strings.

### Possible Drawbacks

The original code, had a comment in ValueInputPane (common code) concerned with abiltiy to read greyed out text (I have no problem reading this), as well as indicating a disabled widget disables scrollbar, in discussiuons we decided to allow this to be the case.

### Verification Process

1. Open new graph.
2. Select Import->From Delimited File ....
3. Select file to import containing a timestamp and drag DateTime from Transaction Attributes column onto it.
4. Right click on Datetime and select Formatter->Datetime.
_Confirm that the Custom Format text entry field is disabled and unable to be selected for edit_
5. Modify the Datetime Format value to CUSTOM.
_Confirm that the Custom Format text entry field is enabled and able to be selected for edit_
6. Change the Datetime Format value a few times including some changes to and from CUSTOM as a value.
_Confirm that the Custom Format text entry field is enabled and able to be selected for edit ONLY when CUSTOM is set as a format_
7. Modify the Datetime Format value to CUSTOM.
8. Enter a valid custom format in line with the raw data and press the OK button.
_Confirm the values are shown with correct formatting_
9. Right click on Datetime and select Formatter->Datetime.
_Confirm the previously set CUSTOM format is displayed and is editable._
10. Repeat step 6 and confirm behaviour is still correct.
**Now testing exception handlling as discussed in #217 **
11. Select CUSTOM format and enter yyyyyyyyyyyyyyyyyyyy.
_Confirm no exception is thrown but datetime values all show with text = ERROR_
12. Right click on Datetime and select Formatter->Datetime.
_Confirm the previously set CUSTOM format is displayed and is editable._
13. Select CUSTOM format and enter aa.
_Confirm no exception is thrown but datetime values all show with text = ERROR_

### Applicable Issues

#34 
#217


CoreImportExport/src/au/gov/asd/tac/constellation/importexport/delimited/translator/DatetimeAttributeTranslator.java
```diff
@@ -77,9 +77,8 @@ public PluginParameters createParameters() {
         final PluginParameter<StringParameterValue> customParam = StringParameterType.build(CUSTOM_PARAMETER_ID);
         customParam.setName("Custom Format");
         customParam.setDescription("A custom datetime format");
-        // The custom field wont enable inside the addMasterController until you
-        // click on OK and open it again. This is a work around until it can be fixed.
-        customParam.setEnabled(true);
+        // customParam should be enabled and editable if "CUSTOM" format has been specified.
+        customParam.setEnabled(formatParam.getStringValue().equals("CUSTOM"));
```
arcturus2: In general it is good practice to do the equals check the other way to avoid potential null pointer errors if you have a constant so:

``"CUSTOM".equals(formatParam.getStringValue())``

Issue #69, opened 2019-07-30
Integrate new JOGL version for the JDK 11 migration


rsabhi:
### Description

As part of JAVA 11 migration, there is a requirement to upgrade JOGL dependent libraries.
Current we use JOGL version 2.3.2 which is compliant to JDK8. 
As per JOGL forum they are progressing to release JOGL update (2.5) for  Java11.
Here is the update from JOGL [forum](http://forum.jogamp.org/jogl-gluegen-available-for-java-11-td4039870.html)

aldebaran30701:
The most current release is currently on JOGL rc 2.4.0 but they have not got the file hosted on maven repository. Investigating other solutions to either get it hosted, or download it from their archives. Would that be a bad solution? It would effectively be like the jhall and javahelp files except from their own repository. I'm just scraping the surface with understanding module dependencies and the maven and ivy work.

arcturus2:
@aldebaran30701 would you have time to follow this up with the JOGL devs via the forum to find out what the reason might be for them not uploading 2.4.0? They may have forgotten.

To get things working it's ok to host the .jar locally as a temporary fix so you can progress with testing.

aldebaran30701:
> would you have time to follow this up with the JOGL devs

Already done. 
http://forum.jogamp.org/Upgrading-from-JOGL-2-3-2-to-Current-master-for-Java-11-support-td4040335.html

aldebaran30701:
Update: 
I added the files locally to the release/modules/etc of dependencies, and on a clean/build it compiled fine with the new JOGL version. 
The thing with this is it deletes the whole directory when doing update-dependencies-clean-build. 
One option would be to mess around with references in the project to point to a local lib folder containing the permanent files.

I've also added file references to the ant build.xml which automatically download the jogl files straight from the servers at [jogamp](https://jogamp.org/deployment/archive/rc/v2.4.0-rc-20200202/).
This implementation seems to work just fine even with update-dependencies-clean-build.
The only issue is reliability of their server and any downtime that url may face. 

I've got a query regarding how to progress wrt deprecated warnings. @arcturus2 
Blue is the additions I put including the suppression of the warning as .md5() still exists, and the deprecation only served as a warning to it's cryptographic security.
![image](https://user-images.githubusercontent.com/58677759/74209658-9f734500-4cd8-11ea-8f7a-59a387cb3c1c.png)

Would this be suitable given the situation? 
[...]

PR #249, opened 2020-02-14
Jogl dependencies

aldebaran30701:
### Description of the Change
This change adds an upgrade to JOGL. The new version is a current WIP of 2.4.0. This PR will need to be rectified once the dependency is available on maven for dependency download.
Updates to attribution file, changelog and build.xml.

### Alternate Designs
1. Download from JOGAMP servers - issues with possible low availability
2. Distribute file locally - increases size of constellation distribution.

It makes way for JDK11 improvements. 

### Benefits
Current code base with recent issues addressed.

### Possible Drawbacks
Possibly less testing done of the JOGL version as it is not a release thus far.

### Verification Process
Test changing of JOGL versions via verifying that 2.4.0 was active. 
Graphs rendered properly, as previous.
Issue regarding offset of graph panel is still present in this version of JOGL

### Applicable Issues
#69


aldebaran30701:
UPDATE: I had some issues early on regarding the re download of old JOGL 2.3.2 files existing after even clearing the cache. I switched off my computer over the weekend and it seems to have fixed the re downloading. 
_However_, it has highlighted some other dependencies which may have been downloaded as a dependency. these are Nativewindow and Newt. I'll verify that this was due to removal of JOGL 2.3.2. dependency download and push a suitable fix if required.

PR #253, opened 2020-02-20
add badges + logo

TheDen:
PR adds badges and the constellation logo to the readme


arcturus2:
Thank you for this @TheDen. It looks so much better!!

PR #257, opened 2020-02-21
ivy-update

cygnus-x-1:
:art: updated to latest ivy distribution

### Description of the Change

Updated to the latest distribution of Ivy, which is used to resolve dependencies for Constellation.

### Alternate Designs

N/A

It's good to stay current.

### Benefits

N/A

### Possible Drawbacks

N/A

### Verification Process

Tested by building / updating dependencies, verifying the old and new Ivy caches were equivalent, and then running Constellation and performing manual testing to look for any obvious issues..

### Applicable Issues

N/A


cygnus-x-1:
@arcturus2 i looked over the downloaded dependencies as you suggested and i believe the differences are all due to dependencies of ivy (and ant) being downloaded.

arcturus2:
A before and after compare. It's an increase of 39 MB or 107 files...

image

arcturus2:
@cygnus-x-1 I'm going to hold off on this PR until I get the `ProjectUpdater` code added to master. Right now those new jars are not being used because it hasn't updated project.xml with references to the jars. I'll test this again once we have those jars included and then check again with you.

I would like to upgrade because there are new features in ivy that I would like to use.

arcturus2:
@cygnus-x-1 it looks like some dependencies are related testing frameworks like mokito. Perhaps there is a bug with the existing ivy that wasn't downloading all dependencies. My thinking is that we would need to go through them and probably add them to the exclude list.

arcturus2:
One nice thing about the new version of ivy is that you can get it to check for latest dependency versions. Here is what it came back with...

```
Dependencies updates available :
	org.geotools#gt-cql	21.2 -> 22.2
	org.geotools#gt-jdbc	21.2 -> 22.2
	org.geotools#gt-main	21.2 -> 22.2
	org.geotools#gt-metadata	21.2 -> 22.2
	org.geotools#gt-opengis	21.2 -> 22.2
	org.geotools#gt-referencing	21.2 -> 22.2
	org.geotools#gt-xml	21.2 -> 22.2
	org.geotools#gt-epsg-hsql	21.2 -> 22.2
	org.geotools#gt-geopkg	21.2 -> 22.2
	org.geotools#gt-shapefile	21.2 -> 22.2
	org.geotools#gt-geojson	21.2 -> 22.2
	org.geotools.xsd#gt-xsd-core	21.2 -> 22.2
	org.geotools.xsd#gt-xsd-kml	21.2 -> 22.2
	org.apache.commons#commons-csv	1.7 -> 20110211
	org.apache.commons#commons-lang3	3.8.1 -> 3.9
	org.apache.commons#commons-math3	3.2 -> 3.6.1
	org.apache.commons#commons-text	1.6 -> 1.8
	com.fasterxml.jackson.core#jackson-core	2.10.1 -> 2.10.2
	com.fasterxml.jackson.core#jackson-core	2.9.9 -> 2.10.2
	com.fasterxml.jackson.core#jackson-databind	2.10.1 -> 2.10.2
	com.google.guava#guava	27.0-jre -> 28.2-jre
	gov.nasa#worldwind	2.0.0 -> 2.0.0-986
	org.apache.poi#poi	4.1.1 -> 4.1.2
	org.apache.poi#poi-ooxml	4.1.1 -> 4.1.2
	org.apache.poi#poi-ooxml-schemas	4.1.1 -> 4.1.2
	org.controlsfx#controlsfx	8.40.16 -> 11.0.1
	org.eclipse.jetty#jetty-server	9.4.24.v20191120 -> 10.0.0.alpha1
	org.eclipse.jetty#jetty-servlet	9.4.24.v20191120 -> 10.0.0.alpha1
	org.python#jython-standalone	2.7.2b2 -> 2.7.2b3
```

cygnus-x-1:
> it looks like some dependencies are related testing frameworks

You're right, and I actually think many of them are coming from geotools. Odd that they didn't show up before. There must be a way to bulk exclude them (based on their 'test' scope in the pom files). Manually excluding individual jars is not sustainable.

cygnus-x-1:
Okay I think I've got Ivy set up correctly now with configurations (and a sensible default). It's still bringing back more jars than previously, but the total disk usage is roughly the same. I updated all dependencies while I was at it, so the additional jars could also be to do with that.

@arcturus2 Please take a look and see if you can find any jars that need to be explicitly ignored (which can be required if our dependencies are not managing their dependencies correctly).

PR #258, opened 2020-02-22
issue-179

cygnus-x-1:
🎨 Updated editor date pickers to listen to their text property in order to provide instant feedback (where the value property fails)
🎨 Updated the converters used by editor date pickers to use the TemporalFormatting.DATE_FORMATTER datetime formatter
🎨 SchemaVertexTypeUtilities.getTypeOrBuildNew and SchemaTransactionTypeUtilities.getTypeOrBuildNew now adds new types to schema
🎨 Improved blaze validation by checking for valid angle
🎨 Imporved vertex_attribute_name and transaction_attribute_name validation by ensuring typed names exist on graph
🎨 Renamed NodeGraphLabelsEditorFactory to VertexGraphLabelsEditorFactory

### Description of the Change

Updated all editors using JavaFX Date Pickers to use Constellation's standard date format and to provide better feedback where there are errors

### Alternate Designs

N/A

Better UX is always a good thing.

### Benefits

N/A

### Possible Drawbacks

N/A

### Verification Process

Tested all editors using manual testing of custom attributes.

### Applicable Issues

N/A

### DIFF
--- a/CoreAttributeEditor/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/AbstractEditorFactory.java
+++ b/CoreAttributeEditor/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/AbstractEditorFactory.java
@@ -235,6 +235,5 @@ public abstract class AbstractEditorFactory<V> {
         public String getReason() {
             return reason;
         }
-
     }
 }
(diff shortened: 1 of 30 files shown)


cygnus-x-1:
Some additional issues I noticed while making these changes:
1. There is a discrepancy between legacy attributes (attr_list, layer_name), which have have no editors (and so their edit button is greyed out), and explicitely non-editable attributes (raw, composite_node_state) which use the default editor but warn they cannot be edited.
2. Time spinners (time, datetime, local_datetime) are shrunk to an unusable size on high resolution (4k) displays.

PR #265, opened 2020-02-25
use constellationapplication/netbeans-runner:11 image

TheDen:
Test out the `constellationapplication/netbeans-runner:11` docker image


arcturus2:
@TheDen great stuff, it worked!

PR #275, opened 2020-03-04
Update to README formatting

arcturus2:
### Description of the Change

Minor update to formatting and moved some of the pre-requisites to the build Constellation section because we bundle the JRE with Constellation so users don't have to have Java installed.


arcturus2:
@TheDen I'm not sure why the build failed. It errors with the `sonar.sh` it seems. I swear I didn't make any changes to it!

PR #299, opened 2020-03-18
properly handled interrupt exceptions

antares1470:
### Description of the Change

Handled interrupt exceptions properly by adding thread interrupts where they were missing (as suggested by sonar)


arcturus2:
Thanks for this @antares1470

Issue #296, opened 2020-03-17
Moving nodes on Graph doesn't update associated attributes in Attribute Editor


serpens24:
## Description
NOTE: This issue was seen on NB11 version of the application, but is known to also occure on NB8 version.

Currently if the Attribute Editor is open along with a graph, and certain attributes on the graph change, these changes are not refelcted in the Attribute Editor.

The most obvious example is moving s node around on the graph... its positional values do not update in the Attribute Editor unless you click off it, then back onto it.

### Steps to Reproduce

1. Open a graph
2. Open the Attribute Editor an expand the Node block
3. Select a Node on the graph and note its X/Y position in the Attribute Editor
4. Move this node and observe its X?Y position do not change
5. Click off the node (onto another node) and back onto it and note the nodes X/Y values have now updated

**Expected behaviour:**
Node atributes uopdate in real time as its postion changed

**Actual behaviour:**
Atributes are not updated

**Reproduces how often:**
100%

### Additional Information

1. Acturus beleives this was working prior to some recent Attribute Editor changes.

2. I had a quick look at AttributeEditorTopComponent.java and put a breakpoint in graphChanged. I noted that when moving the node this breakpoint was hit, but code always falls into the "event.getEditor() == null" block and returns. 
For a quick test I commented out this if block and found that moving the node and releasing it instantly updates the associated X/Y attributes. A git blame suggests that graphChanged has not been modified for at least 8 months, so perhaps getEditor is returning null in error ?
Looking into this, th ecall stack suggests GraphChangedEvent() at line 49 is being called so set th editor. This might be a clue as to what is going on.

** NOTE ** I was able to replicate this issue in release 1.2.1 which is significantly earlier than was thought that the issue existed, so I'm confirming that my steps are in fact demonstrating the issue that was discussed with @arcturus2  before we look to proceed.

arcturus2:
@serpens24 for some reason I thought this was a recent bug but based on your investigation it must have been earlier last year. The reproducer is correct.

serpens24:
Thanks @arcturus2.

Subsequent to our chat earlier today I investigated if editor is ever assigned a non-null value within 
LockingManager.LockingEdit and it does appear it can occur, ot with DoalGraph. As such I have elected to not pursue checking its use any further as its clear it's not alway going to be set to null.

So the change which resolves the issue in question relates to lines 227-232 of AttributeEditorTopComponent.java. These lines are as per below extract:

`
[...]

PR #303, opened 2020-03-19
Removed check for null editor which was preventing some graph changes…

serpens24:
### Description of the Change

Modify check which is preventing some graph changes (ie pan, rotate, node move) from reflecting updates to associated attirbutes in attribute editor.

### Alternate Designs

nil

bug fix, requested by advocates

### Benefits

Changes will immediately reflect in Attribute Editor

### Possible Drawbacks

None known

### Verification Process

1. Open a graph
2. Open the Attribute Editor an expand the Node block
3. Select a Node on the graph and note its X/Y position in the Attribute Editor
4. Move this node and confirm after moving it, that its position updates in the Attribute Editor
5. Rotate the graph and confirm the Graph cmarea attribute changes
6. Pan the graph and confirm the Graph cmarea attribute changes

### Applicable Issues

https://github.com/constellation-app/constellation/issues/296


serpens24:
identified missing line.

PR #305, opened 2020-03-19
Feature/issue#296

serpens24:
### Description of the Change

Modify check which is preventing some graph changes (ie pan, rotate, node move) from reflecting updates to associated attirbutes in attribute editor.

### Alternate Designs

nil

bug fix, requested by advocates

### Benefits

Changes will immediately reflect in Attribute Editor

### Possible Drawbacks

None known

### Verification Process

1. Open a graph
2. Open the Attribute Editor an expand the Node block
3. Select a Node on the graph and note its X/Y position in the Attribute Editor
4. Move this node and confirm after moving it, that its position updates in the Attribute Editor
5. Rotate the graph and confirm the Graph cmarea attribute changes
6. Pan the graph and confirm the Graph cmarea attribute changes

### Applicable Issues

https://github.com/constellation-app/constellation/issues/296

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
@@ -225,12 +225,6 @@ public final class AttributeEditorTopComponent extends TopComponent implements G
     @Override
     public void graphChanged(GraphChangeEvent event) {
         event = event.getLatest();
-        /*event.getEditor() is null whenever this event is invoked due to an action 
-        which is not relevent to this class e.g. zoom in/out, therefore return 
-        from this method without any further execution */
-        if (event.getEditor() == null) {
-            return;
-        }
         if (event.getId() > latestGraphChangeID) {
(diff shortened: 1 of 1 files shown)


arcturus2:
Thanks @serpens24 so this gets the functionality back to how it used to be. I can now see the values like `x` update.

What sparked the original investigation was that when you have a selection and show the group of values in the attribute editor expanded, and then zoom, it will collapse them again. The main concern for me was the possible performance implication of the attribute editor constantly "refreshing". I never profiled it so maybe I'm just over thinking it.

A user would surely find this annoying to have to expand the group each time right? Or is that not really a situation you have? Any thoughts @cygnus-x-1?

See the gif below for a demonstration...:smile:

Attribute editor refreshing

Anyway I don't think I communicated this to you so this is a new request. Happy to merge this PR to get things back which is more important than the performance hit perhaps??

serpens24:
OK, thaka for the GIF, So what you have shown there I wasn't aware of, I do agree that issue is a differnet issue to what I was looking at based on the conversation, so perhaps merging this and raising  one to capture the problem is a good idea.

arcturus2:
I've created #316 to track the attribute editor refresh bug. Going to merge this.

PR #306, opened 2020-03-19
Random reusable

antares1470:
### Description of the Change

- converted uses of Random to be using Single instances
- changed some type casts to occur on operand rather than operation
- added NoSuchElementException to some instances of Iterator.next()
- corrected mistake from previous PR: single quotes now properly escaped

All these brought up by Sonar


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/json/JsonUtilities.java
```diff
@@ -92,6 +92,9 @@ public boolean hasNext() {
 
             @Override
             public String next() {
+                if (!hasNext()) {
```
arcturus2: 👍

PR #315, opened 2020-03-20
Switch statements

antares1470:
### Description of the Change

Changes suggested by Sonar:
- added default cases to switch statements where missing
- moved default cases to end of switch statement
- converted switch statements with only 2 cases to if-else statements


CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
```diff
@@ -208,6 +208,7 @@ protected void edit(final GraphWriteMethods graph, final PluginInteraction inter
                                 graph.setBooleanValue(transactionSelectedAttribute, transactionId, elementIds.contains(transactionId));
                             }
                             break;
+                        default:
```
cygnus-x-1: Although it will fall through and likely work, isn't it good practice to use break after default?

arcturus2: For the reasons mention in [this answer](https://softwareengineering.stackexchange.com/a/201786) I'd agree with adding a break at the end; for consistency - i.e. The Principle of Least Surprise

antares1470: This has now been done. I applied to a few extra files that also didn't have breaks in the default case

arcturus2:
@cygnus-x-1 merged this change given @antares1470 has addressed your feedback.

PR #330, opened 2020-03-23
Collapsible ifs

antares1470:
### Description of the Change
As suggested by Sonar, have collapsed nested if statements where appropriate

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
@@ -31,6 +31,9 @@ import au.gov.asd.tac.constellation.plugins.parameters.PluginParameters;
 import au.gov.asd.tac.constellation.plugins.parameters.RecentParameterValues;
 import au.gov.asd.tac.constellation.plugins.parameters.types.DateTimeRange;
 import au.gov.asd.tac.constellation.plugins.templates.SimplePlugin;
+import au.gov.asd.tac.constellation.utilities.color.ConstellationColor;
+import au.gov.asd.tac.constellation.utilities.icon.AnalyticIconProvider;
+import au.gov.asd.tac.constellation.utilities.icon.UserInterfaceIconProvider;
 import au.gov.asd.tac.constellation.views.dataaccess.CoreGlobalParameters;
 import au.gov.asd.tac.constellation.views.dataaccess.DataAccessPlugin;
 import au.gov.asd.tac.constellation.views.dataaccess.DataAccessPluginCoreType;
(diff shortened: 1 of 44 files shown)


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
```diff
@@ -222,19 +222,17 @@ public DataAccessPane(DataAccessViewTopComponent topComponent) {
...
-                    graphAction.actionPerformed(null);
-                    while(GraphManager.getDefault().getActiveGraph() == null){
-                        // Wait and do nothing while graph is getting made
-                    }
-                    graphId = GraphManager.getDefault().getActiveGraph().getId();
-                    if (!graphState.containsKey(graphId)) {
-                        graphState.put(graphId, new GraphState());
-                    }
-                    currentGraphState = graphState.get(graphId);
+            if(graphId == null && pluginSelected && selectedPluginsValid){
```
arcturus2: Is that the same as?

```
if(graphId == null && (pluginSelected && selectedPluginsValid)){
```

CoreGraphUtilities/src/au/gov/asd/tac/constellation/graph/utilities/io/CopyGraphUtilities.java
```diff
@@ -122,20 +122,11 @@ public static void copyGraphToGraph(final GraphReadMethods original, final Graph
...
-                }
-                if (vertexSelected != Graph.NOT_FOUND) {
-                    if (!original.getBooleanValue(vertexSelected, original.getTransactionSourceVertex(originalTransaction))) {
-                        continue;
-                    }
-                    if (!original.getBooleanValue(vertexSelected, original.getTransactionDestinationVertex(originalTransaction))) {
-                        continue;
-                    }
-                }
+            if (!copyAll && ((transactionSelected != Graph.NOT_FOUND && !original.getBooleanValue(transactionSelected, originalTransaction)) ||
```
arcturus2: if you don't have a preference, can the `||` and `&&` be added to the start of the line rather than the end?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
antares1470: Given they are all &&'s, it should be. I added braces to statement where I felt that distinction was needed (usually if it also involved ||'s). Otherwise (in the case of only &&'s) I opted to not include braces because they wouldn't make a difference to the logic

arcturus2: Yep you are right. For some reason I thought the first was an `||`. Just ignore me!

CoreGraphUtilities/src/au/gov/asd/tac/constellation/graph/utilities/io/CopyGraphUtilities.java
antares1470: Done

Issue #280, opened 2020-03-10
Working through SonarQube Warnings


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

This issue has been created to communicate work being done on fixing SonarQube warnings. If you would like to help out with fixing these warnings then please add a comment to this issue of which "rule" you are going to work on so we don't double up.

Please PR a change which works on 1 rule at a time so that it can be reviewed and tested easily.

cygnus-x-1:
This might be a dumb question - but how do we access SonarQube?

sol695510:
https://sonarcloud.io/dashboard?id=constellation-app_constellation @cygnus-x-1

sol695510:
**RULE** (Java) Unnecessary imports should be removed

sol695510:
> **RULE** (Java) Unnecessary imports should be removed

![SonarQube - Unecessary Imports](https://user-images.githubusercontent.com/58677774/76393941-8687ad80-63c4-11ea-99f7-481f47d64ce2.PNG)
![AbstractGeoExportPlugin java - Static Imports](https://user-images.githubusercontent.com/58677774/76393957-91dad900-63c4-11ea-9ce7-8f1bdc5683f1.PNG)
![NewLineRenderable java - Static Imports](https://user-images.githubusercontent.com/58677774/76393964-956e6000-63c4-11ea-828b-684608c139a9.PNG)

SonarQube points these out as unnecessary imports but checking the classes they're in, they have been used. Noted the difference between these and the other marked unnecessary imports is that they are static.

arcturus2:
Hi @sol695510 so this is a tricky one. Taking ```import static au.gov.asd.tac.constellation.importexport.geospatial.AbstractGeoExportPlugin.ELEMENT_TYPE_PARAMETER_ID;``` as an example the import isn't used from what I can tell because a constant with the same name created at this line:

[...]

PR #360, opened 2020-03-25
RULE (Java) Empty arrays and collections should be returned instead of null

sol695510:
### Description of the Change
Altered code blocks in classes mentioned in SonarQube issue to return an empty collection instead of an empty collection or array.

### Alternate Designs
N/A

Returning null instead of an actual array or collection forces callers of the method to explicitly test for nullity, making them more complex and less readable. (As explained in SonarQube)

### Benefits
More readable code.

### Possible Drawbacks
N/A

### Verification Process
N/A

### Applicable Issues
#280


cygnus-x-1:
With all these methods now returning the correct return type instead of null, have you checked that any usages are not performing redundant null checks?

sol695510:
image
@cygnus-x-1 I thought maybe some of @antares1470 work might've already addressed that so I didn't go modify those usages in case they were editing those classes and maybe cause a conflict. Can you advise on how to proceed?

cygnus-x-1:
Some of these methods may have been designed with null as an explicit "fail condition" in mind, and in those so cases passing an empty list would break that design by signalling "success". I think it's worth checking...

sol695510:
@cygnus-x-1 I'll check them.

arcturus2:
@sol695510 please let me know how you are going with this and whether you need a hand. Note that the PR is failing to build so you might need to pull the latest master and double check things too. Thank you. If you'd like someone to look at this just let us know.

sol695510:
@arcturus2 I'm finishing up checking for null checks as @cygnus-x-1 stated before. About 80% finished, apologies for the delay.

arcturus2:
> @arcturus2 I'm finishing up checking for null checks as @cygnus-x-1 stated before. About 80% finished, apologies for the delay.

Oh don't apologies @sol695510, take your time 👍

sol695510:
@arcturus2 I'm not sure what's causing the SonarCloud and Travic CI - Branch checks to fail.

sol695510:
@arcturus2 For now I'll tackle a different SonarCloud issue.

arcturus2:
>This branch is a branch off of the old master-nb11 so I'm not sure if that is the problem now.
It probably will be because there have been fixes to Travis in the master branch. In theory you should be able to merge master ontop of your branch and then push your branch again. Can you try that? If you'd like someone else to look into this then let me know. No rush as such, happy to wait for when you can get onto this.

sol695510:
> > This branch is a branch off of the old master-nb11 so I'm not sure if that is the problem now.
> 
> It probably will be because there have been fixes to Travis in the master branch. In theory you should be able to merge master ontop of your branch and then push your branch again. Can you try that? If you'd like someone else to look into this then let me know. No rush as such, happy to wait for when you can get onto this.

I've done that and the Travis CI checks have passed but I'm not sure what the SonarCloud is picking up.

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/composite/ContractCompositePlugin.java
```diff
@@ -81,10 +81,10 @@ public void selectItem(String item, final Graph graph, GraphElementType elementT
             if (compositeNodeState != null && compositeNodeState.comprisesAComposite()) {
                 return Arrays.asList("Contract Composite");
             } else {
-                return null;
+                return Arrays.asList();
```
arcturus2, line 84: @sol695510 is there any reason your using this instead of `Collections.emptyList()` because I noticed for one of the files you have use it. If no particular reason then would you be able to standardise on `Collections.emptyList()`?

arcturus2:
> I've done that and the Travis CI checks have passed but I'm not sure what the SonarCloud is picking up.

That's fine. It's complaining about lack of code coverage and there is a bug I still need to fix.

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/composite/ContractCompositePlugin.java
sol695510: I'll make the change.

antares1470:
> 
> 
> image
> @cygnus-x-1 I thought maybe some of @antares1470 work might've already addressed that so I didn't go modify those usages in case they were editing those classes and maybe cause a conflict. Can you advise on how to proceed?

@cygnus-x-1 @sol695510 The sonar rule is only about using Collections.isEmpty() instead of size() >0 so shouldn't have an impact here (unless you happen to be using size >0 checks)

sol695510:
> If you make the change suggested by @arcturus2, the rest of it looks fine

Pushed the change.

PR #361, opened 2020-03-25
Rule (java) asserts should not be used to check the parameters of a public method

formalhaut69:
### Description of the Change
Made changes to conform to this rule.


cygnus-x-1:
I like using asserts in this way, and can see that most of the fixes are on classes I wrote - perhaps I should leave someone else to review this...

arcturus2:
> I like using asserts in this way, and can see that most of the fixes are on classes I wrote - perhaps I should leave someone else to review this...

That's good of you @cygnus-x-1.

I like the convenience of the `assert` usage too but I'd prefer we standardise on what SonarQube suggests. If we run into situations were we disagree with the SonarQube recommendation with strong views then we reconsider.

arcturus2:
Going to merge this @cygnus-x-1

PR #370, opened 2020-03-26
Increase the default sphere size and add some random blazes.

algol60:
### Description of the Change
Increase the default sphere size and add some random blazes.

### Alternate Designs
N/A

N/A

### Benefits
N/A

### Possible Drawbacks
N/A

### Verification Process
N/A

### Applicable Issues
N/A


aldebaran30701:
Looks much better!

PR #391, opened 2020-04-02
RULE (Java) Methods and field names should not be the same or differ only by capitalization

formalhaut69:
### Description of the Change

Made changes to conform to this rule


CoreScriptingView/src/au/gov/asd/tac/constellation/views/scripting/graph/SGraph.java
```diff
@@ -56,14 +56,14 @@
...
-    private final Graph graph;
+    private final Graph graphObject;
 
     private final ArrayList<SReadableGraph> readableGraphs;
     private final ArrayList<SWritableGraph> writableGraphs;
 
     public SGraph(final ScriptEngine engine, final Graph graph) {
         this.engine = engine;
-        this.graph = graph;
+        this.graphObject = graph;
```
cygnus-x-1, line 66: Is this sort of thing really an issue? I thought it was fine given the variables have clearly defined, different scopes.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/templates/RecordStoreQueryPlugin.java
```diff
@@ -111,13 +111,13 @@ protected RecordStore getResult() {
...
-                query = GraphRecordStoreUtilities.getSelectedVertices(graph);
+                recordStoreQuery = GraphRecordStoreUtilities.getSelectedVertices(graph);
                 break;
             case GraphRecordStoreUtilities.TRANSACTION:
-                query = GraphRecordStoreUtilities.getSelectedTransactions(graph);
+                recordStoreQuery = GraphRecordStoreUtilities.getSelectedTransactions(graph);
                 break;
             case GraphRecordStoreUtilities.ALL:
-                query = GraphRecordStoreUtilities.getAllSelected(graph);
+                recordStoreQuery = GraphRecordStoreUtilities.getAllSelected(graph);
```
cygnus-x-1, line 120: I would make this one queryRecordStore, as it is a record store representing a query, not a query of a record store.

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/color/ConstellationColor.java
```diff
@@ -203,9 +203,9 @@ private ConstellationColor(final String name, final float red, final float green
                 && alpha >= 0f && alpha <= 1f;
 
         this.name = name;
-        this.red = red;
-        this.green = green;
-        this.blue = blue;
+        this.redColorValue = red;
```
cygnus-x-1, line 206: Same deal here as my previous comment.

CoreScriptingView/src/au/gov/asd/tac/constellation/views/scripting/graph/SGraph.java
formalhaut69: Yeah I'm not really sure. @arcturus2 do you think that its worth changing the variable names?

antares1470: I think the issue is that it can confuse users of the class. Yes their definitions are clearly defined but unless you happen to know that there are 2 different fields with the same name in the same class, there may be some confusion down the line when you use the wrong one and things aren't happening you expect them to.

Really, this is just poor coding practice. Hence this is a code smell and not something like a bug

arcturus2: Generally I'm a fan of following SonarQube though this particular scenario does seem a little overkill. I think though given the definition of `GRAPH` and `graph` is isolated to just this class it is worth resolving the code smell to avoid the chance someone does get confused. I think what this SQ rule is trying to address is important so happy to support it at the end of the day.

As always, nothing is set in stone for me so if there are any objections then happy to revisit it.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/templates/RecordStoreQueryPlugin.java
cygnus-x-1: Still waiting on this one.

PR #396, opened 2020-04-03
Updating error messages with expected default formats

CrucisGamma:
# Description of the Change

Expand the error message to include the required data format for data input that don't match various elements in the code 

### Alternate Designs

Modifying for everything including floats/integers was considered but I think most people know this already?

As a user, would be helpful to know what internal format will be used

### Benefits

As a user, would be helpful to know what internal format will be used

### Possible Drawbacks

Will need to be maintained if the internal format does end up changing

### Verification Process

- How did you verify that all new functionality works as expected?
Ran and tested this manually (for example, ingesting a new .csv with an incorrect datetime format)
- How did you verify that all changed functionality works as expected?
This is only modifying the existing error message
- How did you verify that the change has not introduced any regressions?
Ran existing tests

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportController.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportController.java
@@ -75,6 +75,7 @@ public class ImportController {
     private ImportFileParser importFileParser;
     private boolean schemaInitialised;
     private boolean showAllSchemaAttributes;
+    private String filterAttributes = "";
     private PluginParameters currentParameters = null;
 
     // Attributes that exist in the graph or schema.
@@ -170,7 +171,12 @@ public class ImportController {
     }
 
     public void setDestination(final ImportDestination<?> destination) {
-        this.currentDestination = destination;
+        if (destination != null) {
+            this.currentDestination = destination;
+        }
+        if (this.currentDestination == null) {
+            return;
+        }
 
         // Clearing the manually added attributes removes them when loading a template.
         // The destination is set with clearmanuallyAdded true before loading the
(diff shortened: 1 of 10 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportController.java
```diff
@@ -170,7 +171,12 @@ public void setClearManuallyAdded(final boolean b) {
     }
 
     public void setDestination(final ImportDestination<?> destination) {
-        this.currentDestination = destination;
+        if (destination != null) {
+            this.currentDestination = destination;
+        }
+        if (this.currentDestination == null) {
+            return;
```
antares1470, line 178: sonar will probably flag this return as unnecessary

arcturus2: Hi @CrucisGamma so @antares1470 has been smashing through SQ rules so any assistance to keep code smells down is greatly appreciated. What i'll do is merge it into a temporary branch  to see what SQ has to say. SQ can't run on your fork.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/RunPane.java
```diff
@@ -186,6 +191,19 @@ public RunPane(final ImportController importController) {
 
         attributePane.addRow(0, sourceVertexScrollPane, destinationVertexScrollPane, transactionScrollPane);
 
+        attributePane.setOnKeyPressed(event -> {
+            final KeyCode c = event.getCode();
```
arcturus2, line 195: :+1:

arcturus2:
Interesting @antares1470 so SQ didn't complain about the empty return  - I expected it would too. See #401. Going to merge #401.

Issue #375, opened 2020-03-27
Table View Column Selection - Search Columns


cygnus-x-1:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The column selection menu in the Table View becomes unwieldy to use when you have a lot of attributes available. This could be improved by allowing the user to search for attributes by name within this menu.

PR #398, opened 2020-04-03
Add feature to search / filter in Table View Column Selection

bentipping:
This addresses issue 'Table View Column Selection - Search Columns #375'

### Description of the Change
As outlined in issue 375 the column selection can become unwieldy when many attributes are used.  A blank analytic graph has more then a page of attributes. 
image
I've added in a simple search that allows the user to filter the attributes by name within the column selection allowing easier selection of attributes.

image
_Simple search form at the top of the attributes._

image
_An example of the filtering of attributes._

 This feature makes the column selection easier for the user.

### Verification Process

1. Open a graph
2. Open the Table View 
3. Click the column selection button
4. Enter search term in the search box

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Constellation Changes
 
+## 2020-04-03 Changes in April
+* Added search feature to Table View Column Selection 
+
 ## 2020-03-01 Changes in March 2020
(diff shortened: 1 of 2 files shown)


arcturus2:
Awesome thank you @bentipping !

Issue #343, opened 2020-03-23
Add ability to manually adjust time Data Acess View "Absolute Range"


serpens24:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Is it possible to modify the time fields when querying date/time in the absolute range field. Currently time can only be adjusted by clicking the up and down arrows. Can it be modified to also allow analysts to type a specific time in the box? for example it the minute is set on 00 and I want to query at minute 59, I have to press the up button 59 times.  

### Steps to Reproduce

1. Open the Data Acess View and note the **Absolute range** widget in the **Global Parameters** fold.
2. Confirm that using up/down arrows allow hour/minute/seconds to be manipulated
3. Confirm that it is not possible to directly modify number by clicking it and typing

**Expected behaviour:**

Time element can be clicked amn directly edited

**Actual behaviour:** 

Times cannot be entered directly

**Reproduces how often:**

100%

### Additional Information

N/A

serpens24:
By default the javafx spinner is set not to allow direct editing - the call 
`spinner.setEditable(true);
`
will allow editing to occur ... however the text widget does not limit input to digits, so further code is required to limit entered characters to digits only.

The code:
```
        TextFormatter<Integer> timeFormatter = new TextFormatter<Integer>(
        new IntegerStringConverter(), 0, filter);
        spinner.getEditor().setTextFormatter(timeFormatter);
```
will hook in a text formatter, howwver filter needs to be defined ... this will filter text input, the following will limit it to digits - the overall code would look like:

```
        NumberFormat format = NumberFormat.getIntegerInstance();
        UnaryOperator<TextFormatter.Change> filter = c -> {
            if (c.isContentChange()) {
                ParsePosition parsePosition = new ParsePosition(0);
                // NumberFormat evaluates the beginning of the text
                format.parse(c.getControlNewText(), parsePosition);
                if (parsePosition.getIndex() == 0 ||
                        parsePosition.getIndex() < c.getControlNewText().length()) {
                    // reject parsing the complete text failed
[...]

PR #415, opened 2020-04-08
Update DateTimeRangeInputPane.java

serpens24:
### Description of the Change

Set time spinners in DateTimeRangeInputPane to allow dsirect edit and put some controls in place to ensure only digits are allowed and that entered values are limited to valid range.

### Alternate Designs

nil

bug fix requested by end users to ensure they do not need to spin through hours/mins/secs to select correct time

### Benefits

time saving

### Possible Drawbacks

nil

### Verification Process

Opened DateTimeRangeInputPane in running application and confirm hour/min/sec are able to be manually entered in the Absolute range From and To widgets.
Confirm that trying to enter non numerical digits is prohibited (ie a-z, . - etc) 
Confirm that entering a value greater than maximum value, ie > 23 for hours, > 59 for min, sec is rejected and values revert to allowed maximum

### Applicable Issues

Issue #343


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/DateTimeRangeInputPane.java
```diff
@@ -580,6 +589,30 @@ private Pane createSpinner(final String label, final int min, final int max, fin
         final Spinner<Integer> spinner = new Spinner<>(min, max, 1);
         spinner.setPrefWidth(NUMBER_SPINNER_WIDTH);
 
+        // Create a filter to limit text entry to just numerical digits
+        NumberFormat format = NumberFormat.getIntegerInstance();
```
cygnus-x-1, line 593: finals?

serpens24: good spot !! Will drop an update when I get a few minutes.

PR #400, opened 2020-04-05
Add in type-to-filter attributes option for Import from CSV

CrucisGamma:
### Description of the Change

Modify the "Import CSV Dialog" to be able to filter the many attributes by typing in a partial match on a number of attributes names.

### Alternate Designs

I did have an extra menu item under the existing 'select all' checkbox but it felt clunky to use.  Simply typing in 'Lati' to get all the Latitude fields or "Geo" to get geo or Date to get the date ones feels a lot more intuitive.

### Why Should This Be In Core?

Looking and pecking through the 140+ 'All' fields feels exceedingly cumbersome to find the field you're interested in.

### Benefits

People (I) won't be frustrated by finding the one attribute they actually want to use

### Possible Drawbacks

A Corona pandemic will occur in early-mid 2020... Does anybody read this?  I don't think there are any.

### Verification Process

- How did you verify that all new functionality works as expected?
Have run it and tested the new functionality without any appreciable differences in performance.  Imported and used a sample CSV file of my choosing.

- How did you verify that all changed functionality works as expected?
Have reviewed the changes to the code to hopefully ensure nothing to outrageous is going to go in.  Have done 'Fix imports' and 'Format' on all files to ensure correctly formatted code.

- How did you verify that the change has not introduced any regressions?
Rebuilt code.

Basically Import -> Delimited file , and then started typing in things I wanted to filter on (geo, lat, long, datetime)

### Applicable Issues

Nil.

PR #397, opened 2020-04-03
Created generric JSON config file read/writer.

serpens24:
### Description of the Change

Moved logic that reads/writes JSON configuration files into a generic package in Core Functionality. (genericjsonio).
Use this functionality to support PArameterIOUtlities.java.
Now that a generic JSON config package exists, this can be wrapped up and used to support saving tableview column preferences - ie columns to display, their order, and any column identiifed for sorting.
Updated TableView functionlaity to include a preferences menu item containing a save and load preferences sub-menu. User can now save current table column state for future use - note, only the current table type (ie vertex/transaction) is saved ./ loaded.

Table preferecnes are saved into users constellation directory, within the TableViewPreferences subfolder. The name of the config file saved is prefixed with an indication of the type of data being displayed in the table, however this is hidden from the user.

Additionally updated TableView code to store sort order - which was previously being lost when columns ere manually reordered.

### Alternate Designs

nil

### Why Should This Be In Core?

Provides ability to precan table layouts for specific purposes.

### Benefits

As above

### Possible Drawbacks

I'm not aware of any.

### Verification Process

1. Open a new graph as well as tableview. Add a selection of nodes and transactions to the graph using draw mode (or some other method).
2. Ensure transactions are being displayed in table and update the displayed columns to include non-standard columns. Sort the data by one of the columns
3. Click the heart icon and select 'Save Table Preferences'
4. On the save dialog enter a name - ie 'test' and click Ok.
5. Modify the displayed columns and their order, change the column being sorted by and the direction.
6. Click on the heart icon and select "Load Table Preferences". Select the entry in the list corresponding to the save name entered at step 4 and press the cancel button. **Confirm that no changes are made to the table**.
7. Click on the heart icon and select "Load Table Preferences". Select the entry in the list corresponding to the save name entered at step 4 and press the OK button.  **Confirm that the columns, their order and sorting reflect the changes saved at step 4**
8. Change the table to display vertexes.
9. Make changes to the displayed columns, their order and sorting.
10. Repeat steps 3-7 and confirm functionlaity works for vertexes as it does for transactions. **use the same filename used above, to confirm this is possible**
11. Click on the heart icon and select "Save Table Preferences". In the save dialog hit enter without entering a filename.
[...]

Issue #387, opened 2020-04-01
Resolve NB11 JDK11 Graph Offset Issue


aldebaran30701:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
The migration to NB11 is nearly complete, with just one issue which effects JDK11.
#11 - Awaiting response from JOGL Developers to develop a fix in the next release.

Issue 1374 is the issue relating to the graph offset issue.
https://jogamp.org/bugzilla//show_bug.cgi?id=1374#c3

Tracking for next release cycle of JOGL. 
https://jogamp.org/wiki/index.php?title=SW_Tracking_Report_Feature_Objectives_Overview#High_DPI

Closing #219 to make this the new master ticket for the remainder of NB11/JDK11 migration.
### Steps to Reproduce

See Ticket #11 

**Expected behaviour:** 

**Actual behaviour:** 

**Reproduces how often:**

### Additional Information
N/A

arcturus2:
I've had a go to build JOGL and add the patch mentioned in http://forum.jogamp.org/canvas-not-filling-frame-td4040092.html#a4040210 but didn't have any luck seeing a change. Here is a write up of what I did in case anyone can try and has better luck.

I am actually developing on a Mac whilst working on this problem and building JOGL turned out to be pretty straightforward.

If you are new to Constellation the you will need to follow the `README.md` to make sure you can compile and run Constellation.

Note that NetBeans 11.3 for the MacOSX was downloaded from http://apache.mirror.serversaustralia.com.au/netbeans/netbeans/11.3/Apache-NetBeans-11.3-bin-macosx.dmg and the JDK from https://cdn.azul.com/zulu/bin/zulu11.37.17-ca-fx-jdk11.0.6-macosx_x64.zip

Once you have Constellation building and running then proceed to the steps mentioned in https://jogamp.org/jogl/doc/HowToBuild.html

The prerequisites mention:

> MacOS and iOS Intel and Arm64
> git ≥ 2.11 provided by Xcode ≥ 8.3.3
[...]

Issue #15, opened 2019-07-19
replace jhall-2.0_05.jar and org-netbeans-modules-javahelp.jar with suitable JARs from Maven central


imranraza83:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description
Please check https://github.com/constellation-app/constellation/issues/3 for detailed beckground.

Names of jhall-2.0_05.jar and org-netbeans-modules-javahelp.jar are hard coded in NetBeans11 source code therefore it not possible to replace then at for NetBeans version currently being used.

### Steps to Reproduce

1. Get a fresh copy of NetBeans11 and open up constellation project
2. Remove following entries from buil.xml

    `<target name="-download-javahelp-file" >`
        `<mkdir dir="${ivy.lib.dir}"/>`
        `<echo message="installing netbeans help files to ... ${nbplatform.default.netbeans.dest.dir}"/>`
        `<get src="https://github.com/rsabhi/NBhelp/raw/master/org-netbeans-modules-javahelp.jar"
             dest="${nbplatform.default.netbeans.dest.dir}/platform/modules/" usetimestamp="true"/>`
    `</target>`
 `<target name="-download-javahelp-indexfile" >`
        `<mkdir dir="${ivy.lib.dir}"/>`
        `<echo message="installing netbeans help index files... ${nbplatform.default.netbeans.dest.dir}"/>`
        `<get src="https://github.com/rsabhi/NBhelp/raw/master/jhall-2.0_05.jar"
             dest="${nbplatform.default.netbeans.dest.dir}/platform/modules/ext/" usetimestamp="true"/>`
    `</target>`

3. Build project

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

aldebaran30701:
As this ticket is specific to the issue of no help - I feel #430 covers this. 
Closing this issue, and future work will be related to #430.

PR #417, opened 2020-04-09
javahelp-update

cygnus-x-1:
# Description of the Change

The javahelp netbeans module and jhall jars are no longer downloaded from the thrid-party-jars repository in the ant build script, instead the javahelp (updated jhall) jar is retrieved as a regular maven dependency, and the updated javahelp netbeans module included with the Netbeans 11.3 distribution should be used.

Dependency versions have been been updated to latest.

A temporary scaling fix has been put in place to resolve the JOGL 2.4.0 HIDPI scaling issues (this should be removed when JOGL 2.4.0 is released on maven as it will have an internal fix for this).

Minor improvements to recent changes to the Table View, Delimited File Importer and JsonIO.

### Verification Process

Tested JavaHelp changes using a fresh Netbeans installation and Ivy cache on two different systems.

Tested HIDPI fix by opening graphs on a HiDPI display (Surface Book 2).

Manual testing of Table View column filter and Delimited File Importer attribute filter.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/15
https://github.com/constellation-app/constellation/issues/387
https://github.com/constellation-app/constellation/pull/398
https://github.com/constellation-app/constellation/pull/400
https://github.com/constellation-app/constellation/pull/397


cygnus-x-1:
@canis-majoris-42 could you confirm this works on your HIDPI display?

cygnus-x-1:
@arcturus2 could you confirm this is correctly building distributions?

arcturus2:
> @arcturus2 could you confirm this is correctly building distributions?

@cygnus-x-1 can you change line 215 in `build.xml` to the folder `distribution`? Sorry I'm asking you to fix my bug. With that change it opens without an error. Thank you.

```java
            <zipfileset dir="${basedir}/distribution/" prefix="${app.name}/"/>
```

arcturus2:
@cygnus-x-1 FYI so the Travis build is failing here because it has the work around in place with the replaced jars. That would need to be corrected which I am happy to do straight after this gets merged when we get to that point.

arcturus2:
Also the README needs to be updated to remove those workaround steps and also possibly a note to devs that previously did the workaround to revert it!

cygnus-x-1:
Okay I believe this is ready to be merged now @arcturus2.

arcturus2:
@cygnus-x-1 I merged this PR before testing it on the Mac. The work around code block has made a difference but it hasn't fixed it.
* Nodes are now being created the camera if that makes sense and I have to zoom out to see them.
* Selection no longer works probably because it is also happening behind the camera
* The x,y,z graph legend on the top right corner also isn't visible.

I'll continue having a look at modifying that code block to see if I can make it work on the Mac.

PR #420, opened 2020-04-09
CSS and other sonar fixes

antares1470:
### Description of the Change

As suggested by sonar:
- added missing generic font family declaration to CSS
- CSS shorthand properties no longer override longhand properties
- removed jump statements from finally blocks
- ensured it wasn't possible to divide by 0
- renamed getters to reflect variable they are getting


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReport.java
```diff
@@ -100,7 +100,7 @@ public PluginReport addPluginReport(Plugin plugin) {
      * tags recognized by this GraphReport.
      *
      * @param pluginReport the specified PluginReport.
-     * @see GraphReport#getTags()
+     * @see GraphReport#getUTags()
```
cygnus-x-1, line 103: Whats up with all these `getUTags` replacements? That doesn't look to be a real method...

arcturus2: I think your correct, `getUtags` doesn't seem to exist.
It would be good to get clarity for the reason behind this change. What is the SQ rule @antares1470 ?
Also if we are going to rename methods can you please update `CHANGELOG.md` @antares1470 ?

antares1470: @arcturus2 @cygnus-x-1 check out line 127. The sonar rule was about ensuring that the names of getter functions matched the field they were getting. In this case, there is a field called tags and a field called uTags (the unmodifiable version of tags) and so with the function getTags returning uTags rather than tags, the sonar rule flagged.

antares1470: Changelog has now been updated

cygnus-x-1: I don't like this change - the whole `uVariable` convention seems silly to me. I get and support using modifiable collections where they are not expected to ever change, but why duplicate the variables and store them both in memory? I'd just set `variable` to the unmodifiable collection - thoughts @arcturus2?

arcturus2: @cygnus-x-1 agreed regarding the increase in memory footprint point.

So we have been using this convention is some classes and I was interested in continuing it :smile:. The obvious benefit is that it is thread safe.

I think you will always need both (modifiable and unmodifiable) because if you were to only have the unmodifiable version available to the class then you can't add anything to it? If there is a way then happy to learn.

antares1470: @arcturus2 @cygnus-x-1 What is the preference then for this change?

cygnus-x-1: As I understand it, this patter is only used where lazy loading happens, and the objects loaded never change after that - so thread safety and additions are non-issues.

arcturus2: @cygnus-x-1 so do you mean to say `tags` won't change after Constellation opens? That's not my understanding though - I thought tags will continue to get added as different plugins run which append respective tags to that `ArrayList`.

antares1470:
@arcturus2 I have also now included a changelog entry as requested in #428

Issue #45, opened 2019-07-26
Make the map view mouse controls the same as the graph


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a user
I would like consistent mouse controls between the map view and graph
so that i don't need to context switch all the time

The map view mouse controls are the polar opposite to the graph view which is a terrible user experience.

Reason for the current issue:

Unfortunately the library we're using to render the map enforces that control scheme, ie. doesn't allow you to change it. The library is an open source project, so in theory we could work around this by contributing code to their project enabling custom control schemes - but this would be a non-trivial task.

### Steps to Reproduce

1. Open the Map View
2. Left click and drag a box
3. Selected markers highlight

**Expected behaviour:** Markers in the selection are highlighted

**Actual behaviour:** The map pans

**Reproduces how often:** 100%

### Additional Information

Start by making a fork of Unfolding Maps and get the new controls to work. If things work OK then consider sending a PR to the Unfolding Maps project.

PR #423, opened 2020-04-11
Change feature in Map View to invert the mouse click controls.

bentipping:
This makes the functionality consistent with the CONSTELLATION graph.  Addresses current issue #45

### Description of the Change
This PR addresses the issue (#45 ) of inverted mouse controls on the Map View when compared to the graph view.  Coming from the graph view the user would expect a right click and drag to pan and a left click and drag to select markers on the map, in reality this is inverted.  This PR fixes this to be consistent with the graph view.

When the user left clicks and drags the markers are selected. (It's hard to tell that it's a left click but trust me :))
image

When the user right clicks and drag the map will pan to the new position ( you may also have to trust me that I'm using the right click).  The other map features function as they were before such as the zoom on on scroll and left click zoom and the measurement tools.

### Alternate Designs
The alternate design was to add the inverted mouse controls as an optional feature to the open source map library, Unfolded, and make use of this change in Constellation.  Due to the following reasons I decided against adding to Unfolded and added to Constellation instead:

- Inactivity of Unfolded 

- The feature is specific to Constellation with no open issue on Unfolded.

- Unfolded API allows the controls to inverted with minimal duplication (a couple of lines)     

This adds to the user experience giving a consistent user experience across the application.  

### Verification Process

1) Open Map View 
2) Left click & drag to select markers
3) Right click and drag to pan the map


arcturus2:
Wow thanks @bentipping :tada: we've been struggling to get this change done and we thought the only way we could make this work was to make a change in Unfolding maps!

Your change is the ideal solution and I really appreciate this contribution. The opposite mouse controls was frustrating our users when switching between the graph and map view.

The only feedback I have is that we should update `whatsnew.txt` or `CHANGELOG.md` depending on the type of change. So `whatsnew.txt` shows end users things that have changed and this is what you see in the What's New section when you open Constellation. In this case we should add something to suggest that the mouse controls have been fixed :smile:.

FYI `CHANGELOG.md` is focused more for developers and those integrating Constellation with their modules. A method signature change, methods added or removed etc should be documented in `CHANGELOG.md`.

I suspect a lot of people are not reading What's new but the reason I continue to want this updated is because over time it shows release notes of user features.

I'm happy to give you time to update this PR with a what's new entry; otherwise I'll merge this and put something in. Cheers again!

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
```diff
@@ -347,7 +347,9 @@ public void setup() {
         map.setTweening(true);
 
         dispatcher = MapUtils.createDefaultEventDispatcher(this, map);
-
+        // The map library, Unfolded, defaults to a hard-coded left click pan
```
cygnus-x-1, line 350: Replace 'Unfolded" with "Unfolding Maps"

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
```diff
@@ -468,6 +459,27 @@ public void mouseDragged(final MouseEvent event) {
...
+            //Checks to see if click was on the map
+            final boolean ignoreMapInteractions = overlays.stream().anyMatch(overlay
+                    -> overlay.isEnabled()
+                    && mouseX > overlay.getX() && mouseY > overlay.getY()
+                    && mouseX < overlay.getX() + overlay.getWidth()
+                    && mouseY < overlay.getY() + overlay.getHeight());
+            if (!ignoreMapInteractions) {
+                // Triggers a pan event for a right click & drag if within map
+                dispatcher.register(map, PanMapEvent.TYPE_PAN, map.getId());
+                Location oldLocation = map.getLocation(pmouseX, pmouseY);
```
cygnus-x-1, line 474: Add finals.

PR #424, opened 2020-04-11
Optimise the download location of artifacts

arcturus2:
### Description of the Change

The location of jars in `third-party-dependencies` was being redirected to `raw.githubusercontent.com` so this uses that url instead.


build.xml
```diff
@@ -62,77 +62,77 @@
     <target name="-download-unfolding-maps">
         <mkdir dir="${ivy.lib.dir}"/>
         <echo message="installing unfolding-maps..."/>
-        <get src="https://github.com/constellation-app/third-party-dependencies/raw/master/unfolding-master-20171010.jar"
+        <get src="https://raw.githubusercontent.com/constellation-app/third-party-dependencies/master/unfolding-master-20171010.jar"
```
cygnus-x-1, line 65: i'd assume you don't need the `raw=true` parameter for all these urls now?

aldebaran30701: Seems to download without it - Just checking through a browser request to that URL

arcturus2: yeah seems to be ok probs because the domain is `raw.githubusercontent.com`. This was what GitHub redirected it to so I was just putting that in.

build.xml
```diff
@@ -62,77 +62,77 @@
...
         <echo message="installing gluegen..."/>
         <get dest="${ivy.lib.dir}" usetimestamp="true">
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt.jar?raw=true"/>
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt-natives-linux-amd64.jar?raw=true"/>
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt-natives-linux-i586.jar?raw=true"/>
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt-natives-macosx-universal.jar?raw=true"/>
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt-natives-windows-amd64.jar?raw=true"/>
-            <url url="https://github.com/constellation-app/third-party-dependencies/blob/master/JOGL%202.4.0/gluegen-rt-natives-windows-i586.jar?raw=true"/>
+            <url url="https://raw.githubusercontent.com/constellation-app/third-party-dependencies/master/JOGL%202.4.0/gluegen-rt.jar?raw=true"/>
```
cygnus-x-1, line 73: also all these... and in fact all the urls you changed.

arcturus2: Ahh ok let me check all of this again. Thanks for pointing it out.

arcturus2: I've reworked this PR and happy for it to be reviewed again @cygnus-x-1

Issue #223, opened 2020-02-03
Layers View


arcturus2:
This is a new view that can work similar to other applications like Adobe Photoshop, Gimp or Visio.

The idea would be to allow a user to slide a graph into layers so that analysis can be done in context of the graph with the ability to dim/hide layers and show others. The current method is to select the part of the graph you are interested in and then open that in a new graph (Ctrl +U). This could be avoided by allowing the user to stay in context of the entire graph but focus on answering one question.

This ticket has a large scope and not everything needs to be done to start with. Based on what is left over, new issues can be created to track them.

Features are:
 * ability to name layers
 * ability to show/hide layers
 * ability to add a selection to a layer
 * nodes and transactions can be part of multiple layers
 * automatically generate layers based on an attributes (e.g transaction types)

aldebaran30701:
Some documentation about Layers View usage. 

**Functionality:**
Add and remove elements of a graph to a layer.
Add via manual selection, or by entry of a query to allow dynamic addition of an element.

**Usage:**
Layers view functionality is accessible from two locations;
 - Right click context menu on an element
 - Experimental -> Views-> Layers View (Main UI View)

The right click context menu allows selected elements to be added or removed from a specified layer.
To do so, have an element selected and right click -> Add to layer -> enter in value from 2-32. Layer 1 is the default show all graph.
![Add to layer](https://user-images.githubusercontent.com/58677759/79290800-309aae00-7f0c-11ea-93bf-380419d80ff3.gif)


The Layers View lists all layers stored on the graph. 
The checkbox marks the visibility of that layer. 
The query input is for entering a query to dynamically show a layer based on element attributes.
The description input can be used to describe the output of the query.
![Toggle Layers](https://user-images.githubusercontent.com/58677759/79291049-c59da700-7f0c-11ea-9f02-9b288d3d215c.gif)

The query will be structured as below

**String comparison:**
Label == Vertex #0<Unknown>
Label != Vertex #0<Unknown>

**Float comparison:**
visibility == 0.0
visibility != 0.0
visibility > 0.0
visibility < 1.0

**Boolean comparison:**
dim == false
dim != true

With conjunction and disjunction of queries, ( ) must surround each query.
Separated by only the boolean operators || or && ie. (Query1)||(Query2)

**Disjunction**
(Label == Vertex #0<Unknown>)||(dim != false)

**Conjunction**
(Label == Vertex #0<Unknown>)&&(dim != false)

[...]

PR #434, opened 2020-04-16
Layers view heading alignments

aldebaran30701:
### Description of the Change
Align headings of layers view UI

### Alternate Designs
N/A

Displays correctly now

### Benefits
Easier usage

### Possible Drawbacks
N/A

### Verification Process
Test multiple view positions

### Applicable Issues
#223


arcturus2:
Nice work @aldebaran30701. I'm going to merge this in with the v2 beta I am just about to publish so merging it in. I did notice that the 2nd column "Visibility" doesn't stretch like the others so on my laptop its showing as "Visib...".

cygnus-x-1:
Good stuff - longer term i think you should make the `layers` a GridPane and have common alignment for the headings and layer info that way.

aldebaran30701:
> I did notice that the 2nd column "Visibility" doesn't stretch like the others so on my laptop its showing as "Visib...".

I had a look with scaling on windows, but couldn't get it to happen. I can add some extra padding to allow for devices with higher or lower res if that is what would have caused that. 
resize

arcturus2:
It might be isolated to Linux then given this is from my Linux laptop

Peek 2020-04-16 16-45

PR #437, opened 2020-04-17
Optional values and some python changes

antares1470:
### Description of the Change
As suggested by sonar:
- no longer call get() on optional variable without checking whether value exists
- cast the value of parameter of calls to Double.longBitsToDouble as a double
- handled duplicate string literals in both python and java (ones that were missed previously)
- added comments to empty python methods


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/translator/DateAttributeTranslator.java
```diff
@@ -50,7 +50,6 @@
         DATE_FORMATS.put("MM/dd/yyyy", "MM/dd/yyyy");
         DATE_FORMATS.put("dd/MM/yyyy", "dd/MM/yyyy");
         DATE_FORMATS.put("yyyyMMdd", "yyyyMMdd");
-        DATE_FORMATS.put("yyyyMMdd", "yyyyMMdd");
```
arcturus2, line 53: :+1:

PR #451, opened 2020-04-22
Null Pointer when right-clicking on graph

formalhaut69:
### Description of the Change
Resolved the null pointer exception when right clicking on a graph


CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/utilities/LayersContextMenu.java
```diff
@@ -50,7 +51,7 @@
         if (elementType == GraphElementType.VERTEX || elementType == GraphElementType.TRANSACTION) {
             return Arrays.asList(ADD_TO_LAYER, REMOVE_FROM_LAYER);
         } else {
-            return null;
```
aldebaran30701, line 53: This may have been between the time I was developing the context menu and the SonarQube fixes were added in. 
My mistake in not remaining consistent and returning an empty list in this case. Great pickup @formalhaut69

arcturus2: All good @aldebaran30701  and thanks for fixing it @formalhaut69 !

PR #446, opened 2020-04-21
Rule (java) switch cases should end with an unconditional break statement

formalhaut69:
### Description of the Change
Added a few breaks to resolve this issue.

This change also contains fixes for "Rule (Java) Methods should not be empty" and should resolve around 200 critical issues. SonarQube recommended that the empty methods should either be removed or have a nested comment explaining why it was left empty. The vast majority just required a comment but I have removed some empty methods that were completely unused and an unused class.

PR #452, opened 2020-04-22
Fix some bugs running Export Glyph Textures

arcturus2:
### Description of the Change

I looked at this bit of code during my review of #446 and couldn't help but fix the following:
* Fixed some exceptions seen running Experimental -> Glyph Textures
* Removed some comments no longer used
* Added some finals

### DIFF
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
@@ -148,7 +148,7 @@ public final class SharedDrawable {
                     try (final OutputStream out = new FileOutputStream(outputFile)) {
                         glyphManager.writeGlyphBuffer(page, out);
                     } catch (IOException ex) {
-                        // Nothing to do here - this is just for developer testing.
+                        LOGGER.severe(ex.getLocalizedMessage());
                     }
                 }
             });
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
@@ -145,7 +145,7 @@ public final class SharedDrawable {
 
                 for (int page = 0; page < glyphManager.getGlyphPageCount(); page++) {
                     final File outputFile = new File(baseFileName + SeparatorConstants.UNDERSCORE + page + ".png");
-                    try ( OutputStream out = new FileOutputStream(outputFile)) {
+                    try (final OutputStream out = new FileOutputStream(outputFile)) {
                         glyphManager.writeGlyphBuffer(page, out);
                     } catch (IOException ex) {
                         // Nothing to do here - this is just for developer testing.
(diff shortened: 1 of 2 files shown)


CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
```diff
@@ -129,28 +122,39 @@ private static void init() {
...
+                final int extensionStart = baseFileName.lastIndexOf('.');
+                if (extensionStart > 0) {
+                    baseFileName = baseFileName.substring(0, extensionStart);
                 }
-            }
-        });
+
+                for (int page = 0; page < glyphManager.getGlyphPageCount(); page++) {
+                    final File outputFile = new File(baseFileName + SeparatorConstants.UNDERSCORE + page + ".png");
+                    try ( OutputStream out = new FileOutputStream(outputFile)) {
```
cygnus-x-1, line 148: out can be final, also remove the rogue space.

formalhaut69:
@arcturus2 I've tested the exportGlyphTextures method from this branch and this image is the output that I get. I'm not really sure what this method is meant to do but if this output is correct then I'm happy with these changes.

image

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
```diff
@@ -129,28 +122,39 @@ private static void init() {
...
+                        glyphManager.writeGlyphBuffer(page, out);
+                    } catch (IOException ex) {
+                        // Nothing to do here - this is just for developer testing.
```
antares1470, line 151: is the intention to remove this once developer testing is done?

cygnus-x-1:
> @arcturus2 I've tested the exportGlyphTextures method from this branch and this image is the output that I get. I'm not really sure what this method is meant to do but if this output is correct then I'm happy with these changes.
> 
> image

That looks correct - the function should create an image representing the texture map of glyphs sent to the graphics card which it then uses to render labels in a graph. It looks like that because it uses an algorithm to tightly pack the characters in order to fit as much as possible into a single texture map, to reduce data transfer to/from the graphics card. @algol60 will be able to describe this better than me.

formalhaut69:
> 
> 
> > @arcturus2 I've tested the exportGlyphTextures method from this branch and this image is the output that I get. I'm not really sure what this method is meant to do but if this output is correct then I'm happy with these changes.
> > image
> 
> That looks correct - the function should create an image representing the texture map of glyphs sent to the graphics card which it then uses to render labels in a graph. It looks like that because it uses an algorithm to tightly pack the characters in order to fit as much as possible into a single texture map, to reduce data transfer to/from the graphics card. @algol60 will be able to describe this better than me.

Cool, thanks for the explanation @cygnus-x-1

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
arcturus2: Good pickup regarding the `final` and rogue space. I added the `final` but the space seems to be part of the auto-formatter. I'd say that's a NB 11.3 bug.

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/SharedDrawable.java
arcturus2: I'd like to keep this plugin around - so I've replaced the comment with a `logger` call.

arcturus2:
@cygnus-x-1 I'm going to merge this one given I believe I've addressed the minor feedback, cheers.

PR #453, opened 2020-04-23
RULE (Java) Null should not be returned from a "Boolean" method

sol695510:
### Description of the Change
Changed methods of return type boolean to only return boolean expressions, as some cases were present where a null could be returned.

### Alternate Designs
N/A

Resolves code smell identified by SonarQube.

### Benefits
Reduces code complexity.

### Possible Drawbacks
N/A

### Verification Process
Executed Constellation and attempted to open, draw to, create new and save graphs to observe if any functionality was lost or new bugs were introduced. Application still functioned as normal.

### Applicable Issues
#280

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
@@ -2133,10 +2133,7 @@ public class StoreGraph extends LockingTarget implements GraphWriteMethods, Seri
         expression = expression.substring(0, expression.length() - 1);
 //        System.out.println("Before evaluation: " + expression);
 
-        final Boolean result = QueryEvaluator.evaluatePostfix(expression);
-//        System.out.println("Result calculated: " + result);
-
-        return result;
+        return QueryEvaluator.evaluatePostfix(expression);
     }
 
     // update the layer visibility of an element
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
@@ -2136,7 +2136,7 @@ public class StoreGraph extends LockingTarget implements GraphWriteMethods, Seri
         final Boolean result = QueryEvaluator.evaluatePostfix(expression);
 //        System.out.println("Result calculated: " + result);
 
-        return result == null ? false : result;
+        return result;
     }
 
     // update the layer visibility of an element
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/attribute/BooleanObjectAttributeDescription.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/attribute/BooleanObjectAttributeDescription.java
@@ -47,7 +47,7 @@ public final class BooleanObjectAttributeDescription extends AbstractObjectAttri
     @SuppressWarnings("unchecked") // Casts are manually checked
     protected Boolean convertFromObject(final Object object) {
         if (object == null) {
-            return null;
+            return false;
         } else if (object instanceof Number) {
             return ((Number) object).longValue() != 0;
         } else if (object instanceof Boolean) {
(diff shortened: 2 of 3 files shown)


CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/attribute/BooleanObjectAttributeDescription.java
```diff
@@ -47,7 +47,7 @@ public BooleanObjectAttributeDescription() {
     @SuppressWarnings("unchecked") // Casts are manually checked
     protected Boolean convertFromObject(final Object object) {
         if (object == null) {
-            return null;
```
cygnus-x-1, line 50: This was actually an intentional design choice. A boolean object attribute has `null` as its default value, and we take that to essentially mean "no value" as oppose to false.

sol695510: Just wondering what that does as opposed to returning false. In other words where was that being checked for to determine different outcomes for cases of false or null?

Asking because the one time I found where it checked for null was in an inline "if" where if the returned value was null it would return false. (Its one of the four changes made on this pull request, not sure how to link code blocks from the code changes into comments.)

antares1470: To summarise the rule, it basically doesn't like returning null because Boolean can be confused with its primitive counterpart boolean (which can't return null) and so can cause problems with callers code. Based on that, it sounds like we could decide to ignore the rule if we understand the distinction.

On the other hand, lines like the one @sol695510 fixed with "bool == null ? false : bool" looks a bit silly and it would probably be ideal to remove the need for such lines of code

sol695510: Should I just revert the code block mentioned above back to returning a null and keep the others as is?

cygnus-x-1: Yes, I think so.

cygnus-x-1: To answer your previous question - the Attribute Editor, for example, will display "false" if the value of a boolean object attribute is false and "<No Value>" if it is null. This allows a user to distinguish between, for example, a "false" value retrieved from a database and no value having been set (no data, or no database query performed).

sol695510: > To summarise the rule, it basically doesn't like returning null because Boolean can be confused with its primitive counterpart boolean (which can't return null) and so can cause problems with callers code. Based on that, it sounds like we could decide to ignore the rule if we understand the distinction.
> 
> On the other hand, lines like the one @sol695510 fixed with "bool == null ? false : bool" looks a bit silly and it would probably be ideal to remove the need for such lines of code

> To answer your previous question - the Attribute Editor, for example, will display "false" if the value of a boolean object attribute is false and "" if it is null. This allows a user to distinguish between, for example, a "false" value retrieved from a database and no value having been set (no data, or no database query performed).

Thanks @antares1470 and @cygnus-x-1. I'll make the changes.

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
```diff
@@ -2136,7 +2136,7 @@ private boolean evaluateLayerQuery(final GraphElementType elementType, final int
         final Boolean result = QueryEvaluator.evaluatePostfix(expression);
 //        System.out.println("Result calculated: " + result);
 
-        return result == null ? false : result;
+        return result;
```
antares1470, line 2139: Actually, instead of returning result, just return `QueryEvaluator.evaluatePostfix(expression)` and remove the need for the result variable

PR #454, opened 2020-04-23
Layers View State

aldebaran30701:
### Description of the Change
This Pull Request includes the addition of a State attribute and all relevant handlers to keep it up to date across multiple open graphs.
A graph now has an attribute description for the Layers View State. This state stores the following:
 - Id of the layer
 - Boolean value if it was viewable
 - String value of the query
 - String value of the Description

This implementation allows one graph to show a different set of selected layers compared to another graph's layers view. 
The default state will be the standard default layer, as well as layer 2 expanded. This allows the user to intuitively use that open layer promptly.

### Alternate Designs
N/A
Current implementation maps closely to Analytic view state usage.

Because it allows a graph to have layer preferences saved.

### Benefits
Allows layer preferences to be saved to a graph
Allows user to switch between multiple graphs and have the layers pre-filled.

### Possible Drawbacks
Slightly slower speed while loading the layers view 0.1s approximately

### Verification Process
**Layers view open**
 - with graph (with state saved)
     - select checkboxes + new layer 
 - with graph (without state)
     - select checkboxes + new layer 
 - without graph
     - select checkboxes + new layer 

**Testing opening of layers view**
 - with graph (with state saved)
 - with graph (without state)
 - without graph

**Many graphs open**
 - Switching back and forth between many graphs.
 - Testing reload of 32 layers

### Applicable Issues
#223


CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -86,13 +96,116 @@ public void submit() {
...
+        }
+
+        @Override
+        public void edit(final GraphWriteMethods graph, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException, PluginException {
+            if (pane == null) {
+                return;
+            }
+            int stateAttributeId = LayersViewConcept.MetaAttribute.LAYERS_VIEW_STATE.ensure(graph);
+            // no state or on a graph update
+            if (graph.getObjectValue(stateAttributeId, 0) == null || isUpdateCall) {
```
cygnus-x-1, line 146: Since the state object is used in both the if and else blocks, you should get a reference to it above.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -86,13 +96,116 @@ public void submit() {
...
+    /**
+     * Update the display by reading and writing to/from the state attribute.
+     */
+    private static final class LayersViewStateUpdater extends SimpleEditPlugin {
+
+        private LayersViewPane pane = null;
+        private boolean isUpdateCall = false;
+
+        public LayersViewStateUpdater(final LayersViewPane pane, final boolean isUpdateCall) {
+            this.pane = pane;
```
cygnus-x-1, line 135: You don't need to store `pane` within the plugin here - you should just be able to call `parent.getContent()` in the `captureState` method. That being said, rather than processing the pane within the plugin at all, I think it would make more sense for this class to offer `getLayers` and `setLayers` methods which operate on a list of `LayerDescription` objects. That way you could break the plugin away from this class (and add it to a registry).

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -86,13 +96,116 @@ public void submit() {
...
+            }
+            int stateAttributeId = LayersViewConcept.MetaAttribute.LAYERS_VIEW_STATE.ensure(graph);
+            // no state or on a graph update
+            if (graph.getObjectValue(stateAttributeId, 0) == null || isUpdateCall) {
+                // Take a snapshot of the UI and store that state
+                LayersViewState newState = captureState();
+                graph.setObjectValue(stateAttributeId, 0, newState);
+            } else {
+                // When a state exists, write it to the Layers View
+                updateLayersView(graph.getObjectValue(stateAttributeId, 0));
```
cygnus-x-1, line 152: You should perform this operation in response to a change to the state attribute using a listener in `LayersViewTopComponent`.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/state/LayersViewConcept.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Attribute provider for attributes specific to the Layers View.
+ *
+ * @author aldebaran30701
+ */
+@ServiceProvider(service = SchemaConcept.class)
+public class LayersViewConcept extends SchemaConcept {
+
+    private static final LayersViewState DEFAULT_STATE = new LayersViewState(
```
cygnus-x-1, line 39: It might make more sense to move this to the `LayerViewState` class.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
aldebaran30701: I tried to call parent.getContent(); within the plugin but it would not allow me because of the static context of the plugin. I'm in the process of cleaning a heap up - thanks to your suggestions it should be a load simpler.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/state/LayersViewConcept.java
aldebaran30701: I can see how that could be a beneficial change, and I was thinking the same thing myself. With the current implementation I've just copied the same structure as AnalyticViewState and -Concept. I may be wrong but a change could make it inconsistent with the rest of the state implementations.

arcturus2:
Just an observation is that trying to open the layers view from a Mac crashes Constellation and it just exists. Opening the Analytic View hangs forever so it's more an issue with the Mac rather than your view.

Data Access view works fine so its really strange.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
arcturus2: @aldebaran30701 so should I wait before reviewing this PR again?

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/state/LayersViewConcept.java
cygnus-x-1: Haha i'm pretty sure there's almost no consistency to the state implementations right now - i'll let you decide what's better and we'll go with that for now.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
aldebaran30701: Yes, I will ping you both when the fixes are all pushed up. 👍

aldebaran30701:
@arcturus2 @cygnus-x-1 
This is now out of draft and ready for review. 
A few things to note - I decoupled the view from the state by passing a list of LayerDescription objects to handle. 
Also I've included the Concept classes for both the Layers view and Analytic views inside of the respective state classes. As it's related to the state, it makes sense to keep them together.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/state/LayersViewConcept.java
aldebaran30701: I've implemented this in the current commit.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
aldebaran30701: Also added in the new commits

aldebaran30701:
>I then proceeded to create a new graph and all the nodes are not visible (before adding any layers).

@arcturus2 I have just checked this and you are correct, It doesn't function as intended. I've checked on another branch and it appears that the problem was there from the last layers view fix.

As this state only adds functionality, and does not break any existing code (That we know of) I propose the following:
 - That we merge this state functionality in.
 - That i get a PR fix in for the Layers View issue you outline above.

Let me know how this sounds - I'm happy to hear other ideas also

arcturus2:
I'm ok with that approach @aldebaran30701 - long running PRs can be a pain.

@cygnus-x-1 I see you've been added as a reviewer but not approved this yet - ok for me to merge this?

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
```diff
@@ -90,17 +98,55 @@ public void clearAnalyticQuestions() {
...
         }
     }
+
+    /**
+     * Attribute provider for attributes specific to the Analytic View.
+     *
+     * @author cygnus_x-1
+     */
+    @ServiceProvider(service = SchemaConcept.class)
+    public static class AnalyticViewConcept extends SchemaConcept {
```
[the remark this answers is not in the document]
aldebaran30701: I may have misunderstood your previous comment listed here : https://github.com/constellation-app/constellation/pull/454#discussion_r413588608

My interpretation of it was that it made more sense to add it to that specific class's state class.
By doing so I feel like maybe you were in actual fact - talking about having the code inside the .java file but not making it an inner class. (if my description is possible)

This change may be as a result of my misunderstanding of "move this to the `LayersViewState` class"

aldebaran30701:
@arcturus2  I've included the fix to the issue you mentioned about the visibility of the elements always being invisible. I must have gotten rid of a line of code or refactored something incorrectly in the previous PR to cause this.

@cygnus-x-1 I've added a comment about the State Concept being within the State class - Let me know if a misunderstanding was the case and I can get a fix in for that inner class. 
Also, I abstracted the controller grabbing the whole pane to fetch the `List<LayerDescription>` - does that satisfy your comment regarding the getter and setter pattern for layers? 
I tried again regarding the static reference using `parent.getContent()` within the plugin and ran into the same issue.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/state/AnalyticViewState.java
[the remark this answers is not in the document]
cygnus-x-1: Ahh yep I should have been clearer - I was suggesting you move `DEFAULT_STATE` to the `LayerViewState` class.

cygnus-x-1:
> @arcturus2 I've included the fix to the issue you mentioned about the visibility of the elements always being invisible. I must have gotten rid of a line of code or refactored something incorrectly in the previous PR to cause this.
> 
> @cygnus-x-1 I've added a comment about the State Concept being within the State class - Let me know if a misunderstanding was the case and I can get a fix in for that inner class.
> Also, I abstracted the controller grabbing the whole pane to fetch the `List<LayerDescription>` - does that satisfy your comment regarding the getter and setter pattern for layers?
> I tried again regarding the static reference using `parent.getContent()` within the plugin and ran into the same issue.

See my note about the inner class above.

What I was trying to get across about the "set state" plugin was that it shouldn't need to know about panes - you should be able to make it work by just passing it minimal information (in this case the list of layers) as all it should be doing is writing that information to the graph. The Layers View should then respond using an `AttributeValueChangeHandler` to update the panes appropriately with that new value.

aldebaran30701:
I've just committed a change that reverts the addition of the inner class for the State Concepts of Analytic view and Layers view. This is per https://github.com/constellation-app/constellation/pull/454#discussion_r419874346 which explains a misunderstanding in changes requested.

I'm happy to have this merged in, and create a ticket to refactor the way the state currently is updated for a future fix if that is alright with both of you @arcturus2 @cygnus-x-1

aldebaran30701:
#524 now supersedes this

PR #457, opened 2020-04-23
Rule (java) null should not be returned from a boolean method (COMMENT)

sol695510:
### Description of the Change
Added comment to affirm returned non-boolean value is intentional.

### Alternate Designs
N/A

N/A

### Benefits
Code clarification.

### Possible Drawbacks
N/A

### Verification Process
N/A

### Applicable Issues
#280


sol695510:
@cygnus-x-1 @antares1470 I hadn't pushed the change with the comment together with the last pull request, my apologies.

PR #464, opened 2020-04-28
Compiler warnings part 3

antares1470:
### Description of the Change

Handled most of the compiler warnings for the following modules:
- Core Histogram View
- Core Import Export Plugins
- Core Interactive Graph
- Core Map View
- Core Named Selection View
- Core Plugin Framework
- Core Quality Control View
- Core Scatter Plot View


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/model/CellValue.java
```diff
@@ -23,6 +23,7 @@
 
     private final String originalText;
 
+    @SuppressWarnings("rawtypes") //raw type needed as comparable value can be string or integer
```
arcturus2, line 26: @antares1470 does it help get rid of the warning if you change this to Comparable<Object> though that is what we are dealing with already?

antares1470: Sadly not in this case

PR #466, opened 2020-04-28
Sanitise the user.name property prior to use

arcturus2:
### Description of the Change

* Perform basic sanitisation of system properties
* Added `StringUtilities.removeSpecialCharacters`
* Fix the location of the `StringUtilities` unit test

### DIFF
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/text/StringUtilities.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/text/StringUtilities.java
@@ -129,7 +129,7 @@ public class StringUtilities {
      */
     public static List<String> splitEscaped(final String string, final String delimiter) {
         final List<String> splits = new ArrayList<>();
-        
+
         if (string == null) {
             return splits;
         }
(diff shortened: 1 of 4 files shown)


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/text/StringUtilities.java
```diff
@@ -369,4 +369,14 @@ public static String removeSquareBracketsFromString(final String text) {
...
     }
+
+    /**
+     * Remove special characters from the {@link String}
+     *
+     * @param text the string to be modified.
+     * @return A {@link String} without special characters
+     */
+    public static String removeSpecialCharacters(final String text) {
+        return text != null ? text.replaceAll("[^A-Za-z0-9]", "") : null;
```
cygnus-x-1, line 380: You could use `[^\\w]` here instead; that would additionally allow underscores , but that seems reasonable to me.

arcturus2: @cygnus-x-1 yeah but an _ is a special character right? I'd prefer to leave this as is if that's ok?

arcturus2: @cygnus-x-1 (re-sending in case you missed it) did you prefer this still be changed or happy as is?

cygnus-x-1: I'm fine with it if you are.

Issue #465, opened 2020-04-28
Null Pointer when selecting "Show Key Columns" with no graph open


sol695510:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When the Table View is opened when no graph is opened or loaded, selecting "Show Key Columns" causes a Null Pointer.

### Steps to Reproduce

1. Open Constellation 2.0.0.
2. Open the Table View.
3. Select the "Column Visibilty" button (first button at the top on the left side of the Table View) then select "Show Key Columns".

**Expected behaviour:** Do nothing.

**Actual behaviour:** Causes Null Pointer.

**Reproduces how often:** 100%

### Additional Information

![Table View - Show Key Columns no graph loaded with error](https://user-images.githubusercontent.com/58677774/80445280-022cc200-8953-11ea-871f-bfdffe4f2098.PNG)

formalhaut69:
Took a look at this in v1.3. Clicking the Show Key Columns option does nothing so this is definitely new to v2. Here's the call stack of the exception;
![image](https://user-images.githubusercontent.com/58677762/80554338-fcdf7e00-8a0b-11ea-8858-ccddfdd57aa5.png)

PR #473, opened 2020-04-30
Update TableViewPane.java

sol695510:
### Description of the Change

Added an "if" statement in the lamba for the action on the "Show Key Columns" button to check if the current graph was Null before proceeding.
### Alternate Designs

N/A

N/A
### Benefits
Avoids a Null Pointer Exception.

### Possible Drawbacks

N/A
### Verification Process

1. Opened the Table View.
2. Select the "Column Visibilty" button (first button at the top on the left side of the Table View) then select "Show Key Columns".

No Null Pointer Exception occurred when the above actions were performed while no graph was open/loaded.

Repeated the above actions with both an empty graph and a sphere graph was open to ensure that the actual function of the button wasn't lost.

### Applicable Issues

#465


sol695510:
Apologies @arcturus2 forgot to change the name of the pull request from the default name.

arcturus2:
Thank you @sol695510

PR #480, opened 2020-05-02
Scatter3d - allow negative numbers to have logarithmic expansion as well

CrucisGamma:
### Description of the Change

Allow logs for negative numbers.  Log(-veNumber) 
### Alternate Designs

As it was, I had set everything to be 0

Already is.

### Benefits

Able to see more visual based information for negative numbers.

### Possible Drawbacks

Nil

### Verification Process

Was checked via different datasets being thrown at it.

### Applicable Issues

Nil.

### DIFF
--- a/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
+++ b/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
@@ -176,12 +176,12 @@ public class Scatter3dArranger implements Arranger {
         }
     }
 
-    private float scaleValue(float value, boolean logarithmic) {
+    private float scaleValue(final float value, final boolean logarithmic) {
         if (logarithmic) {
-            if (value <= 0.0) {
-                return (float) Math.log10(-value) * -1;
+            if (value != 0.0) {
+                return (float) Math.log10(Math.abs(value)) * Math.signum(value);
             }
-            return (float) Math.log10(value);
+            return 0.0f;
         }
         return value;
     }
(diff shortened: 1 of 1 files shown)


CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
```diff
@@ -179,7 +179,7 @@ public void arrange(final GraphWriteMethods wg) throws InterruptedException {
     private float scaleValue(float value, boolean logarithmic) {
```
cygnus-x-1, line 179: Not enough finals :)

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
```diff
@@ -179,7 +179,7 @@ public void arrange(final GraphWriteMethods wg) throws InterruptedException {
...
         if (logarithmic) {
             if (value <= 0.0) {
-                return 0.0f;
+                return (float) Math.log10(-value) * -1;
```
cygnus-x-1, line 182: You could do this in one operation using `Math.log10(Math.abs(value)) * Math.signum(value)`.

spjspj: Hmm, I actually broke it with the request :( 
log10(0) = -Infinity is the thing I want to avoid.

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
CrucisGamma: Added.

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/scatter/Scatter3dArranger.java
CrucisGamma: Seems that log10(0) * 0 still gives NaN so kept the 0.0 case in.

cygnus-x-1: Good catch.

Issue #475, opened 2020-05-01
Editing a date attribute by typing in a value does not update that attribute.


cygnus-x-1:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When editing a date attribute (data, datetime, local_datetime) in the Attribute Editor, the date value is only applied to the attribute if you used the calendar ui to update it; if you instead typed in a value, it will be reset upon confirmation.

### Steps to Reproduce

1. Edit a date/datetime/local_datetime attribute in the Attribute Editor.
2. Type in a new date value and hit ok.

**Expected behaviour:** The attribute will now be set to the date value you entered.

**Actual behaviour:** The attribute will be reset to the previous date value.

**Reproduces how often:** 100%

PR #481, opened 2020-05-03
Editing a date attribute by typing in a value does not update that attribute. #475

formalhaut69:
Fix for date attributes not accepting typed values

### Description of the Change
Added a fix to allow date attributes to accept typed dates

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/DateEditorFactory.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/DateEditorFactory.java
@@ -28,6 +28,7 @@ import javafx.scene.control.CheckBox;
 import javafx.scene.control.DatePicker;
 import javafx.scene.layout.GridPane;
 import javafx.util.converter.LocalDateStringConverter;
+import org.apache.commons.lang3.StringUtils;
 import org.openide.util.lookup.ServiceProvider;
 
 /**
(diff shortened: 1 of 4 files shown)


CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/DateEditorFactory.java
```diff
@@ -69,9 +69,13 @@ protected LocalDate getValueFromControls() throws ControlsInvalidException {
             if (noValueCheckBox.isSelected()) {
                 return null;
             }
+            final String dateString = datePicker.getEditor().getText();
+            //The converter is being used here to try and determine if the entered date is a LocalDate
+            //It will throw an exception and won't convert it if its invalid
             try {
-                final String dateString = datePicker.getEditor().getText();
-                datePicker.getConverter().fromString(dateString);
+                if (!dateString.equals("")) {
```
cygnus-x-1, line 76: I would suggest using `String.isBlank` or even `StringUtil.isBlank` to additionally ignore null.

arcturus2: @formalhaut69 I'd second what @cygnus-x-1 says with `StringUtil.isBlank`.

Side tracking a little and this is purely me sharing a tip I've learnt (from SQ) is in general if you have to compare a string with an object, putting the string to the left of the operation helps avoid an NPE - e.g. `"Some Text".equals(someObject)` 😄

Issue #407, opened 2020-04-06
Timeline View out of memory error


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The timeline view can cause Constellation to crash with an out of memory exception if you open a graph with too many transactions. All of the graphics you see in the timeline view are JavaFX objects which take up a considerable amount of memory.

The following line seems to be where the memory shoots right up on a large graph:
https://github.com/constellation-app/constellation/blob/c997b6ff4a6f9bc18494c8726f085e11d1a4c5c0/CoreTimelineView/src/au/gov/asd/tac/constellation/views/timeline/TimelinePanel.java#L354

### Steps to Reproduce

1. Experimental -> Build Graph -> Sphere Graph
2. Enter a large number and this will depend on your hardware specs and how much memory you have allocated to Constellation. For example 50,000 transactions.
3. Open the timeline view

**Expected behaviour:** The timeline view and the rest of the application should be responsive

**Actual behaviour:** The application should be responsive to mouse selections

**Reproduces how often:** 100%

### Additional Information

I'm not sure how much we can do to fix this though it is worth investigating and seeing what optimisations we can make to the timeline view.

Nova-2119:
Hmm, it does use a fair amount of memory for me but nothing unreasonable, not enough to cause the issues described in the ticket?
Ive managed to reduce the memory usage a little, and the amount of memory it is using seems comparable to the amount used when opening the same graph. It's also only a temporary increase (there is a permanent increase in memory usage but not on that level).

Ill put up a pull request for the small improvement I made but i'm tempted to say nothing to see here unless someone can confirm this is still an issue?

Nova-2119:
My thinking being that from what I am seeing if you can open the graph you should then be able to run the timeline view on it as the memory required is roughly on par?

PR #482, opened 2020-05-04
Rendering blocks on timeline view

antares1470:
### Description of the Change
Have added a boolean condition to the updateTimeline function which ensures that blocks that fall completely out of visibility are not rendered.

Note that this was originally done as an attempt to help fix the memory issues with the timeline view (#407). After some testing it doesn't appear to have helped with this problem but nonetheless, having this change seems like a worthwhile inclusion.

Because the Timeline view is a part of Core

### Benefits

This ensures that we aren't rendering things in the timeline that can't be seen

### Verification Process

- Adjusted the boolean condition slightly so that a block being rendered or no longer being rendered was more obvious
- Created a sphere graph and opened up the timeline view
- panned across the timeline view to see the blocks appear and disappear


antares1470:
Doing some further digging, it looks like I haven't really added any functionality as it turns out the boolean condition I'm applying has already been done in another place (before the function is run)

Therefore, I'm gonna go ahead and close this PR

Issue #448, opened 2020-04-21
Drag from Attribute Editor to Graph View results in Exception


serpens24:
### Description
Exception thrown when dragging from Attribute Editor to Graph. This occurs for both left click & drag and right click and drag.

If the drag operation is valid, then Exception needs to be addressed, if it is not a valid operation, then it sholdn't be permitted.

### Steps to Reproduce
1. Start Consty
2. Open a new graph
3. Open the attribute Editor
4. Left click in Attribute Editor, drag to the Graph View and release
5. Right click in the Attribute Editor, drag to the Graph View and release

**Expected behaviour:**
Steps 4 and 5 should NOT throw exceptions. Need to discuss with others to confirm if somethign should be copied to Graph View or not.

**Actual behaviour:** 
An exception is being thrown - see attached image

**Reproduces how often:** 
100%

### Additional Information

nil

serpens24:
![Exception](https://user-images.githubusercontent.com/58677108/79836803-cbedc080-83ef-11ea-9e99-a344d135ce09.jpg)

antares1470:
I can confirm that after playing around this issue also happens with non-empty graphs. Will update the ticket name accordingly

PR #485, opened 2020-05-05
Drag from Attribute Editor the graph exception resolution

antares1470:
### Description of the Change
Added a catch for ClassCastException to account for the issue where you would get the exception when you drag a label into the graph.

### Alternate Designs

Looked at options around handling the case when casting ByteBufferInputStream to ByteBuffer (the case that was causing the exception) but ByteBufferInputStream is contained in a private which we can't edit.

Also looked at other options around what to do in the catch block for ClassCastException but ultimately doing nothing felt right.

### Verification Process

- opened a new graph and the attribute editor
- left-clicked and dragged an attribute label from the attribute editor to the graph space
- right-clicked and dragged an attribute label from the attribute editor to the graph space

### Applicable Issues
#448

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/LabelDropper.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/LabelDropper.java
@@ -18,9 +18,11 @@ package au.gov.asd.tac.constellation.views.attributeeditor;
 import au.gov.asd.tac.constellation.graph.Graph;
 import au.gov.asd.tac.constellation.graph.GraphElementType;
 import au.gov.asd.tac.constellation.graph.GraphWriteMethods;
+import au.gov.asd.tac.constellation.graph.schema.visual.GraphLabel;
+import au.gov.asd.tac.constellation.graph.schema.visual.GraphLabels;
+import au.gov.asd.tac.constellation.graph.schema.visual.concept.VisualConcept;
 import au.gov.asd.tac.constellation.graph.visual.dragdrop.GraphDropper;
 import au.gov.asd.tac.constellation.graph.visual.dragdrop.GraphDropper.DropInfo;
-import au.gov.asd.tac.constellation.graph.schema.visual.concept.VisualConcept;
 import au.gov.asd.tac.constellation.plugins.PluginException;
 import au.gov.asd.tac.constellation.plugins.PluginExecution;
 import au.gov.asd.tac.constellation.plugins.PluginInfo;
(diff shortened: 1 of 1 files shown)


mimosa2:
I have tested this. It doesn't throw an exception anymore.

CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/LabelDropper.java
```diff
@@ -133,6 +133,8 @@ public void edit(final GraphWriteMethods wg, final PluginInteraction interaction
 
             } catch (final UnsupportedFlavorException | IOException | ClassNotFoundException ex) {
                 Exceptions.printStackTrace(ex);
+            } catch (final ClassCastException ex) {
+                //Do nothing
```
arcturus2, line 137: @antares1470 would you be able to explain a bit more of some of the scenarios that lead to this happening "e.g. dropping a attribute label causes this"; something like that so we know why we are doing nothing? Thank you.

antares1470: Done

antares1470:
@cygnus-x-1 `DataFlavor` comes from a Java library so I'm unable to edit anything in it.

As for the code itself, the issue is in line 82:
`transferable.getTransferData(ATTRIBUTE_LABEL_FLAVOR)` has a return type of Object and it is dynamically being decided that it is a `ByteBufferInputStream` which happens to be defined in a private class `DataFlavorUtils` (which isn't our code). This makes it difficult to come up with a better solution than what I went with.

cygnus-x-1:
> @cygnus-x-1 `DataFlavor` comes from a Java library so I'm unable to edit anything in it.
> 
> As for the code itself, the issue is in line 82:
> `transferable.getTransferData(ATTRIBUTE_LABEL_FLAVOR)` has a return type of Object and it is dynamically being decided that it is a `ByteBufferInputStream` which happens to be defined in a private class `DataFlavorUtils` (which isn't our code). This makes it difficult to come up with a better solution than what I went with (as all the better ones I thought of required `ByteBufferInputStream` being publicly accessible).

I might be missing something here - but if that's the case then drag and drop for Attribute Editor labels will always fail and there's no point keeping these classes around. I would have thought there must be a way to update the definition of the `Transferrable` coming from the Attribute Editor to either return the correct object type or change its definition to use `ByteBufferInputStream`?

antares1470:
> I might be missing something here - but if that's the case then drag and drop for Attribute Editor labels will always fail and there's no point keeping these classes around. I would have thought there must be a way to update the definition of the `Transferrable` coming from the Attribute Editor to either return the correct object type or change its definition to use `ByteBufferInputStream`?

The problem with the latter idea is that because it's a private class in a library we don't control, we can't really play around with it in code (someone let me know if that's not true).

If we can't get it to work though, you make a good point about whether we still need LabelDropper. Looking at its usage, there doesn't appear to be other classes using it directly and so it could be removed with little impact on the rest of the code base and may be the better option if can't determine a way to get this working as intended. Interested in thoughts from others. @arcturus2 @algol60

arcturus2:
I'm going to merge this in to fix the Exception. If we want to make some use of this like create nodes then that can still happen later on.

PR #487, opened 2020-05-07
Added code to ensure temp POI directory exists prior to excel export …

serpens24:
…... this shouldn't be needed  (as POI code is meant to create the folder as needed), however I found cases where without it the export failed...

### Description of the Change

Added a couple lines of code to force temporary poifiles directory to be created prior to call to Excel export  (SXSSFWorkbook package) which seemed to have some occasional/rare uissues if poifiles directory didn't exist (even though this code does create the directory).

### Alternate Designs

Nil

Address occasional issue exporting to Excel.

### Possible Drawbacks

None known.

### Verification Process

1. Open Constelattion
2. Create a graph
3. Open table view
4. Select Export to Excel and export the data - confirm file is created
5. using Windows Expolrer or alternate means, delete the directory %TEMP%/poifiles
6. Select Export to Excel and export the data - confirm file is created
7. Close constellation.

Repate the above steps multiuple times and confirm files are created in expected location.

### Applicable Issues

Nil

### DIFF
--- a/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewUtilities.java
+++ b/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewUtilities.java
@@ -228,9 +228,9 @@ public class TableViewUtilities {
         @Override
         public void execute(final PluginGraphs graphs, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException, PluginException {
             try (final SXSSFWorkbook workbook = new SXSSFWorkbook(SXSSFWorkbook.DEFAULT_WINDOW_SIZE)) {
-                File poi_dir = new File(System.getProperty("java.io.tmpdir"), "poifiles");
-                if (!poi_dir.exists()) {
-                    poi_dir.mkdir();
+                File poiDir = new File(System.getProperty("java.io.tmpdir"), "poifiles");
+                if (!poiDir.exists()) {
+                    poiDir.mkdir();
                 }
                 final Sheet sheet = workbook.createSheet(sheetName);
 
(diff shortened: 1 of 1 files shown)


CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewUtilities.java
```diff
@@ -1,354 +1,358 @@
...
+            this.file = file;
+            this.table = table;
+            this.selectedOnly = selectedOnly;
+            this.sheetName = sheetName;
+        }
+
+        @Override
+        public void execute(final PluginGraphs graphs, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException, PluginException {
+            try (final SXSSFWorkbook workbook = new SXSSFWorkbook(SXSSFWorkbook.DEFAULT_WINDOW_SIZE)) {
+                File poiDir = new File(System.getProperty("java.io.tmpdir"), "poifiles");
```
arcturus2, line 231: @serpens24 can we get a comment put here just to remind us years from now why we did this? Actually...I'm also wondering given you found this bug - would you like to commit this fix in the Apache poi library?

serpens24: @arcturus2 ... I've just spent the past hour trying to reproduce the original problem again, and haven't been able to (without my fix). I'm now at the point where I think the reason I was seeing the problem was how I was selecting the target directory sometimes ... in that I must have not double clicked to folder. .. because the default directory is the Network Shortcut folder that I've mentioned. I've discovered that clicking the desktop folder once is not selecting the folder, just highlighting it ... I suspect that is what I was doing. 
I'm going to suggest we discard my changes . ... as they are probably overkill brought about by me not double clicking.

arcturus2: @serpens24 though I didn't mention it at the time I did feel this was overkill - happy to discard too - thanks mate!

Issue #389, opened 2020-04-01
Box selection stops working when transitioning to (x2, y2, z2) 


serpens24:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

You can manually transition node positions between (x, y, z) and (x2, y2, z2) using the Page-Up and Page-Down keys. However, doing so results in box selection no longer working if the nodes are at any position other than (x, y, z). Click selection works regardless of the node positions. 
(Editors node - does page up/down, toggle between (x, y, z) and (x2, y2, z2) .. or zoom in/out graph, I suspect its a form of zoom in / out)

### Steps to Reproduce

1. Crerate a new empty graph
2. Use Edit->Draw Mode and add several nodes to the graph at positions othewr than 0 (say [10,10], [10, -10], [20, 5])
3. Press page up 5 or 6 times
4. Use rectangle select to make a tight selection around the nodes and Note that not all nodes (if any) are selected
5. Scroll the mouse wheel down so that entire graph is zoom way out and nodes are mere specks
6.Use rectangle seelction to make a large selection of all of graph and note the nodes are selected

From the above steps, it appears the coordinates the rectangle selection is extracting from the graph, and the coordinates the nodes are using are differnet meaning interesction tests are failing

**Expected behaviour:**
Step 4 should select all nodes

**Actual behaviour:** 
Step 4 selected 0 nodes

**Reproduces how often:**
100%

### Additional Information

NIL

serpens24:
I'm quite suspicious of the following code in BoxSelectionPlugin.java

```
            // If mixing is required then mix the main location with the alternative location.
            if (requiresMix) {
                x = inverseMix * x + mix * graph.getFloatValue(x2Attr, vxId);
                y = inverseMix * y + mix * graph.getFloatValue(y2Attr, vxId);
                z = inverseMix * z + mix * graph.getFloatValue(z2Attr, vxId);
            }

            <snip>

            if (sceneLocation.getZ() < 0) {
[...]

PR #489, opened 2020-05-07
Box selection when moving to x2,y2,z2 coordinates

antares1470:
### Description of the Change

Fixed the bug where box selection doesn't work properly when not at the x,y,z coordinates (moving towards x2,y2,z2 coordinates). This was caused by a mismatch in mix ratio ranges where the code was expecting a different range to what it was actually receiving.

### Benefits

Users can box select properly regardless of the number of times PgUp and PgDn are pressed

### Verification Process

1) Opened a new graph and created a few nodes in different places
2) Verified I could box select each node both individually and all together
3) Pressed PgUp a few times and ran the same verification process as in step 2. Did this a few times for different number of PgUps and PgDns
4) Added some transactions and did the same thing

### Applicable Issues
#389

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/select/BoxSelectionPlugin.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/select/BoxSelectionPlugin.java
@@ -58,8 +58,7 @@ public final class BoxSelectionPlugin extends SimpleEditPlugin {
     @Override
     public void edit(final GraphWriteMethods graph, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException {
 
-        //Divide by 20 to change the mix range from 0-20 to 0-1
-        final float mix = camera.getMixRatio() / 20.0f;
+        final float mix = camera.getMix();
         final float inverseMix = 1.0f - mix;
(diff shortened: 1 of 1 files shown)


CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/select/BoxSelectionPlugin.java
```diff
@@ -58,7 +58,8 @@ public BoxSelectionPlugin(final boolean isAdd, final boolean isToggle, final Cam
     @Override
     public void edit(final GraphWriteMethods graph, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException {
 
-        final float mix = camera.getMixRatio();
+        //Divide by 20 to change the mix range from 0-20 to 0-1
+        final float mix = camera.getMixRatio() / 20.0f;
```
cygnus-x-1, line 62: it looks like camera provides a method for doing this in `Camera#getMix()` - this would avoid the magic number.

antares1470: Good find. Didn't even see that. Made the change

Issue #486, opened 2020-05-05
Labels not rendering on MacOS


arcturus2:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Following on from the graph offset fix #387, further testing showed that graph labels are not rendering. By panning the graph there is a brief glimpse of them (not shown in this gif) but essentially they don't render.

![mac no labels](https://user-images.githubusercontent.com/39325530/81053451-6f6ac500-8f08-11ea-8ee6-edcab9031357.gif)

### Steps to Reproduce

1. Open a graph
2. Change to edit mode
3. Click on the graph to add a node

**Expected behaviour:** Nodes should render with a node label beneath the node

**Actual behaviour:** Labels are not displayed

**Reproduces how often:** 100%

### Additional Information

This issue has been a problem for many years. Some Linux machines suffer for the same problem.

arcturus2:
The fix to this is in https://github.com/constellation-app/constellation/releases/tag/v2.0.0-beta3 and confirmed working. Thanks @hydra41 and @TheTimurid for this too :pray:.

PR #490, opened 2020-05-08
Fix labels not rendering on MacOS (#486)

hydra41:
There is an issue on MacOS where JOGL code to change the GL context using release()/makeCurrent() can cause a thread lock. It looks like this issue was occurring when trying to update the glyph textures (nodeLabel and connectionLabel text). As the Blaze render calls happen after the labels it also caused the Blazes to disappear.  An explanation of the issue was found in Chapter 2 of 'Pro Java 6 3D Game Development':
   This coding approach means that the context is current for the
   entire duration of the thread's execution. This causes no problems
   on most platforms (e.g. it's fine on Windows), but unfortunately
   there's an issue when using X11. On X11 platforms, a AWT lock is
   created between the GLContext.makeCurrent() and GLContext.release()
   calls, stopping mouse and keyboard input from being processed.

### DIFF
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/GraphRenderable.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/GraphRenderable.java
@@ -35,6 +35,7 @@ import au.gov.asd.tac.constellation.visual.opengl.utilities.GLTools;
 import au.gov.asd.tac.constellation.visual.opengl.utilities.RenderException;
 import com.jogamp.opengl.GL3;
 import com.jogamp.opengl.GLAutoDrawable;
+import com.jogamp.opengl.GLContext;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


arcturus2:
Thank you for this @hydra41. I've done some testing and some issues have come up:

Scenario 1

1. If I open a new graph and add new nodes it works fine!
2. If I then try to open a view (e.g Views -> Data Access View) then it locks up

Scenario 2

1. If I open the Data Access View it loads
2. If I then open a new graph and add nodes, it works fine
3. If I close the Data Access View the app locks up and the nodes disappear from the graph

hydra41:
> Thank you for this @hydra41. I've done some testing and some issues have come up:
> 
> Scenario 1
> 
> 1. If I open a new graph and add new nodes it works fine!
> 2. If I then try to open a view (e.g Views -> Data Access View) then it locks up
> 
> Scenario 2
> 
> 1. If I open the Data Access View it loads
> 2. If I then open a new graph and add nodes, it works fine
> 3. If I close the Data Access View the app locks up and the nodes disappear from the graph

Thanks for testing, I've just pushed a fix for the locking up issue when opening the Data Access View. Let me know how it goes.

PR #496, opened 2020-05-11
Fixed a bug preventing v1 graphs from being open.

arcturus2:
### Description of the Change

Fixed a bug preventing v1 graphs from being open. The exception was:

```
SEVERE [global]
java.lang.ArrayIndexOutOfBoundsException: Index -1107 out of bounds for length 64
	at au.gov.asd.tac.constellation.graph.StoreGraph.getFloatValue(StoreGraph.java:1882)
	at au.gov.asd.tac.constellation.graph.visual.framework.GraphVisualAccess.getConnectionVisibility(GraphVisualAccess.java:1134)
	at au.gov.asd.tac.constellation.visual.opengl.renderer.batcher.LineBatcher.bufferColorInfo(LineBatcher.java:258)
	at au.gov.asd.tac.constellation.visual.opengl.renderer.batcher.LineBatcher.lambda$createBatch$1(LineBatcher.java:165)
	at java.base/java.util.ArrayList.forEach(Unknown Source)
	at au.gov.asd.tac.constellation.visual.opengl.renderer.batcher.LineBatcher.createBatch(LineBatcher.java:161)
	at au.gov.asd.tac.constellation.visual.opengl.renderer.GraphRenderable.lambda$getChangeProcessor$4(GraphRenderable.java:178)
	at au.gov.asd.tac.constellation.utilities.visual.VisualProcessor.lambda$processChangeSet$1(VisualProcessor.java:296)
	at java.base/java.util.ArrayList.forEach(Unknown Source)
	at au.gov.asd.tac.constellation.utilities.visual.VisualProcessor.lambda$processChangeSet$2(VisualProcessor.java:295)
	at java.base/java.util.HashMap.forEach(Unknown Source)
	at au.gov.asd.tac.constellation.utilities.visual.VisualProcessor.processChangeSet(VisualProcessor.java:293)
	at au.gov.asd.tac.constellation.utilities.visual.VisualProcessor.lambda$update$0(VisualProcessor.java:206)
[catch] at java.base/java.lang.Thread.run(Unknown Source)
```

### Verification Process

1. Created a graph in v1.3.1
2. Tried to open the graph in v2

PR #500, opened 2020-05-11
Hotfix/backward compatibility

aldebaran30701:
### Description of the Change
Check if the attribute is found on a graph, if not then create that attribute.
This helps with support for older graphs which would not have the new attributes used within the layers view.

### Alternate Designs
N/A

It fixes backwards compatibility with elements on V1 graphs

### Benefits
It fixes backwards compatibility with elements on V1 graphs

### Possible Drawbacks
Has to recheck the new graphs if they also have those attributes.

### Verification Process
Add transactions to layers from graphs which were created using V1.
Add vertices to layers from graphs which were created using V1.

### Applicable Issues
Found through #496

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
@@ -125,20 +125,20 @@ public class StoreGraph extends LockingTarget implements GraphWriteMethods, Seri
     private NativeValue oldValue = new NativeValue();
     private NativeValue newValue = new NativeValue();
     private GraphEdit graphEdit;
-    private int layerMaskSelectedAttributeId = -1;
-    private int vertexLayerMaskAttribureId = -1;
-    private int transactionFilterBitmaskAttrId = -1;
-    private int vertexLayerVisibilityAttributeId = -1;
-    private int transactionLayerVisibilityAttributeId = -1;
-    private int cameraAttributeId = -1;
-    private int vertexSelectedAttributeId = -1;
-    private int transactionSelectedAttributeId = -1;
-    private int vertexXAttributeId = -1;
-    private int vertexYAttributeId = -1;
-    private int vertexZAttributeId = -1;
-    private int transactionXAttributeId = -1;
-    private int transactionYAttributeId = -1;
-    private int transactionZAttributeId = -1;
+    private int layerMaskSelectedAttributeId = Graph.NOT_FOUND;
+    private int vertexLayerMaskAttribureId = Graph.NOT_FOUND;
+    private int transactionFilterBitmaskAttrId = Graph.NOT_FOUND;
+    private int vertexLayerVisibilityAttributeId = Graph.NOT_FOUND;
+    private int transactionLayerVisibilityAttributeId = Graph.NOT_FOUND;
+    private int cameraAttributeId = Graph.NOT_FOUND;
+    private int vertexSelectedAttributeId = Graph.NOT_FOUND;
+    private int transactionSelectedAttributeId = Graph.NOT_FOUND;
+    private int vertexXAttributeId = Graph.NOT_FOUND;
+    private int vertexYAttributeId = Graph.NOT_FOUND;
+    private int vertexZAttributeId = Graph.NOT_FOUND;
+    private int transactionXAttributeId = Graph.NOT_FOUND;
+    private int transactionYAttributeId = Graph.NOT_FOUND;
+    private int transactionZAttributeId = Graph.NOT_FOUND;
 
     // TODO: TEMP VARIABLE to test updates - This will have to be migrated to a graph attribute
     private static int currentVisibleMask = 1;
@@ -2001,9 +2001,9 @@ public class StoreGraph extends LockingTarget implements GraphWriteMethods, Seri
      */
     private int recalculateLayerMask(final GraphElementType elementType, final int elementId) {
         int bitmask = 0;
-        if (elementType == GraphElementType.VERTEX && vertexLayerMaskAttribureId >= 0 && vertexLayerVisibilityAttributeId >= 0) {
+        if (elementType == GraphElementType.VERTEX && vertexLayerMaskAttribureId != Graph.NOT_FOUND && vertexLayerVisibilityAttributeId != Graph.NOT_FOUND) {
             bitmask = getIntValue(vertexLayerMaskAttribureId, elementId);
-        } else if (elementType == GraphElementType.TRANSACTION && transactionFilterBitmaskAttrId >= 0 && transactionLayerVisibilityAttributeId >= 0) {
+        } else if (elementType == GraphElementType.TRANSACTION && transactionFilterBitmaskAttrId != Graph.NOT_FOUND && transactionLayerVisibilityAttributeId != Graph.NOT_FOUND) {
             bitmask = getIntValue(transactionFilterBitmaskAttrId, elementId);
         }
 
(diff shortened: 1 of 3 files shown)


CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
```diff
@@ -2206,6 +2208,43 @@ private void updateLayerMask(final int attributeId, final GraphElementType eleme
 
     // update the layer visibility for an element when an attribute changes
     private void updateLayerMask(final int attributeId, final int elementid) {
+        if(transactionLayerVisibilityAttributeId < 0) {
```
arcturus2, line 2211: @aldebaran30701 if there isn't a specific reason to do this I would prefer the explicit check for `Graph.NOT_FOUND` which keeps this consistent with the rest of the app.

FYI going off on a tangent and a more for your info, we also have `AttributeUpdateProvider` which can be used to "upgrade" attributes to a new version. Say you stored the layer visibility as a String and you wanted to change it to an List (for whatever reason) you can use an update provider to upgrade the attribute during load.

arcturus2:
@aldebaran30701 thank you for being so responsive at short notice. Really appreciate it.

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
```diff
@@ -123,20 +125,20 @@
...
+    private int transactionLayerVisibilityAttributeId = Graph.NOT_FOUND;
+    private int cameraAttributeId = Graph.NOT_FOUND;
+    private int vertexSelectedAttributeId = Graph.NOT_FOUND;
+    private int transactionSelectedAttributeId = Graph.NOT_FOUND;
+    private int vertexXAttributeId = Graph.NOT_FOUND;
+    private int vertexYAttributeId = Graph.NOT_FOUND;
+    private int vertexZAttributeId = Graph.NOT_FOUND;
+    private int transactionXAttributeId = Graph.NOT_FOUND;
+    private int transactionYAttributeId = Graph.NOT_FOUND;
+    private int transactionZAttributeId = Graph.NOT_FOUND;
```
arcturus2, line 141: @cygnus-x-1 @algol60 I asked @aldebaran30701 to change the layer view mask attributes to `Graph.NOT_FOUND` instead of `-1`. I wanted to ask either of you if you have any issues with this change for the camera, selection and x,y,z?

@cygnus-x-1 I would have expected you would have picked this up ages ago in your mass refactor/cleanup PRs and made them `Graph.NOT_FOUND` so a little curios/converned if there is something I should know  - is there anything special with `-1`?

I get very nervous/paranoid changing `StoreGraph`.

aldebaran30701:
I've just added another commit - I'm going through all the new code that may access the attributes in a dangerous way to make sure it doesn't occur in other cases

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
aldebaran30701: @arcturus2 I was the one who added the `cameraAttributeId` and alike into `StoreGraph`
The purpose was to capture the attribute id of the camera and other commonly changed values. When it is one of the commonly changed atributes - the graph does not trigger a recalculation of layers. The recalculation of layers currently loops all layers, for every element on the graph. So this was implemented to restrict updates.

cygnus-x-1: I think I suggested this change when they were added...

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
```diff
@@ -2179,36 +2175,73 @@ private void updateLayerMask(final int attributeId, final GraphElementType eleme
...
+                    attributeModificationCounters[transactionLayerVisibilityAttributeId] += operationMode.getModificationIncrement();
                 }
             }
         }
     }
 
     // update the layer visibility for an element when an attribute changes
     private void updateLayerMask(final int attributeId, final int elementid) {
+        if(transactionLayerVisibilityAttributeId == Graph.NOT_FOUND) {
+            transactionLayerVisibilityAttributeId = this.addAttribute(
```
cygnus-x-1, line 2206: It feels wrong having these defined in a concept and then copied here. Perhaps we need a low-level concept in the graph framework module that classes like `StoreGraph` can access?

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/utilities/UpdateElementBitmaskPlugin.java
```diff
@@ -65,8 +65,8 @@ private void setVertices(final GraphWriteMethods graph, final int currentBitmask
     }
 
     private void setTransactions(final GraphWriteMethods graph, final int currentBitmask) {
-        final int transactionSelectedAttributeId = VisualConcept.TransactionAttribute.SELECTED.get(graph);
-        final int transactionBitmaskAttributeId = VisualConcept.TransactionAttribute.LAYER_MASK.get(graph);
+        final int transactionSelectedAttributeId = VisualConcept.TransactionAttribute.SELECTED.ensure(graph);
+        final int transactionBitmaskAttributeId = VisualConcept.TransactionAttribute.LAYER_MASK.ensure(graph);
         if (transactionSelectedAttributeId != Graph.NOT_FOUND) {
```
cygnus-x-1, line 70: You don't need to use ensure if you're then checking for `Graph.NOT_FOUND`. the ensure method will create the attribute so that it can never be "not found" - i'd stick with get here.

aldebaran30701: My thoughts were that if .get() was called, and it did not exist - then the code within the block would not run. Meaning the user would try adding elements of the graph to a layer and it would appear as nothing was added.

Adding ensure means that when that plugin is ran, it will 100% be added to the layer on first try. I can change this @cygnus-x-1 if the above reasoning doesn't change things.

cygnus-x-1: I missed that this was in a plugin and not `StoreGraph` - so if your intention is for these attributes to always exist, then using `ensure` is correct. However that means you don't need the checks for `Graph.NOT_FOUND`.

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
aldebaran30701: I'd have no real idea on what that means for implementation - Would this stop this PR going ahead?

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/utilities/UpdateElementBitmaskPlugin.java
aldebaran30701: Sound good to me

aldebaran30701:
I'm happy to have this merged in, and look at the issue @cygnus-x-1 highlighted as an enhancement https://github.com/constellation-app/constellation/pull/500#discussion_r422974764

Would you be able to make a ticket to explain this further so we don't forget about doing this?

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
arcturus2: As discussed today can we get this merged in and create the low-level concept (which is a great idea by the way) as a separate task @cygnus-x-1? If so please approve this one when you get the chance.

aldebaran30701:
#524 now supersedes this

PR #501, opened 2020-05-11
Layers context menu improvements

aldebaran30701:
Right click context menu now adds or removes from specific layers

### Description of the Change

Right click context menu on an element now shows the following structure when no Layers view state is present. Default layer 1 is not shown, only layer 2 which is the next usable layer is applicable.

Layers ->
 - Add Selection to Layer... - >
   - 2
 - Remove Selection from Layer... - >
   - 2

In the case of 5 layers on a graph:

Layers ->
 - Add Selection to Layer... - >
   - 2
   - 3
   - 4
   - 5
 - Remove Selection from Layer... - >
   - 2
   - 3
   - 4
   - 5

### Alternate Designs
One other possibility would be to have Layer 1 show.
eg. 
Layers ->
 - Add Selection to Layer... - >
   - 1 (Default)
   - 2
 - Remove Selection from Layer... - >
   - 1 (Default)
   - 2

This would not necessarily be a good idea, as the current structure of the `ContextMenuProvider` doesn't allow for you to disable a `MenuItem`. Plus it will not be intuitive for the user to see the default layer - but not be able to add or remove elements to/from it.

Faster and less error prone than text entry

### Benefits
Faster and less error prone than text entry

### Possible Drawbacks
N/A

### Verification Process
Open new graphs, existing graphs.
Add items to and from layers
Create new layers, open context menu.
Add item to layer without a Layers view state. 
Add item to layer with a Layers view state. 

### Applicable Issues
#223


aldebaran30701:
This is a Draft PR until other layers view PRs have been merged in. This currently relies on the Layers View State to work as intended.

arcturus2:
@aldebaran30701 I have a suggestion which can come in a later PR but its an idea I had that I wanted to share related to this. I'd like to see the name beside the number as that give more meaning to the layer when adding/removing a layer from selection.

aldebaran30701:
#524 now supersedes this

PR #503, opened 2020-05-12
Windows-DPI-Scaling-Fix

TheTimurid:
* JOGL version 2.3.2 on Windows doesn't correctly support DPI scaling.
  setSurfaceScale() is not overridden in WindowsJAWTWindow so it is not
  possible to scale the canvas and mouse events at this level.  It
  should be noted that it is overridden in MacOSXJAWTWindow.  Where manual
  scaling is required the caller scales each GL viewport and
  ensures hit tests take that size into account.
* Changes annoted with Windows-DPI-Scaling comment.

### Possible Drawbacks

This code is restricted only to Windows OSes, but has only been tested on Windows 10.  It is possible this may cause incorrect rendering and selection offsets on other versions of Windows.


arcturus2:
@cygnus-x-1 I know you've mentioned the DPI scaling issue on Windows - would you mind testing this out?

Issue #374, opened 2020-03-27
Table View Column Selection - Split Attributes into Source/Destination/Transaction


cygnus-x-1:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The column selection menu in the Table View becomes unwieldy to use when you have a lot of attributes available. This could be improved by change the layout of the menu so that the attributes are split up by source/destination/transaction instead of all being in one big list.

serpens24:
@arcturus2  - I've pulled this ticket in as a good first issue for Auriga2 - I'm not sure if you have set up their access yet  ?

Anyway, the following screenshot shows one possibility for solving the problem:
It provides 3 additional 'Filter' options which allow source/destination/transaction records to be turned on/off. This works interchangeably with the Filter text field, so in the example below, the user has said they only want to view source and destination fields, and return any containing ".y".

These three 'Filter Source/destination/transaction' options can probably be only visible when required ... 
We went for this over nested menus for each as they are a little harder to use when wanting to search for text across multiple types.
![image](https://user-images.githubusercontent.com/58677108/81357995-5afb1800-9114-11ea-995b-aef2c9e86b0e.png)

arcturus2:
@serpens24 sorry mate the access for @Auriga2 should be waiting for them now.

In my mind I had assumed the solution would be 3 sub menus and possibly a filter at each sub menu but I think your saying that is much harder to implement? If so I think this is OK and achieves the requirement.

@cygnus-x-1 I'd be grateful for any feedback you have when you get a moment?

serpens24:
> @serpens24 sorry mate the access for @Auriga2 should be waiting for them now.
> 
> In my mind I had assumed the solution would be 3 sub menus and possibly a filter at each sub menu but I think your saying that is much harder to implement? If so I think this is OK and achieves the requirement.
> 
> @cygnus-x-1 I'd be grateful for any feedback you have when you get a moment?

[...]

PR #507, opened 2020-05-12
Table view column selection split in new menu

Auriga2:
### Description of the Change
Table View Column Selection - Split attributes into Source/Destination/Transaction sub menus

### Alternate Designs
Added the initial design in a separate branch, as well as in the previous commit of  this branch, which uses filter check boxes and doesn't have sub menus. It allows the existed filter to work across all types.

Better UX, instead of all attributes being in one big list

### Benefits
Better UX, instead of all attributes being in one big list

### Possible Drawbacks
Moved the existed filter into sub menus which filters within the subcategory only.

### Verification Process
Click on the "Column Visibility" menu button in the Table View of a graph.
It should display separate menu items for Source, Destination and Transaction which will load the relevant attributes in a sub menu, with a Filter at the top.

### Applicable Issues
#374


arcturus2:
Thank you for this @Auriga2 and @serpens24. I did a test and noticed some issues like unable to filter source types and the destination menu wasn't filtering. However, don't worry about it.

testing table view v1

I think I agree with you both and your first design is superior; having the option to decide what you filter on and using one filter has its advantage, primarily being able to search across all 3 types - that is powerful.

You probably expected this but can we go with your first design and use one filter field with options to select source, destination or transactions? Sorry to be a pain!

Auriga2:
No worries, made a separate Pull request #519

arcturus2:
@Auriga2 based on feedback it looks like this option is preferred by people. Therefore when you get a chance, would you be able to look at fixing the issues I pointed out and then test this PR again?

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -124,6 +125,9 @@
     private static final ImageView COPY_ICON = new ImageView(UserInterfaceIconProvider.COPY.buildImage(16));
     private static final ImageView EXPORT_ICON = new ImageView(UserInterfaceIconProvider.UPLOAD.buildImage(16));
     private static final ImageView SETTINGS_ICON = new ImageView(UserInterfaceIconProvider.SETTINGS.buildImage(16));
+    private static final ImageView MENU_ICON_SOURCE = new ImageView(UserInterfaceIconProvider.MENU.buildImage(16));
+    private static final ImageView MENU_ICON_DESTINATION = new ImageView(UserInterfaceIconProvider.MENU.buildImage(16));
+    private static final ImageView MENU_ICON_TRANSACTION = new ImageView(UserInterfaceIconProvider.MENU.buildImage(16));
```
antares1470, line 130: I'm not entirely familiar with ImageViews so feel free to correct me I'm wrong. Given the context that these 3 seem to be used, you may only need to create one instance of this ImageView which can be used for all of 3 rather than creating one for each

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -319,23 +322,25 @@ private ToolBar initToolbar() {
...
-            final String filterTerm = columnFilterTextField.getText().toLowerCase().trim();
-            columnCheckboxes.forEach(item -> {
-                final String columnName = item.getId().toLowerCase();
-                item.setVisible(filterTerm.isBlank() || columnName.contains(filterTerm));
-            });
-        });
-        
+        final ArrayList<CustomMenuItem> columnCheckboxesSource = new ArrayList<>();
+        final ArrayList<CustomMenuItem> columnCheckboxesDestination = new ArrayList<>();
+        final ArrayList<CustomMenuItem> columnCheckboxesTransaction = new ArrayList<>();
```
antares1470, line 327: For these 3, it is better to declare them simply as Lists and instantiate as an arraylist.
i.e. `final List<CustomMenuItem> foo = new ArrayList<>();`

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -827,17 +912,17 @@ private void loadPreferences() {
...
                 final List<ThreeTuple<String, Attribute, TableColumn<ObservableList<String>, String>>> orderedColumns
-                    = newColumnOrder.stream()
-                        .map(c ->
-                            {
-                                for (ThreeTuple<String, Attribute, TableColumn<ObservableList<String>, String>> col : columnIndex) {
-                                    if (c.getText().equals(col.getThird().getText())) {
-                                        return col;
+                        = newColumnOrder.stream()
+                                .map(c
+                                        -> {
```
antares1470, line 917: I think this was done in the other PR but for readability combine these 3 lines into 1:
`...stream().map(c -> {`

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
Auriga2: I had the same idea, but for some reason I cannot reuse the same ImageView. The image was appearing only on the last place I used it.

Auriga2: image
(Same when I tried to reuse the same filter on 3 sub menus- it works only on the last place)

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
Auriga2: Fixed.

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
Auriga2: Yeah, caused by Alt+Shift+F
Fixed now.

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
antares1470: Yep that makes sense

arcturus2:
@Auriga2 many thanks for this PR btw.

PR #509, opened 2020-05-12
layers-ui

cygnus-x-1:
Improvements to the layer view state design.


arcturus2:
@aldebaran30701 I'm not sure how this change will impact the other branches you have open but please consider this change when you have time and hopefully we can get your branches merged in and then do another review when necessary.

cygnus-x-1:
@aldebaran30701 I also couldn't find a nice way to avoid passing in the pane, so I left that as is. I did however refactor the state pattern - any changes are written to the graph and this triggers the listener in the top component to read those changes from the graph and react.

I also appear to have left some debug logs in there... i'll remove them.

aldebaran30701:
I did a little digging, it appears to be from the `Platform.runLater(() -> { });` call within `LayersViewPane.createLayer(...)`
It must have been a race condition. Removing it fixes the duplication of the pane.

In saying that, further issues are still present. Adding a few extra layers to some new graphs, checking different checkboxes allows a user to add a layer next to the current layers state, and also mess up the order of layers.
image

Note: Selecting layer 9 will toggle layer 5, 4 with 2 etc. Likely due to the reordering of the elements.

cygnus-x-1:
Huh - I was seeing some of that early on but thought I fixed it all. I did have some merge issues, perhaps I missed something.

arcturus2:
I did see a similar issue to the picture @aldebaran30701 put up when I was testing but after I did a clean build that issue went away so I didn't mention it - assumed I must have not clean built it properly.

One thing I have seen was that when adding new layers by clicking on the "Add layers button" I can kind of make out the pane being rebuilt each time; I can see the "Add layers button" shift downwards from top to bottom as new input fields are being created. I wasn't going to mention this either but just in case I'm not alone. If I'm the only one seeing this then I will put it down to a glitch in the matrix :wink:

aldebaran30701:
So it appeared to be a few issues with concurrency and race conditions. What I did was go back and slowly introduce the previous code back in. 
As it stands with my last commit, It was @cygnus-x-1 original files, but I copied over `LayersViewPane.java`
It still allows the state to work as @cygnus-x-1 intended. I did go back over `LayersViewPane.java` and fill in any fixes you included in your PR. 
I am putting together a document for testing sections of the layers view so in the future we can have a complete list of steps to hit all cases during UI testing. 
It would be good to have someone else test this before it goes in.

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -38,142 +41,149 @@
...
-        stateChanged = true;
-        PluginExecution.withPlugin(new LayersViewStateUpdater(pane, layers, wasUpdate)).executeLater(GraphManager.getDefault().getActiveGraph());
+    public void readState() {
+        final LayersViewPane pane = parent.getContent();
+        if (pane == null) {
+            return;
+        }
         
+        final Graph graph = GraphManager.getDefault().getActiveGraph();
+        if(graph == null){
```
antares1470, line 98: This if statement and the one above can be merged into one: `if (pane == null || graph == null)`

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -38,142 +41,149 @@
...
-            stateChanged = false;
-            PluginExecution.withPlugin(new LayersViewStateWriter(layers)).executeLater(GraphManager.getDefault().getActiveGraph());
+    public void writeState() {
+        final LayersViewPane pane = parent.getContent();
+        if (pane == null) {
+            return;
         }
+        
+        final Graph graph = GraphManager.getDefault().getActiveGraph();
+        if(graph == null){
```
antares1470, line 117: can be merged with above if statement: `if (pane == null || graph == null)`

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -38,142 +41,149 @@
...
 
-        @Override
-        protected boolean isSignificant() {
-            return false;
+            final LayersViewState currentState = graph.getObjectValue(layersViewStateAttributeId, 0);
+            if (currentState == null) {
+                return;
+            }
+
+            if (pane == null) {
```
antares1470, line 152: can merge with first if statement above: `if (graph== null || pane == null)`

aldebaran30701:
@antares1470 The following changes you suggest will be fixed in the next commit to #524 
This PR should be closed - as it has been merged and implemented into #524

PR #519, opened 2020-05-13
Tabe view column selection

Auriga2:
### Description of the Change
Table View Column Selection - Split attributes into Source/Destination/Transaction, using filter check boxes

### Alternate Designs
https://github.com/constellation-app/constellation/pull/507

Better UX, instead of all attributes being in one big list

### Benefits
Better UX, instead of all attributes being in one big list

### Possible Drawbacks
N/A

### Verification Process
Click on the "Column Visibility" menu button in the Table View of a graph.
It should display check boxes "Filter Source", "Filter Destination" and "Filter Transaction" which will filter the attributes, along with the text input from the "Filter" below.

### Applicable Issues
#374

### DIFF
--- a/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
+++ b/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
@@ -83,7 +83,7 @@ import org.controlsfx.control.table.TableFilter;
 
 /**
  * Table View Pane.
- * 
+ *
  * TODO: some javafx classes no are longer supported, fix it.
  *
  * @author elnath
@@ -97,7 +97,7 @@ public final class TableViewPane extends BorderPane {
 
     private static final String ATTEMPT_PROCESS_JAVAFX = "Attempting to process on the JavaFX Application Thread";
     private static final String ATTEMPT_PROCESS_EDT = "Attempting to process on the EDT";
-    
+
     private static final String ALL_COLUMNS = "Show All Columns";
     private static final String DEFAULT_COLUMNS = "Show Default Columns";
     private static final String KEY_COLUMNS = "Show Key Columns";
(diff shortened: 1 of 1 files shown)


CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -413,6 +440,27 @@ private ContextMenu initColumnVisibilityContextMenu() {
         return cm;
     }
 
+    private void refreshColumnSelectionMenu(ArrayList<CustomMenuItem> columnCheckboxes, TextField columnFilterTextField, CheckBox columnCheckboxSource,
```
antares1470, line 443: If it doesn't break anything, it would be better for the `columnCheckboxes` to be a list rather than specifically an arraylist

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -827,17 +874,17 @@ private void loadPreferences() {
...
                 final List<ThreeTuple<String, Attribute, TableColumn<ObservableList<String>, String>>> orderedColumns
-                    = newColumnOrder.stream()
-                        .map(c ->
-                            {
-                                for (ThreeTuple<String, Attribute, TableColumn<ObservableList<String>, String>> col : columnIndex) {
-                                    if (c.getText().equals(col.getThird().getText())) {
-                                        return col;
+                        = newColumnOrder.stream()
+                                .map(c
+                                        -> {
```
antares1470, line 879: Move this arrow and open brace to the line above just for readability.
The `.map(c -> {` can then probably be moved to the line above that one (`... stream().map(...)`

Auriga2:
Based on feedback, separate sub menus is the preferred option by people. ( #507)

Issue #533, opened 2020-05-19
Merge Nodes by Geospatial Distance plugin errored with an OutofBounds exception 


mimosa2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Running Merge Nodes plugin in Data Access View with the option Geospatial Distance errored with an ArrayIndexOutofBounds exception.

### Steps to Reproduce

1. Open Data Access View and expand Clean -> Merge Nodes plugin
2. Select Merge By -> Geospatial Distance
3. Click Go button to run.

**Expected behaviour:** Plugin run without error

**Actual behaviour:** Throws an ArrayIndexOutofBounds exception.

**Reproduces how often:** 100%

### Additional Information

![DA_error](https://user-images.githubusercontent.com/54261086/82316029-8e407f80-9a0f-11ea-81b9-0640ac5e451d.PNG)

PR #538, opened 2020-05-20
merge by geospacial distance exception

aldebaran30701:
### Description of the Change
Ensure attributes exist before using. I had to also switch to a write access of the graph.

@mimosa2 I did not realize you were assigned until I looked back at the ticket! Let me know if you had the same thoughts

### Alternate Designs
N/A

Fixes a bug

### Benefits
Fixes a bug

### Possible Drawbacks
None realized

### Verification Process
Tested running the plugin with elements on the graph.

### Applicable Issues
#533

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesByLocation.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesByLocation.java
@@ -122,12 +122,12 @@ public class MergeNodesByLocation implements MergeNodeType {
         return nodesToMerge;
     }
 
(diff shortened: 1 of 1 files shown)


arcturus2:
Thanks for this @aldebaran30701 though I've proposed an alternative solution in PR #541.

What I noticed was that `calculateDistances()` requires the lat/long attributes for it to do the calculation so if they are not defined on the graph, using `ensure()` won't add any value because they will be empty lat/long values.

The side effect of using ensure when you are getting attributes and expect a value is that you end up adding attributes to the graph but they will all be null values. Using `ensure` when you set values to attributes is fine though because if they don't exist you can't add the computed value to the attribute.

Does that make sense? Let me know what you think.

aldebaran30701:
Closed as PR #541 Is a much better solution moving forward.

PR #541, opened 2020-05-20
merge nodes by location (alternative solution)

arcturus2:
### Description of the Change

Fix to Merge nodes by location (alternative solution to #538) which will exit at the start of the plugin and won't create attributes that are not required.

### Verification Process

I've run the plugin after creating some nodes which don't have the lat/long attributes.

### Applicable Issues

#533


mimosa2:
I tested this with nodes w/o the attributes and didn't get the exception. Also tested with nodes with same values of lat/long attributes. The merge worked fine.

Issue #543, opened 2020-05-21
Hierarchical clustering results in null pointer


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Running cluster in Hierarchical clustering results in null pointer and other errors.

```
java.lang.NullPointerException
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.NestedHierarchicalDisplayPanel.updateColorsAndBar(NestedHierarchicalDisplayPanel.java:88)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.NestedHierarchicalDisplayPanel.positionLines(NestedHierarchicalDisplayPanel.java:248)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.NestedHierarchicalDisplayPanel.calculateLines(NestedHierarchicalDisplayPanel.java:216)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.NestedHierarchicalDisplayPanel.setState(NestedHierarchicalDisplayPanel.java:80)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.HierarchicalControllerTopComponent.setGroups(HierarchicalControllerTopComponent.java:532)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.HierarchicalControllerTopComponent.updateInteractivity(HierarchicalControllerTopComponent.java:463)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.HierarchicalControllerTopComponent.interactiveButtonActionPerformed(HierarchicalControllerTopComponent.java:403)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.HierarchicalControllerTopComponent$9.actionPerformed(HierarchicalControllerTopComponent.java:268)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
[...]

Issue #190, opened 2020-01-09
Hierarchical and K-Truss Cluster Interactive button state review


aldebaran30701:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When navigating the Clustering plugin, the disabling of other buttons does not seem enough to hint interactivity is disabled. A UX improvement to the Hierarchical clustering and K Truss menus could be as follows:
![image](https://user-images.githubusercontent.com/58677759/72027165-e6cb7780-32cd-11ea-84af-4ceaf4c45f72.png)
![image](https://user-images.githubusercontent.com/58677759/72027184-fb0f7480-32cd-11ea-9693-f118a047b058.png)
As opposed to the current 
![image](https://user-images.githubusercontent.com/58677759/72027211-0fec0800-32ce-11ea-90ca-560f6d95da0c.png)
Which doesn't change no matter which state the Interactivity is in.

### Steps to Reproduce

1. Open graph
2. Build example graph
3. Open clustering menu
4. Select Interactive button

**Expected behaviour:** 
Some extra user feedback to show button state

**Actual behaviour:** 
Some components are disabled, but the button remains the same

**Reproduces how often:** 
100%

### Additional Information

Determining Interactive state at a glance is often not user friendly

PR #544, opened 2020-05-21
:art: Clustering interactions ux + bug fix

aldebaran30701:
### Description of the Change
Fix a bug with hierarchical clustering #543
Fix a bug with auto-save/save of attribute not working with state.step = -1
Fix a null pointer issue when adding vertices to graph, and toggling interactive. 
Add text to button to allow visible confirmation that the view interactive mode is enabled or disabled. #190 

The clustering menus are now consistent, and when loaded on a graph without state will have the interactive button disabled, and cluster enabled. When cluster is ran, the interactive button will be enabled for use. For graphs with states, this consistency is kept at close and open of the view and or graph.
image

### Alternate Designs
N/A

Fixes bugs, improves ux

### Benefits
Now less error prone and easier for a user to utilize.

### Possible Drawbacks
Fix a bug with auto-save/save of attribute not working with state.step = -1
This fix is outlined in the IOProvider. I would appreciate it if someone else checked it out as it appeared to be getting a value but doing nothing with the returned value. I removed it to solve the bug. 

### Verification Process
Testing with and without state on graphs, opening states and components for the view at different times. Ensuring that the button was disabled at the correct time and enabled when necessary.

### Applicable Issues
#543 
#190

### DIFF
--- a/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/hierarchical/HierarchicalControllerTopComponent.java
+++ b/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/hierarchical/HierarchicalControllerTopComponent.java
@@ -560,7 +560,11 @@ public final class HierarchicalControllerTopComponent extends TopComponent imple
             if (state != null && doUpdate) {
                 updateGraph();
             }
-            interactiveButton.setText(state != null ? state.interactive ? INTERACTIVE_ENABLED : INTERACTIVE_DISABLED : INTERACTIVE_DISABLED);
+            if(state != null && state.interactive) {
+                interactiveButton.setText(INTERACTIVE_ENABLED);
+            } else {
+                interactiveButton.setText(INTERACTIVE_DISABLED);
+            }
         }
         interactiveButton.setEnabled(state != null);
         interactiveButton.setSelected(false);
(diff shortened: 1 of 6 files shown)


CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/hierarchical/HierarchicalControllerTopComponent.java
```diff
@@ -551,7 +560,14 @@ private void setGroups(final boolean doUpdate) {
             if (state != null && doUpdate) {
                 updateGraph();
             }
+            if(state != null && state.interactive) {
+                interactiveButton.setText(INTERACTIVE_ENABLED);
+            } else {
+                interactiveButton.setText(INTERACTIVE_DISABLED);
+            }
```
antares1470, line 567: Up to you but given the if and else code are almost exactly the same bar the parameter value, can changed this to:
`interactiveButton.setText((state != null && state.interactive) ? INTERACTIVE_ENABLED : INTERACTIVE_DISABLED)`

aldebaran30701: I'm all for changing it - I originally had it as a nested ternary and SQ found a code smell. I'll change it and see how SQ likes it

Issue #332, opened 2020-03-23
Migrate all existing views over to the generic top component framework


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
There is now a generic top component framework which was designed to abstract away a lot of the graph listening and layout issues which always arise when creating a new view in Constellation as well as provide consistent behaviour for each view. It would be ideal if all views in Constellation used this framework in order to take advantage of the consistency it would provide.

Below are the views that still need to migrate to framework:

 * Attribute Editor

 * Conversation View

 * Histogram

 * Named Selections

 * Plugin Reporter

 * Timeline

 * Tutorial Page

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** All views fit it under same generic top component framework
**Actual behaviour:** Views have their own frameworks

**Reproduces how often:** Sometimes

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

aldebaran30701:
Up to date list of views still needing conversion to the Top Component framework
 - Histogram View
 - Timeline View
 - Find View
Web view is currently not used and is under vote for removal.

arcturus2:
Thank you @aldebaran30701 for progressing this. This is a huge leap to standardising the views to use a common framework.

aldebaran30701:
#32 When the remainder of these views are migrated, it will be a good idea to also add the performance feature that avoids view updates when hidden.

GammaVel:
Hello!  When this issue gets fixed also check the default font type and size apply to these views. This is available under Constellation > Setup> Options > Constellation > Application font

PR #554, opened 2020-05-27
Convert Tutorial View to Top Component framework

aldebaran30701:
### Description of the Change
Extract pane out into it's own class and move view to listening framework of JavaFx

### Alternate Designs
N/A

Code consistency

### Benefits
Code consistency

### Possible Drawbacks
None realised

### Verification Process
Opening view

### Applicable Issues
#332

### DIFF
--- a/CoreFunctionality/nbproject/project.xml
+++ b/CoreFunctionality/nbproject/project.xml
@@ -111,6 +111,14 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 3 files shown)


arcturus2:
The Travis build pipeline had an issue today but is now working again @aldebaran30701. So you may not have seen the 10 code smells that existed in the Tutorial Pane. If you have time then could you fix them too? This is obviously not anything you have introduced so I'm OK if you don't want to as well.

Issue #508, opened 2020-05-12
Closing the close graph prompt exits graph without save


aldebaran30701:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Constellation should be consistent with close file behaviour from other systems (Windows apps)
Closing a graph when unsaved changes exist produces the Close graph prompt.
Clicking save produces the Save File Dialog.
Closing the Save File dialog using the X button closes the graph without saving or re-prompting.
Closing the Close graph prompt using the X on windows still produces the Save File Dialog. 

This to me goes against the usability of Constellation, especially coming from Windows where the below behaviour exists:

Close unsaved application
     - prompt Save | Don't Save | Cancel
    
Save - opens Save File Dialog
Don't Save - closes unsaved application
Cancel - disrupts the close event AND the save event
Top right X - disrupts the close event AND the save event

Save File Dialog
Save - Saves file
Cancel - closes Save File Dialog but application with unsaved changes remains open
Top right X - closes Save File Dialog but application with unsaved changes remains open

### Steps to Reproduce
_**Case A**_
1. Open new graph
2. Make a change
3. X on graph tab to close
4. X on close prompt

**_Case B_**
1. Open new graph
2. Make a change
3. X on graph tab to close
4. Save
5. Close Save File Dialog || Press Cancel

**Expected behaviour:**
_**Case A**_
Operation is cancelled. - No Save File Dialog opening

_**Case B**_
Close operation is cancelled unsaved graph still exists

**Actual behaviour:**
_**Case A**_
Save File Dialog opens
![SaveGraph](https://user-images.githubusercontent.com/58677759/81650008-5b771400-9470-11ea-9d5f-221f80ffca42.gif)

_**Case B**_
Unsaved graph closes without prompt
![SaveGraphB](https://user-images.githubusercontent.com/58677759/81650018-5f0a9b00-9470-11ea-9d0e-0b37ef3b5761.gif)
**Reproduces how often:** 
100%

### Additional Information
[...]

PR #557, opened 2020-05-28
Closing the close graph prompt exits graph without save

Auriga2:
### Description of the Change
When closing an unsaved graph, closing the 'Close' dialog no longer opens the 'Save Graph' dialog. 
When cancelling or closing the 'Save Graph' dialog, the unsaved graph is no longer discarded.

### Alternate Designs
N/A

Better UX as this is preventing users from loosing unsaved graphs.

### Benefits
Better UX as this is preventing users from loosing unsaved graphs.

### Possible Drawbacks
N/A

### Verification Process
Case A

Open new graph
Make a change
X on graph tab to close
X on close prompt
-> No Save Graph Dialog opening and close operation is cancelled (unsaved graph still exists).

Case B
Open new graph
Make a change
X on graph tab to close
Save
Close Save File Dialog || Press Cancel
->Close operation is cancelled (unsaved graph still exists)

### Applicable Issues
#508

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
@@ -189,7 +189,7 @@ public class SaveAsAction extends AbstractAction implements ContextAwareAction {
     @Override
     public void actionPerformed(final ActionEvent e) {
         refreshListeners();
-        Collection<? extends SaveAsCapable> inst = lkpInfo.allInstances();        
+        Collection<? extends SaveAsCapable> inst = lkpInfo.allInstances();
         if (!inst.isEmpty()) {
             SaveAsCapable saveAs = inst.iterator().next();
             File newFile = getNewFileName();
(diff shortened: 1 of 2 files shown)


aldebaran30701:
I know this wasn't in the original ticket, but I'm wondering if you'd like to check out if the following can be fixed. When you have an unsaved graph, and close constellation with the X, it prompts the save menu. When you click X (or cancel) on that save menu - it consequently closes Constellation and discards the graph.
If possible, it should cancel the close event - to match the above implementation.

Auriga2:
Hmm it's hard to debug, but nice if we can fix this too.

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
```diff
@@ -181,10 +182,14 @@ public boolean isEnabled() {
...
 
+    public boolean isSaved() {
+        return isSaved;
+    } 
+
     @Override
     public void actionPerformed(final ActionEvent e) {
         refreshListeners();
-        Collection<? extends SaveAsCapable> inst = lkpInfo.allInstances();
+        Collection<? extends SaveAsCapable> inst = lkpInfo.allInstances();
```
cygnus-x-1, line 192: Looks like you've added some random whitespace to the end of this line. It's not a big enough deal to worry about fixing now, but there is an 'on save' option in Netbeans which will prevent this sort of thing next time.

Auriga2: That's a good tip. Thanks. Removed this.

arcturus2:
@aldebaran30701 would you be ok if we merge this PR in and @Auriga2 can look at your suggestion when possible?

aldebaran30701:
Yep I'm fine with this

Issue #331, opened 2020-03-23
Make global parameters singletons


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently, global parameters are being rebuilt inside each plugin which has resulted in discrepancies where the global parameter is not configured the same way within a plugin. 

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** global parameters built outside of plugin

**Actual behaviour:** global parameters built inside each plugin
**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

sol695510:
@antares1470 did you have any suggestions as to where the global parameters should be built?

arcturus2:
Just jumping in here too. For some more background had a look at the following and see how `DateTimeRangeParameterType` is being built again.

https://github.com/constellation-app/constellation-training/blob/32d97260c4e7b1bfeebe3ed30435fd215ce8b2a5/constellation-training-solutions/TrainingSolutions/src/au/gov/asd/tac/constellation/training/solutions/chapter4/ChainCitiesPlugin.java#L60

To be able to use the date range the user wants to query on the `DateTimeRangeParameterType` needs to be available to the plugin. you can see how its being used 

https://github.com/constellation-app/constellation-training/blob/32d97260c4e7b1bfeebe3ed30435fd215ce8b2a5/constellation-training-solutions/TrainingSolutions/src/au/gov/asd/tac/constellation/training/solutions/chapter4/ChainCitiesPlugin.java#L68

So back on line 60 what we should do rather than newing instances is get the existing instance of `DateTimeRangeParameterType` - i.e these should be singletons.

I was thinking of using a builder pattern so when say build - it will either built it or give you an instance of it if the builder has previously built it.

Happy for other suggestions to. Note that global parameters are used in other modules (like the Training module suite) so minimal breaking API changes is highly desirable.

antares1470:
[...]

PR #559, opened 2020-06-02
Make global parameters singletons

sol695510:
### Description of the Change
Moved creation of globally accessed variables to CoreGlobalParameters.java.

### Alternate Designs
N/A

Significant change to how globally accessed variables are created.

### Benefits
Avoids possible discrepancies during singleton configuration.

### Possible Drawbacks
Current solution works but requires a lot of repeated code.

### Verification Process
Executed Data Access View with breakpoints in code to verify actions pass through code correctly.

### Applicable Issues
#331


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/CoreGlobalParameters.java
```diff
@@ -51,7 +55,26 @@
      */
     private static final int DATETIME_RANGE_PARAMETER_ID_INDEX = 1;
     public static final String DATETIME_RANGE_PARAMETER_ID = PluginParameter.buildId(CoreGlobalParameters.class, "datetime_range");
-
+    
+    // Map of string types, datetime types, etc.
+    private static final Map<String,PluginParameter> SINGLETON_MAP = new HashMap<>();
+    
+    public static PluginParameter getStringParameterType(final String PARAMETER_ID) {
+        if (SINGLETON_MAP.get(PARAMETER_ID) == null) {
```
antares1470, line 63: If checking for whether a key exists, better using `containsKey()`. 
i.e. `if (!SINGLETON_MAP.containsKey(PARAMETER_ID)) {`

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/CoreGlobalParameters.java
```diff
@@ -51,7 +55,26 @@
...
+            SINGLETON_MAP.put(PARAMETER_ID, StringParameterType.build(PARAMETER_ID));
+        }
+        
+        return SINGLETON_MAP.get(PARAMETER_ID);
+    }
+    
+    public static PluginParameter getDateTimeRangeParameterType(final String PARAMETER_ID) {
+        if (SINGLETON_MAP.get(PARAMETER_ID) == null) {
```
antares1470, line 71: same here

arcturus2:
@sol695510 this is good. Can you please update the CHANGELOG.md to advertise the new methods and that they should be used instead? Also, can you look at updating CONTRIBUTING.md's checklist section to mention for global parameters the new method should be used.

The training doco no doubt needs to be updated to. For now don't worry about the PDF but for the exercises and solutions, can you please send through a PR using the new methods? The repo I am referring to is https://github.com/constellation-app/constellation-training.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/experimental/TestParametersPlugin.java
```diff
@@ -178,10 +178,10 @@ public PluginParameters createParameters() {
         selected.setDescription("Only use selected elements");
         params.addParameter(selected);
 
-        final PluginParameter<StringParameterValue> queryName = StringParameterType.build(CoreGlobalParameters.QUERY_NAME_PARAMETER_ID);
+        final PluginParameter<StringParameterValue> queryName = CoreGlobalParameters.getStringParameterType(CoreGlobalParameters.QUERY_NAME_PARAMETER_ID);
```
cygnus-x-1, line 181: Rather than having the parameter id's as keys (and having to deal with the issue of figuring out what type of parameter it is), couldn't you just build the parameters upfront and make the actual parameter objects the keys?

sol695510: Is that approach scalable? Could you elaborate?

arcturus2: > What would the "value" be if the object was the "key"?

@cygnus-x-1 this is what I was going to ask too. Please elaborate when you can.

cygnus-x-1: Key was probably the wrong word to use there. If the parameters are built up-front, you could refer to them directly (ie. `CoreGlobalParameters.QUERY_NAME_PARAMETER`), or look them up using their parameter ID. I guess this technically wouldn't make them singletons - however the root issue here was having multiple instances of them built across multiple modules, and that would be addressed. It would also make using them simpler as the developer wouldn't have to know what type they are up-front.

arcturus2:
@sol695510 the change looks good. Are you able to mention something in the change log to let developers know they should change how the global parameters are used.

Would you be able to PR the training module as well?

sol695510:
> @sol695510 the change looks good. Are you able to mention something in the change log to let developers know they should change how the global parameters are used.
> 
> Would you be able to PR the training module as well?

I'm going through changing those at the moments but I've got some more questions about this implementation.

arcturus2:
>Apologies if I'm over complicating this, please clarify @arcturus2 and @cygnus-x-1.

Not a problem @sol695510. So the focus is only on the "Global Parameters" being duplicated. The other plugins having its own parameters is fine. For cases where those plugins use the Date time range for example would need to use the parameter already built rather than creating it again.

With regards to the syntax, I do agree that what you originally had looks more consistent than now doing `CoreGlobalParameters.DATETIME_RANGE_PARAMETER;` but using build for the rest.

What do you think @cygnus-x-1?

sol695510:
> > Apologies if I'm over complicating this, please clarify @arcturus2 and @cygnus-x-1.
> 
> Not a problem @sol695510. So the focus is only on the "Global Parameters" being duplicated. The other plugins having its own parameters is fine. For cases where those plugins use the Date time range for example would need to use the parameter already built rather than creating it again.
> 
> With regards to the syntax, I do agree that what you originally had looks more consistent than now doing `CoreGlobalParameters.DATETIME_RANGE_PARAMETER;` but using build for the rest.
> 
> What do you think @cygnus-x-1?

Thanks for clarifying the parameters @arcturus2. I'll make the changes regarding the DateTime Range Parameter.

sol695510:
> > Apologies if I'm over complicating this, please clarify @arcturus2 and @cygnus-x-1.
> 
> Not a problem @sol695510. So the focus is only on the "Global Parameters" being duplicated. The other plugins having its own parameters is fine. For cases where those plugins use the Date time range for example would need to use the parameter already built rather than creating it again.
> 
> With regards to the syntax, I do agree that what you originally had looks more consistent than now doing `CoreGlobalParameters.DATETIME_RANGE_PARAMETER;` but using build for the rest.
> 
> What do you think @cygnus-x-1?

Just to add, would you prefer a getter method over directly accessing the value?

PR #568, opened 2020-06-05
Fix a bug with split nodes plugin

arcturus2:
### Description of the Change

:bug: Fix a bug with split nodes plugin
* The split nodes plugin was only working splitting for one node. It now works as expected for all nodes selected on the graph.

### Verification Process

Did some test driven development; wrote the unit test and then fixed the code until it passed.

Also did some manual testing too.

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
@@ -157,19 +157,16 @@ public class SplitNodesPlugin extends RecordStoreQueryPlugin implements DataAcce
         result.set(GraphRecordStoreUtilities.DESTINATION + VisualConcept.VertexAttribute.IDENTIFIER, right);
 
         //Loops through all of the Node attributes and copies them to the new node
-        query.reset();
-        while (query.next()) {
-            for (final String key : query.keys()) {
-                if (key.endsWith(".[id]") || SOURCE_IDENTIFIER.equals(key)) {
-                    //Skips the id and Identifier to make the new node unique
-                } else if ((GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.X).equals(key)
-                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Y).equals(key)
-                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Z).equals(key)) {
-                    //The coordinates are also skipped so that the second node is not created in the exact same location
-                    //as the first node
-                } else {
-                    result.set(GraphRecordStoreUtilities.DESTINATION + key.replace(GraphRecordStoreUtilities.SOURCE, ""), query.get(key));
-                }
+        for (final String key : query.keys()) {
+            if (key.endsWith(".[id]") || SOURCE_IDENTIFIER.equals(key)) {
+                //Skips the id and Identifier to make the new node unique
+            } else if ((GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.X).equals(key)
+                    || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Y).equals(key)
+                    || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Z).equals(key)) {
+                //The coordinates are also skipped so that the second node is not created in the exact same location
+                //as the first node
+            } else {
+                result.set(GraphRecordStoreUtilities.DESTINATION + key.replace(GraphRecordStoreUtilities.SOURCE, ""), query.get(key));
             }
         }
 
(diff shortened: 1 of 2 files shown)


CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPluginNGTest.java
```diff
@@ -96,7 +97,7 @@ public void testQueryWithTypes() throws Exception {
         expResult.set(GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.IDENTIFIER, "+123456789");
         expResult.set(GraphRecordStoreUtilities.SOURCE + AnalyticConcept.VertexAttribute.TYPE, AnalyticConcept.VertexType.TELEPHONE_IDENTIFIER);
         expResult.set(GraphRecordStoreUtilities.DESTINATION + VisualConcept.VertexAttribute.IDENTIFIER, "192.168.1.1");
-        expResult.set(GraphRecordStoreUtilities.DESTINATION + AnalyticConcept.VertexAttribute.TYPE, AnalyticConcept.VertexType.URL);
+        expResult.set(GraphRecordStoreUtilities.DESTINATION + AnalyticConcept.VertexAttribute.TYPE, AnalyticConcept.VertexType.HOST_NAME);
```
cygnus-x-1, line 100: Why set the type to HostName when the identifier is an IP address?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -157,19 +157,16 @@ private void editResultStore(final RecordStore result, final String left, final
...
-                    //Skips the id and Identifier to make the new node unique
-                } else if ((GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.X).equals(key)
-                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Y).equals(key)
-                        || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Z).equals(key)) {
-                    //The coordinates are also skipped so that the second node is not created in the exact same location
-                    //as the first node
-                } else {
-                    result.set(GraphRecordStoreUtilities.DESTINATION + key.replace(GraphRecordStoreUtilities.SOURCE, ""), query.get(key));
-                }
+        for (final String key : query.keys()) {
```
cygnus-x-1, line 160: I must be missing something here - but I would have thought that by removing the `RecordStore` iterator you're treating it as a `Record` and so only the one data record currently pointed at will be modified.

arcturus2: The original bug was the the `query` RecordStore (passed by reference) in the this method was looping through to the end which meant that when you returned from `editRecordStore` you were at the end so all other nodes selected would not get a turn.

The intension of this bit of the code was to share the attributes of the node that is being split with the splitter nodes. Therefore we don't need to look and we only care about the keys of the current record.

I wrote a unit test to confirm that it works as expected.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPluginNGTest.java
arcturus2: I didn't think about this change too much though I didn't think it was a problem until your comment. The reference to URL was wrong and the unit test was failing so I changed it to HOSTNAME which is what that IP was resolving to.  Though it should be IP address so I think there is a bug with Types not resolving correctly @cygnus-x-1 ?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
cygnus-x-1: Ahh I see, the iteration is happening elsewhere.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPluginNGTest.java
cygnus-x-1: I don'r see how the type could be causing issues here - the type should be irrelevant to the operation of this plugin. However, if the type is causing issues - then perhaps there is a schema-related bug that needs resolving?

arcturus2: > the type should be irrelevant to the operation of this plugin

Not really @cygnus-x-1, the plugin has to pick a type for the text it splits up.

>then perhaps there is a schema-related bug that needs resolving

Not sure. For the IP 192.168.1.1 there are 3 possible types and using the `AnalyticDominanceCalculator` the one with highest priority is `HOST_NAME` so that's why it's getting picked. Below is the output of the possible types when I run the unit test:

```
Jun 17, 2020 4:34:22 PM au.gov.asd.tac.constellation.graph.schema.type.SchemaVertexTypeUtilities matchVertexTypes
WARNING: Multiple types matched identifier 192.168.1.1: [Network Identifier.IP Address.IPv4 Address, Online Location.Host Name]
```

Note that if I run the 'Extract Types from Text" plugin and put in 192.168.1.1 I get "IPv4 Address" for the type which is inconsistent to Hostname here. The issue there looks like different methods are used. Split nodes uses `SchemaVertexTypeUtilities.matchVertexTypes()` and the applies the `VertexDominantCalculator` whereas "Extract Types from Text" plugin uses `SchemaVertexTypeUtilities.extractVertexTypes()`.

I would suggest this is a bug and should be resolved asap but outside the scope of how Split Nodes plugin works

arcturus2: @cygnus-x-1 I created #596 to continue this discussion. I'm going to merge this PR to fix the original bug and unit test.

Issue #409, opened 2020-04-06
Table View causing application lag


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

On large graphs the table view causes Constellation to lock up in particular when switching between graphs.

During debugging one of the issues seems to be that a separate `String` object is created for each cell rather than reusing objects. This may be due to the table model underneath storing the values as rows rather than columns. If columns were used then it makes it easier for the build in JavaFX code or our code to optimise how the values are stored for that column. For example the `background_icon` attribute could be the same for all nodes and therefore the column could store it once.

Interesting to note that the old Table View which was written in Swing is available from Experimental Views and when that is used instead on the same sized graph, there are no performance issues noticed.

Note that the new table view needs to be closed completely, having it docked or minimised still means that it runs and slows the application down.

### Steps to Reproduce

1. Create a new graph
2. Experimental ->Build Graph -> Sphere Graph
3. Set to 50k transactions
4. Create a new empty graph (now you should have 2 graphs)
5. Open the table view (and notice the memory jump up)
6. Constantly switch between the 2 graphs and see the memory go up and lag Constellation

**Expected behaviour:** The table view should be responsive on large graphs

**Actual behaviour:** The table view slows down the entire application and sometimes runs out of memory.

**Reproduces how often:** 90%

### Additional Information

The performance issue has to do with the size of the graph, the hardware you are running it on and the amount of memory allocated to Constellation.

serpens24:
@arcturus2 - do you have a ballpark feel for how large a graph needs to be to view this issue ?

arcturus2:
@serpens24 I updated the reproducer steps. Setting Constellation with a lower memory limit will reproduce the issue quicker.

serpens24:
Thanks @arcturus2. I have a thought buried deep in the back of my head about how to improve this - as I noted elsewhere (a ticket about sorting the table). I may try and experiment with it a little and see if I can kill two birds with one stone, but suspect it make take a bit of time to get to the bottom of it.

[...]

PR #570, opened 2020-06-09
Table View Pagination

antares1470:
### Description of the Change

Have added pagination to the table view as a way to (hopefully) increase the performance of the table view.

### Alternate Designs

Column storage was considered as a way to increase performance but after doing some research, it was determined that this may not be possible

Improving the performance of table view

### Benefits

Less lag in the table view for larger graphs

### Possible Drawbacks

Having to go through each page to find what you're after may be a tad annoying but is necessary for this to work

### Applicable Issues

#409


antares1470:
Looking to do some further testing and receive feedback before marking this as ready for review

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -941,6 +944,25 @@ private void loadPreferences() {
...
+            final int toIndex = Math.min(fromIndex + MAX_ROWS_PER_PAGE, rows.size());
+
+            table.setItems(FXCollections.observableArrayList(rows.subList(fromIndex, toIndex)));
+        }
+        
+        return new BorderPane(table);
+    }
+    
+    private void paginate(final List<ObservableList<String>> rows) {
+        final Pagination pagination = new Pagination(table.getItems().size() / MAX_ROWS_PER_PAGE + 1);
```
cygnus-x-1, line 960: Have you checked whether sorting and filtering still work across all table data?

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -181,8 +184,8 @@ public TableViewPane(final TableViewTopComponent parent) {
         this.table = new TableView<>();
         table.itemsProperty().addListener((v, o, n) -> table.refresh());
         table.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
-        table.setPadding(new Insets(5));
-        setCenter(table);
+        table.setPadding(new Insets(5));      
+        paginate(null);
```
cygnus-x-1, line 188: Why do you pass null in here initially?

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
antares1470: Haven't gotten to that yet. Like I said previously, there is still a bit of testing to be done and this would be part of it

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
antares1470: The case I happened to be testing before I made the PR was with Table View already open when starting Constellation. Passing null here is more because I couldn't find a better object to pass here (since there doesn't appear to be a useful field I could find containing an array of the data). There probably is a better value to pass.. I just haven't come up with it yet :)

antares1470: Note: If there is a useful field with the data I'm after, there probably won't be a need to pass any parameters into `paginate`

antares1470:
So having done some further testing:
- selection works but if not in Selected Only mode, the selection disappears as soon as you switch pages.
- sorting works but is only applied to the current page. This sorting is lost as soon as you switch pages.
- filtering just doesn't work at all

antares1470:
This is now mostly working. All that needs to be done now is get this all working in selected only mode. It all works when it's not in selected only mode (as far as I'm aware) but behaviour seems to be slightly different in selected only mode and so it is playing up in ways that I wouldn't expect it to play up (I'm still in the progress of figuring out why, open to thoughts from others if they have any)

antares1470:
Glad to say this is finally ready for review :)

formalhaut69:
@antares1470 I think I might be a bit confused about how to use this. I can only seem to get a single page to load and the entire table loads in this page. Am I missing something?
tableViewPagination

antares1470:
> @antares1470 I think I might be a bit confused about how to use this. I can only seem to get a single page to load and the entire table loads in this page. Am I missing something?

So the limit on each page is 500 items, so if your focus is on transactions (which is seems to be in your case), a second page won't be needed unless you exceed 500 transactions

formalhaut69:
> 
> 
> > @antares1470 I think I might be a bit confused about how to use this. I can only seem to get a single page to load and the entire table loads in this page. Am I missing something?
> 
> @formalhaut69 The limit on each page is 500 items, so if your focus is on transactions (which is seems to be in your case), a second page won't be needed unless you exceed 500 transactions

I must look like a bit of an idiot 😆 but its working great though! The page swapping is real smooth. This is a cool feature.

aldebaran30701:
Not sure if this is with the new additions but I managed to duplicate an entry in the table.
I made a new graph, open table view, draw one node, show vertices on table view, switch column order, and pon dragging two times it refreshed underneath itself.
image

antares1470:
Thanks @aldebaran30701 and @formalhaut69 . As an FYI, I managed to replicate the duplicate row issue in master so I don't think that is an issue with my branch. I haven't replicated the null pointer in either master or my branch yet though so looking into that

arcturus2:
Overall this change has helped improve CPU performance as scrolling through a paged list of 500 items is smoother than having to deal with the entire result set.

However, with regards to memory usage after doing some heap dump and other profiling steps both master and the branch seem to be using the same amount. With the addition of the pagination objects the amount of memory has slightly increased with regards to the javafx objects. Given the entire observable list is in memory this is one of the reasons but there may not be much possible here.

Overall though I am happy to merge this in given the investment made to include pagination and this change probably sets us up to do further enhancements. @cygnus-x-1, @aldebaran30701 and @formalhaut69 what are your thoughts?

@antares1470 some minor feedback which I can put into new ticket as an enhancement would be:
* Can we get a jump to first and jump to last buttons too? With 50,000 transactions I can't go to the last entry easily and have to basically reverse the sort order. If these buttons don't come in by default and you have to write custom code just to get these in then don't bother.
* Having a preference to change the default 500 would be useful.

antares1470:
Thanks @arcturus2, I've added a setting to enable the ability to change the page size. I just plucked some numbers out of my head but I can easily add/remove values if needed.

pageSizeDemo

As for your other suggestion, the first/last page button don't come by default (and look like they'll take an effort to include) so I'll create a ticket for it

antares1470:
> I generated a sphere graph, and used the table view. I then closed the graph and table view. When I opened it back up it looked as follows - it still has the pagination in effect.

The reason I believe this is happening is because the pagination isn't being triggered on closing a graph. It all fixes itself though when you open up another graph and so given users are unlikely to be using Constellation without a graph open, I'm somewhat ok with leaving this as is. Curious as to thoughts from others though. @arcturus2 


> Would it be beneficial to store the previous "page size" as a preference for the user?

This is something that could probably be added to the preferences users can store in table view. I don't know whether it would be better to make that change here and create another ticket and do it in a separate PR

arcturus2:
Hi @antares1470 with regards to the zombie pages when there is no graph, I would prefer we fix this as it looks odd and can confuse users. Having the pages go back to 1 is a visual queue of everything being reset properly and I think seeing it like this feels a tad buggy.

antares1470:
So I discovered while adding page size to loaded/saved preferences that I hadn't updated the preferences load to factor in the pagination. Fixing this though has proven to be a challenging task.

Simply calling `paginate()` at the end of `loadPreferences()` wasn't working and I discovered this is because operations aren't happening in the correct order (i.e. paginating is occuring before all other operations are complete). This ends up being because some of the work done in loadPreferences() is done in another thread. More specifically:

`PluginExecution.withPlugin(new TableViewUtilities.UpdateStatePlugin(newState)).executeLater(graph);` which is called at the end of `updateVisibleColumns()` (which `loadPreferences()` calls)

Adding a (less than ideal) 1 second sleep between `loadPreferences()` and `pagiante()` works on smaller graphs but doesn't work on larger graphs (when testing around 50000 transactions) and given this is already a less than ideal solution, I don't want to sleep any longer and would prefer a solution that removes the need for it entirely. Since `executeLater()` returns a Future, I tried a `Future.get()` (which should block until the Future is done) but even that doesn't work (it doesn't look like there are any other threads created as far as I could tell). I've tried a few other things but also with no success.

Anyone have any ideas?

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/io/TableViewPreferencesIOUtilities.java
```diff
@@ -141,8 +146,11 @@ public static void savePreferences(GraphElementType tableType, final TableView<O
                 if (colSortNode.get(sortColumn).asText().equals("DESCENDING")) {
                     sortType = TableColumn.SortType.DESCENDING;
                 }
+                
+                // Extract page size details
+                pageSize = pageSizeNode.intValue();
```
aldebaran30701, line 151: I managed a null pointer here by loading an old preference. Maybe check is necessary?

antares1470: Good pick up. Will add a null check for old preferences

antares1470:
> I also am wondering if you are off by one in the pages?
> 1000 vertices and i can only see 999 on the 10th page.

Only guessing but maybe 999 is the id (id's do start at 0)?
As for the extra page, this is something I've noticed previously when your total number of rows is divisible by the page size. I got the algorithm for this on the web but I could potentially take a closer look to see why (I suspect it is the +1)

PR #571, opened 2020-06-09
map-view-fix

cygnus-x-1:
:art: removed reference to `ToolOverlay` from `MapViewTileRenderer` in order to keep it loosely coupled


CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
```diff
@@ -985,13 +986,4 @@ public String toString() {
             return String.format("%s: %d", name, count);
         }
     }
-
-    public boolean isIgnoreMapInteractions() {
```
arcturus2, line 989: @cygnus-x-1 wouldn't it be better to keep this method as this bit of code is duplicated twice within this class? It can be instead `private` though.

arcturus2:
Thanks for doing this @cygnus-x-1. This was something I said I would do. Only a minor comment for your review.

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
cygnus-x-1: Perhaps - there are other code blocks just like it all throughout `MapViewTileRenderer`, the whole thing needs a clean up but that feels like a task for another day.

arcturus2: True @cygnus-x-1 but this would be one less we would have had to do.

PR #576, opened 2020-06-11
Layers View select layer twice to enable - fix.

aldebaran30701:
### Description of the Change
Layers View has an issue where if you are running a dynamic query, it will require selection, and deselection to have the graph refresh. This was fixed by adding a future to ensure that the call to submit the queries was finalized before executing the visual changes.

The other section of this PR is to seek assistance in finding an odd bug. 
The bug relates to vertex layer visibility and vertex layer bitmask values updating on unrelated graph changes.
layers
To get to this state, I added 3 elements, added query "color == #ffffff" <#ffffff should be replaced with a correct node colour>
I then selected that layer, then changed that HEX value to one of the other elements. I then could get the element to appear on and off of the graph depending on graph changes such as camera movements. I deleted the other elements for simplicity.

GraphVisualAccess contains a comment on the change, where you can detect this behaviour. FloatAttributeDescription never sets the float value to anything when this behaviour occurs, which leads me to believe it is something weird happening elsewhere. 

### Alternate Designs
N/A

Bug Fix TBA

### Benefits
Bug Fix

### Possible Drawbacks
None realized

### Verification Process
Visual testing

### Applicable Issues
N/A

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
@@ -15,8 +15,6 @@
  */
 package au.gov.asd.tac.constellation.graph;
 
-import au.gov.asd.tac.constellation.graph.operations.GraphOperation;
-import au.gov.asd.tac.constellation.utilities.datastructure.IntHashSet;
 import static au.gov.asd.tac.constellation.graph.GraphConstants.NOT_FOUND;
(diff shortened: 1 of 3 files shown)


aldebaran30701:
When I get this behaviour to occur, I can save the graph, and reload it with the correct layer behaviour. this leads me to believe it is some sort of odd behaviour with the state kept in a store graph.

aldebaran30701:
Closing this as it is superseded by #602

Issue #550, opened 2020-05-25
Fix to broken unit tests


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The unit tests have been broken for a little while it seems. Travis has been silently failing since migrating to NB 11.

This is current master

<img width="1280" alt="image" src="https://user-images.githubusercontent.com/39325530/82662888-00b09a00-9c72-11ea-9f31-6aef773fedcb.png">

This is 6b4df60 (v2 beta)

<img width="1280" alt="image" src="https://user-images.githubusercontent.com/39325530/82822182-b2f19700-9ee8-11ea-8ddd-78d818ffd7b6.png">

mimosa2:
Looks like the expected result and testDateTime offset needs to have colon instead of a dot.
final String expResult = "2019-04-23 01:16:42.000 +**10:00** [Australia/Sydney]";
String testDateTime = "Tue Apr 23 01:16:42 +**10:00** 2019";
final String expResult = "2019-04-23 01:16:42.000 +**10:00**";
final String expResult = "2019-12-25 00:00:00.000 +**00:00**";

mimosa2:
testDateTime Should be 
String testDateTime = "Tue Apr 23 01:16:42 +1000 2019";

arcturus2:
After PR #579 there are only 4 left to fix

<img width="1619" alt="image" src="https://user-images.githubusercontent.com/39325530/84624718-4258fb80-af25-11ea-9e5f-b363cecfa9dd.png">

PR #579, opened 2020-06-15
Fix to ModCount Tests

arcturus2:
### Description of the Change

Unit tests in `ModCountNGTest` was failing because the mod counter was being increased by 1 sometimes due to layer attributes being added to the `StoreGraph`.

These unit test now pass and the change made was to change calls to `ensure()` to `get()` in `StoreGraph`. The ensuring now happens in the `LayersController`.

This means that if you open a new graph with no other views open and add attributes, nodes are added with no understanding of layers (i.e same as before the introduction of the Layers View).

If you open the Layers view to play with layers then required attributes are added to the graph to make the layers view work.

The context menus to add and remove nodes from layers work with the layers view opened or not and based on testing seems to function as before.

Also added `LayerConcept` to group all of the layer mask and layer visibility attributes together. 

---

This PR also fixes a unit test with `IONGTest`. If the default value of a StringAttributeDescription attribute is null then it should honour that and be allowed. The use of `StringUtils.isBlank()` prevented that from happening.

### Verification Process

I've tested adding layers to nodes with the layers view open and when it was closed and all scenarios seem to work and the layers view functions as expected.

### Applicable Issues

#550


CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
```diff
@@ -2223,43 +2200,13 @@ private void updateLayerMask(final int attributeId, final GraphElementType eleme
...
-                    LAYER_MASK_ATTRIBUTE_LABEL,
-                    "Bitmask identifying the layers this transaction belongs to", 1, null);
-        }
+        layerMaskSelectedAttributeId = LayersConcept.GraphAttribute.LAYER_MASK_SELECTED.get(this);
+        vertexLayerMaskAttribureId = LayersConcept.VertexAttribute.LAYER_MASK.get(this);
+        vertexLayerVisibilityAttributeId = LayersConcept.VertexAttribute.LAYER_VISIBILITY.get(this);
+        transactionLayerMaskAttrId = LayersConcept.TransactionAttribute.LAYER_MASK.get(this);
+        transactionLayerVisibilityAttributeId = LayersConcept.TransactionAttribute.LAYER_VISIBILITY.get(this);
+
+        // TODO: not sure about why we have to keep id's recorded
```
aldebaran30701: The idea here was so that we didn't call to update the layer bit masks of Vertices or Transactions when an update to their Id was called. Every other instance that a Vertex or Transaction was altered will result in an update here. Just prevents a loop: update color -> trigger bitmask recalc -> change vertex bitmask -> trigger bitmask recalc. Value is set again and it continues to loop.

arcturus2: Thanks @aldebaran30701. I updated the PR by removing the TODO comment but you saw it already. In that case a follow up question would be if all you want to do is see if the attribute is the vertex/transaction layer visibility attributes then rather than storing a copy of the attribute can we get it from the attribute description like:

```java
if (!LAYER_VISIBILITY_ATTRIBUTE_LABEL.equals(getAttributeDescription(attributeId))){
...
}
```

aldebaran30701: That looks like it would work, if i remember correctly I check those Ids in ```updateLayerMask()``` somewhere, if that is the case would it be alright if we checked those the same way? Just wondering from a performance standpoint.

arcturus2: >if that is the case would it be alright if we checked those the same way?

Yep we can. Though I'll do this is a separate PR as my focus is to get the unit tests over the line. It will also mean this PR is not making too many changes.

> Just wondering from a performance standpoint

Good point to keep in mind. I will do some profiling to compare both cases though I think it will be negligible because it will be looking up arrays, though the String compare will be the only issue.

arcturus2:
> it seems to just work somehow

Yeah I was expecting I had to change the context menus and initialise them with the attributes but your code worked as is so I was happy. I think the defencing programming checking if the attributes exists helps here.

Issue #574, opened 2020-06-11
Import invalid XML file (applies to all delimited import parsers)


serpens24:
### Description

Created a text file containing the following:
```
This is bad data
This is bad data
This is bad data
This is bad data
```
Try to import it using the XML importer - exceptions occur, then the importer reports successful import with a heap of exceptions.

### Steps to Reproduce

1. Create a file containing the above lines and save as baddata.xml
2. Select File->Import->From Delimited File...
3. Set Import file Parser to XML, then press the "+" button and locate the file, and press Open.
4. Acknowledge warning dialog twice (this is an issue - see below)
5. On the main import dialog press the Import button. (this is enabled despite no data being valid in file)
**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** 
At step 3, the following dialog is shown (twice)
![image](https://user-images.githubusercontent.com/58677108/84336132-c3d32580-abd5-11ea-96b9-a2255e999078.png)
After acknowledging twice, user is returned to main Import dialog, with the Import button enabled .... 

At step 5, despite data being invalid - so nothing to import, a dialog is raised suggesting "X" lines of data were imported along with multiple Exception dialogs. See below:
![image](https://user-images.githubusercontent.com/58677108/84336391-4d82f300-abd6-11ea-9d3d-7ff164b1f744.png)

**Reproduces how often:** 100%

### Additional Information

One of the issues here is the code generating the "Success dialog" is independant of the actual import and just counting lines of text in the file.

So there are really 4 problems:
1. Invalid XML files don't get neatly handled up front ... a single warning should be displayed.
2. The import button is enabled (I can live with that if it doesn't die ugly) - but when clicked tries to import invalid data.
3. Press the import button and excpetions are thrown (ugly) 
4. The number of "rows" imported is in no way correlated to rows of records .... in fact, this occurs even if the import XML is successful as the counter is assuming one record per line is is the case for CSV files. Surely we can get the number of records from the importer record list instead ?

This problem really needs to be checked for all importers in case there is similar.

Note: I fond the warning dialogs seem to pop up behind the import dialog generally so are hard to see... which means context is taken away from the user and the tool becomes un-useable until usr clicks back on the main window ... at which point the warning dialogs appear.

Note 2: The "Success" dialog is displayed before the data is imported.... surely you can only show success after the import is completed.

serpens24:
[...]

PR #581, opened 2020-06-16
Feature/issue 468 json ingest

serpens24:
### Description of the Change

Added JSON ingestion code + tests.

### Alternate Designs

Nil

Allow JSON to be imported

### Benefits

As above.

### Possible Drawbacks

none known.

### Verification Process

Refer to unit tests. But eesntially rotate through a heap of JSON files and attempt to import. confirm those with lists that match criteria spin out columns and rows of data.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/574


CoreImportExportPlugins/test/unit/src/au/gov/asd/tac/constellation/plugins/importexport/JSONImportFielParserNGTest.java
```diff
@@ -0,0 +1,429 @@
+/*
```
cygnus-x-1, line 1: Please add the license header here.

serpens24:
@arcturus2 _ SonarCloud is failing due to no coverage of new code. I don't beleive this as I've written a dedicated unit test. Does something need to coccur to ensure unit test is hooked into overall test set ?

arcturus2:
> @arcturus2 _ SonarCloud is failing due to no coverage of new code. I don't beleive this as I've written a dedicated unit test. Does something need to coccur to ensure unit test is hooked into overall test set ?

Your right @serpens24 something has gone wrong. When I had a closer look at the logs, unit tests fail after the 4 module so it never got to run your unit tests; and therefore no code coverage information is available for computation.

Also this is really amazing work @serpens24 and the unit tests are pretty good. This is the ideal standard I would like to get to with PRs (including my own!).

@cygnus-x-1 it looks like the problem is that there `CoreGraphNode` has no unit tests and so the "BUILD FAILED". See https://api.travis-ci.com/v3/job/349523275/log.txt. We already now that some modules have no unit tests so I would have expected it would continue and fail the build when a unit test fails. Instead that does not seem to be the case. If you have time to investigate this let me know or I can have another go at it.

arcturus2:
This is amazing stuff @serpens24. Thank you so much for adding in the JSON importer!

arcturus2:
> Your right @serpens24 something has gone wrong. When I had a closer look at the logs, unit tests fail after the 4 module so it never got to run your unit tests; and therefore no code coverage information is available for computation.

Unit tests are passing @serpens24 and you can now see the code coverage from the extensive unit tests you wrote https://sonarcloud.io/component_measures?id=constellation-app_constellation&metric=coverage&selected=constellation-app_constellation%3ACoreImportExportPlugins%2Fsrc%2Fau%2Fgov%2Fasd%2Ftac%2Fconstellation%2Fplugins%2Fimportexport%2Fdelimited%2Fparser%2FJSONImportFileParser.java&view=list

Issue #32, opened 2019-07-26
Investigate views to not listen to events when they are docked


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a developer
I want docked windows to not listen to graph updates
so that I dont waste computational resources updating views that are not visible.

The idea with this was to reduce the computational load by not having views do work when they are not visible.

To see more debug information of listening events enable the following lines in `project.properties`:

`
run.args.extra=-J-Dau.gov.asd.tac.constellation.functionality.views.ListeningTopComponent.level=FINE \
-J-Dau.gov.asd.tac.constellation.graph.node.pluginframework.FrameworkGraphManager.level=FINE
`

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

antares1470:
@aldebaran30701 Can this ticket be closed or is there something else that needs to be done for it?

aldebaran30701:
I forgot to update this ticket. The below views are not implemented in this ticket for the following reasons.
- Find view - Not currently on Generic TopComponent Framework
- Histogram - Not currently on Generic TopComponent Framework
- Timeline - Not currently on Generic TopComponent Framework
- NamedSelection - Utilises an odd listening framework for graph listening and could not implement without bugs

I feel like we can close it, and I can link the tickets to the relevant tickets that will need to be updated when the views are converted.

arcturus2:
Let's close it given the above points @aldebaran30701 mentioned.

PR #582, opened 2020-06-16
Avoid view updates when hidden

aldebaran30701:
### Description of the Change
This is a feeler PR for the way it is implemented and if approved, will allow me to extend it to any view requiring an update. 
Please pay attention to the section where booleans are set as I am not 100% sure on what sequence of events will need to occur when a view is hidden but floating?

The view framework now has an easily accessible boolean flag for testing if it isvisible or not.

**~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~**

**EDIT:**
Implemented in all views except the following few below:

- Find view - Not currently on Generic TopComponent Framework
- Histogram - Not currently on Generic TopComponent Framework
- Timeline - Not currently on Generic TopComponent Framework
- NamedSelection - Utilises an odd listening framework for graph listening and could not implement without bugs
- Quality Control - Once DAV is opened, QCV launches an instance and ALWAYS refreshes on graph changes.

I worked on getting QCV avoiding updates when none of (QCV or DAV) were opened, but struggled with cyclic dependencies. I pose this as a question to wether this should be something tackled in another issue, or if the default behaviour is acceptable.

The others will need to be switched over to the generic framework as in #332 

### Alternate Designs
None realised

Performance improvements

### Benefits
Performance improvements

### Possible Drawbacks
none realised

### Verification Process
Testing of update within Table view when it is visible or not.

### Applicable Issues
#32


arcturus2:
I will have a closer look at this @aldebaran30701. It was good you put this in the `AbstractTopComponent`.

arcturus2:
@aldebaran30701 Travis is failing here too. Can you please marge changes from master?

aldebaran30701:
I've implemented most of the views to not listen when minimized, and tested switching graphs as well as reloading the view to ensure with the correct graph's state.

arcturus2:
@aldebaran30701 I was doing testing on the Table View and noticed that it was still updating when I put some break points in. I'll let you know what I find.

aldebaran30701:
Hi @arcturus2 It appeared that I assumed I had implemented the table view, and skipped looking through all cases when I went to fully implement this.
I've added the case now and it should be ready for review. 👍

arcturus2:
>    Quality Control - Once DAV is opened, QCV launches an instance and ALWAYS refreshes on graph changes.
I worked on getting QCV avoiding updates when none of (QCV or DAV) were opened...I pose this as a question to wether this should be something tackled in another issue, or if the default behaviour is acceptable.

I've had a think about this and for now lets deal with this one later @aldebaran30701.

Also, I've also done a lot of testing and this PR will help a lot with less CPU cycles! I've made some suggestions to this PR and created PR #626 with suggestions. Note that there is way more files in my PR which has changed but that was due to me having to merge in `master`. If you merge `master` into this PR then you will be able to see the actual suggestions I'm proposing.

Issue #577, opened 2020-06-12
Run_plugin endpoint in the Rest API should return a valid error when an incorrect graph_id is supplied


Auriga2:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
If you specify an invalid value for `graph_id - The id of a graph to run the plugin on (optional)`
it doesn't run the plugin, but returns 200, as if it was successful. Instead it should return error 422.
(The valid graph_id could be retrieved from the `get_graph` endpoint.)

![image](https://user-images.githubusercontent.com/64826928/84459666-2ba57100-aca7-11ea-9d22-c840c69e0063.png)

### Steps to Reproduce

1. Load the Rest API (Tools-Start REST Server)
2. Try out POST - `run_plugin` with an invalid text for `graph_id`

**Expected behaviour:** Return error 422.

**Actual behaviour:** Returns 200, as if it was successful.

**Reproduces how often:** 100%

Auriga2:
Current code doesn't support custom responses. In a similar scenario,  in SetGraph.java, it throws `RestServiceException` with 500 code.

arcturus2:
Thanks for the investigation @Auriga2. Based on this we would need an elegant solution to refactor the code to be able to support custom HTTP error codes. It looks like you have assigned yourself to this one which is cool. If you want to have a discussion around implementations feel free to do a mini write up here or if its easier to just send a PR that is fine too.

In case if @algol60 is interested whom recently refactored the Rest API code then feel free to make suggestions if you had already thought about this but decided it was not necessary at the time to implement customised HTTP exit codes.

Auriga2:
I managed to get it to work by passing the `HttpServletResponse response` object to the services classes. But tricky to refactor the FileListener.java class.

![image](https://user-images.githubusercontent.com/64826928/84625506-8dbdda80-af22-11ea-8322-f671bab7eed7.png)

arcturus2:
>But tricky to refactor the FileListener.java class.

[...]

PR #583, opened 2020-06-17
Rest API returns custom error codes and Run_plugin endpoint returns a valid error when an incorrect graph_id is supplied.

Auriga2:
### Description of the Change

* When any errors are thrown from the services, the Rest API was throwing them only as Error 500.
With this fix we can response with custom error codes.

* Run_plugin endpoint returns a valid error when an incorrect graph_id is supplied.

### Alternate Designs
I managed to get it to work by passing the HttpServletResponse response object to the services classes. But tricky to refactor the FileListener.java class.

Better UX

### Benefits
Better UX

### Possible Drawbacks
N/A

### Verification Process
* In the Rest API, try the `SetGraph` endpoint with an invalid graph ID. It should return an Error 422 instead of 500
* Try the `RunPlugin` endpoint with an invalid graph ID. It should return an Error 422 instead of 200

### Applicable Issues

#577


CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestService.java
```diff
@@ -38,9 +38,56 @@
  * <p>
  * All services are accessed via RestServiceServlet; see that for more details.
  *
+ * This also stores status codes.
+ * Top section copied from public interface HttpServletResponse.
+ * 
  * @author algol
  */
 public abstract class RestService {
+    public static final int SC_CONTINUE = 100;
```
arcturus2, line 47: Can you please use the `HttpURLConnection` class instead of creating these constants?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/AddRecordStore.java
```diff
@@ -182,6 +183,7 @@ public void callService(final PluginParameters parameters, final InputStream in,
         }
 
         addToGraph(graphId, rs, completeWithSchema, arrange, resetView);
+        return new ServiceResponse(SC_OK, "Successful");
```
arcturus2, line 186: So in here it would be `HttpURLConnection.HTTP_OK`.
What is the reason to show the message "Successful" given the HTTP code will inform this. Does the message appear in the response output?
Also if you are going to keep it then its work making the string a constant.

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestService.java
Auriga2: 422 is not there. Shall I leave only that in the new class then?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/AddRecordStore.java
Auriga2: @arcturus2 It's not really used afterwards, as I am processing only if it's not a 200. I just typed it thinking it's better than an empty string. What do you think?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestService.java
arcturus2: Yep @Auriga2. Sorry I didn't realise 422 was not there.

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/ServiceResponse.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.webserver.restapi;
+
+/**
+ * This class is used to send custom HTTP error codes generated from the 
+ * callService in webserver.services classes.
+ */
+public class ServiceResponse { 
+    private String errorMessage;
```
arcturus2, line 23: how about `statusMessage`?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/ServiceResponse.java
```diff
@@ -0,0 +1,46 @@
...
+    private int errorCode;
```
arcturus2, line 24: how about `statusCode` given the response is not always when there is an error.

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/AddRecordStore.java
arcturus2: If its not used then how about creating another constructor like this?

```
public ServiceResponse(int errorCode){
        this.errorCode = errorCode;
        this.errorMessage = null;
}
```

Auriga2: Agree, fixed.

Auriga2:
Fixed the above issues and also noticed that `set_graph` endpoint was still throwing 500 if no graph was loaded in Constellation. Added a check with `GraphNode.getAllGraphs().isEmpty()` but currently the PUT method doesn't show a custom message because it doesn't have a response body.
image

arcturus2:
Thats awesome stuff @Auriga2. When you are ready please take it off draft and i will ask @algol60 to also do a review too.

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/api/RestServiceServlet.java
```diff
@@ -101,7 +102,10 @@ private void callService(final HttpMethod httpMethod, final HttpServletRequest r
         try {
             response.setContentType(rs.getMimeType());
             response.setStatus(HttpServletResponse.SC_OK);
-            rs.callService(parameters, request.getInputStream(), response.getOutputStream());
+            ServiceResponse serviceResponse = rs.callService(parameters, request.getInputStream(), response.getOutputStream());
+            if (serviceResponse.getStatusCode()!=200){
```
arcturus2, line 106: @Auriga2  should the `200` be replaced with `HttpURLConnection.HTTP_OK`?

Auriga2: Yes Thanks! Fixed.

arcturus2:
So @Auriga2 your change is pretty good. After discussing with @algol60 a suggestion was made if we could try something different. 

So, the idea would be to reuse the `RestServiceException` instead of creating `ServiceResponse`. The `callService` would go back to being return `void` but we'd instead throw a `RestServiceException` when things go wrong. The `RestServiceException` will need to be modified to pass an `int` like your `ServiceResponse` class.

What are the benefits of this approach? Well, the change is to throw an exception instead of returning a HTTP code when something goes wrong. Some other benefits is because `RestServiceException` is a `RuntimeException` all serverlets can remain unchanged (i.e they don't need to declare `throws RestServiceException`). It will also mean that the original functionality of when there are no exceptions everything is assumed to be a HTTP 200 so that you don't need to add the return call in all of those serverlet.

It would mean that the change will be changing perhaps 4 files rather than 25 files. Noted that you have already made the change but I would like to propose @algol60's solution too.

If you are happy to continue working on this PR and make the above change then your welcome to. If however you don't feel like doing this suggestion then that is totally fine; I'm happy to do it and put a PR for you to see.

Auriga2:
No worries @arcturus, but I cannot figure out how throw a specific number error with that because the super (RuntimeException) doesn't support that. Happy to improve it if you can give me a clue.

arcturus2:
Hi @Auriga2 you could do it like this:

```java
public class RestServiceException extends RuntimeException {

    private final int code;
    
    public RestServiceException(final int code, final String message) {
        super(message);
        this.code = code;
    }
    
    public RestServiceException(final String message) {
        super(message);
        this.code = HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
    }

    public RestServiceException(final Exception ex) {
        super(ex);
        this.code = HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
    }
    
    public int getHttpCode(){
        return this.code;
    }
}
```
In the servelets when you throw the exception using the one allowing you to pass the `int`.

Then in `ConstellationApiServlet` in the `doGet` and `doPut` you can add the extra catch for the `RestServiceException`

```java
            try {
                get(request, response);
            } catch (final RestServiceException ex) {
                response.reset();
                response.sendError(ex.getHttpCode(), ex.getMessage());
            } catch (final IOException | ServletException ex) {
                response.reset();
                response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, ex.getMessage());
                LOGGER.log(Level.INFO, "in doGet", ex);
            }
```

Thanks to @algol60 for the suggestion!

Auriga2:
Thanks for the suggestion @algol60 and @arcturus2. Yes it's much better.
Let me know when this is reviewed, I can apply it to other endpoints as well. (Or create a case to fix later)

arcturus2:
@Auriga2 thanks for the update

arcturus2:
> Let me know when this is reviewed, I can apply it to other endpoints as well. (Or create a case to fix later)

@Auriga2 I've reviewed it. Which endpoints are left to do?

Auriga2:
@arcturus2 For instance `get_attributes` throws `NullPointerException` for invalid `graph id`:
image

Other endpoints would still throw 500 for invalid user input.   
E.g.: `Open Graph`:
image

arcturus2:
Thanks understood @Auriga2. I'm OK with either option; keep working on it in this PR or a new issue + PR. Let me know and based on this we can either merge this PR in or not.

Auriga2:
Following endpoints do not show the message because of missing `Response body`. I couldn't figure out the cause. 
Get_graph_image
Get_icon
Set_graph

image

@algol60  Do you know how to fix that? I feel like it'd be nice if we can show the failure reason to the user.

Auriga2:
get_plugin_description is throwing nullpointer exception for any valid plugin with no parameters. E.g.: SELECTALL, UNDIMALL, CONTRACTGRAPH
image

 Plugins with parameters E.g.: RESOURCEALLOCATIONINDEXANALYTIC, STRUCTUREDGRAPHBUILDER work.

Fixed that also here.

Auriga2:
I didn't add a check for missing parameters. Feels like they should also be marked as `*required` similar to the `CONSTELLATION secret`
image
May be we can create a low priority case to fix later.

arcturus2:
>Following endpoints do not show the message because of missing Response body. I couldn't figure out the cause.
Get_graph_image
Get_icon
Set_graph

I couldn't figure it out either but what you've done is fantastic.

> I didn't add a check for missing parameters. Feels like they should also be marked as *required similar to the CONSTELLATION secret

In the `SwaggerServlet` I saw
```
param.put("required", false); // TODO Hard-code this until PluginParameters grows a required field.
```
so once we implement the required property for plugins we should be able to get this automatically in Swagger.

Auriga2:
Created https://github.com/constellation-app/constellation/issues/633 to fix the above issue to mark as *required.

PR #587, opened 2020-06-17
Fix broken unit tests

mimosa2:
### Description of the Change

DateTimeFormatter fails after the Netbeans/Java upgrade.


arcturus2:
Hi @mimosa2 it looks like there is a merge conflict. Can you fix that so we can merge it?

Issue #591, opened 2020-06-19
SceneBatcherNGTest Failing


Nova-2119:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The unit tests for the class SceneBatcher in the class SceneBatcherNGTest are failing. 

### Steps to Reproduce
NA

**Expected behaviour:** [What you expect to happen]
The unit tests in SceneBatcherNGTest are all passed.

**Actual behaviour:** [What actually happens]
Some unit tests in SceneBatcherNGTest are failing.
**Reproduces how often:** All of the time

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Nova-2119:
The file actually tests LabelUtilites. Some of the tests that pass don't test what the name and comment suggest it tests.
The fail seems to be genuine though, there seems to be a bug in splitTextIntoLines. Will comment again when confirmed.

PR #594, opened 2020-06-19
Bugfix/scene batcher issue591

Nova-2119:
### Benefits

Attribute values to big to fit on a single line will now be split on spaces if possible. Previously these would have been split in a word.

### Possible Drawbacks

Unknown

### Applicable Issues

https://github.com/constellation-app/constellation/issues/591


arcturus2:
Thanks for this fix @Nova-2119

PR #607, opened 2020-06-26
Update ui icons

formalhaut69:
### Description of the Change
Made new icons for the help section of Constellation.

image

### Applicable Issues

The help contents, check for updates and about page are missing an icon. I made icons for them but I couldn't add them in because these appear to be inbuilt netbeans functions that can't have an icon added to them. Tutorial page is also missing an icon but this is because there is no action class for the tutorial page so I wasn't able to add an icon to it. I added an action registration to the tutorialTopComponent class with an icon in it but no luck. Any suggestions would be appreciated.


arcturus2:
@formalhaut69 I'd like to better understand your rational/motivation/vision for using gray icons with a bit of colour. Are you following a particular theme?

Also, can you include a screenshot of this in the dark theme?

Thank you.

formalhaut69:
@cygnus-x-1 @arcturus2 I chose gray because it feels like I guess a system colour and it works well in both light and dark mode.
image

image

I'll update the show logs icon I've noticed that there's a thin gray line inside the eye that's only noticeable in dark mode and it distorts the icon a bit.

arcturus2:
ok @formalhaut69 let us know when its ready to merge!

formalhaut69:
@arcturus2, @cygnus-x-1 I've updated the last few icons in the file and edit drop downs. I'm not really happy with how its turned out though. I made icons for all of these but the majority of these were inbuilt netbeans actions so I was unable to update them so it just looks a bit empty. I've also fixed the show logs icon so this is ready to merge if you guys are happy.
image
image
image

PR #615, opened 2020-07-01
Fix for for manage templates layout

formalhaut69:
### Description of the Change

Fixed the layout issue of the manage templates window. The buttons couldn't be clicked and were pushed off the edge of the screen and had unusual colons overlapping with the label text.
image

### Possible Drawbacks
The buildParameterLabel method in the PluginParametersPane class had to be modified. The only pane that I am aware of that uses this is the analytic view pane and that seemed to be visually unaffected so I'm unsure of how other panes have been impacted.


arcturus2:
>The buildParameterLabel method in the PluginParametersPane class had to be modified. The only pane that I am aware of that uses this is the analytic view pane and that seemed to be visually unaffected so I'm unsure of how other panes have been impacted.

Thanks for this @formalhaut69. So this template is used by the graph actually. So you can change the graph attributes (from Views -> Attribute Editor) and change the background graph colour to purple say and then save that as a template.

If you set that as the default then when you open a new graph it will be used instead of the original Analytic Graph template.

PR #616, opened 2020-07-01
Null check and logging for plugin execution

aldebaran30701:
### Description of the Change
This change adds a null check to the graph parameter and in the case of run later, stops execution and logs the null graph error.
The change also includes a null check to runNow methods to log when the graph which is passed in is null.

### Alternate Designs
Catch failing silently - Unsure where this was meant to be occurring. 

Additional logging

### Benefits
Additional logging


CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginEnvironment.java
```diff
@@ -58,6 +58,10 @@
     @Override
     public Future<?> executePluginLater(final Graph graph, final Plugin plugin, final PluginParameters parameters, final boolean interactive, final List<Future<?>> async, final PluginSynchronizer synchronizer) {
 
+        if (graph == null) {
+            LOGGER.log(Level.WARNING, plugin.getName() + " plugin was executed on a graph which was null");
```
arcturus2, line 62: getting super picky here but I think NetBeans provides an optimisation here to convert the string concatenation with a message template

`LOGGER.log(Level.WARNING, "{0} plugin was executed on a graph which was null", plugin.getName());`

PR #622, opened 2020-07-03
Dav help fix

formalhaut69:
### Description of the Change
I've made a small fix so that the help icon for "Range" in the data access view is functional. 

### Applicable Issues

I can't seem to figure out how to prevent it's help icon from overlapping on top of the months buttons so I haven't resolved the entire issue yet. The problem I ran into is that the icon is added using a separate template within the PluginParametersPane so I couldn't figure out how to control how its added to the pane. Does anyone have any suggestions?

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
@@ -180,7 +180,7 @@ public class DataAccessPane extends AnchorPane implements PluginParametersPaneLi
                 .addListener((oldValue, newValue) -> update());
 
         final Button helpButton = new Button("", new ImageView(UserInterfaceIconProvider.HELP.buildImage(16, ConstellationColor.BLUEBERRY.getJavaColor())));
-        helpButton.paddingProperty().set(HELP_INSETS);
+        helpButton.paddingProperty().set(new Insets(0, 8, 0, 64));
         helpButton.setTooltip(new Tooltip("Display help for Data Access"));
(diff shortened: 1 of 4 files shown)


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
```diff
@@ -180,7 +180,7 @@ public DataAccessPane(DataAccessViewTopComponent topComponent) {
                 .addListener((oldValue, newValue) -> update());
 
         final Button helpButton = new Button("", new ImageView(UserInterfaceIconProvider.HELP.buildImage(16, ConstellationColor.BLUEBERRY.getJavaColor())));
-        helpButton.paddingProperty().set(HELP_INSETS);
+        helpButton.paddingProperty().set(new Insets(0, 8, 0, 64));
```
antares1470, line 183: Set `HELP_INSETS` to this new value and then use the constant

formalhaut69: @aldebaran30701 made that change to only fix the options help icon. Changing the constant to be the new value will modify all of the padding to all of the icons which pushes them into an awkward position. Would it be a good idea to create a separate constant and store the value in this just for the options help icon?

antares1470: Ah ok. When I did my review I only saw the constant only being applied here so I must've missed that.
For what it's worth, I probably wouldn't bother creating a constant simply just for this unless you reckon this value will be used else where.

PR #602, opened 2020-06-24
Modify storage of Layers view queries and preferences

aldebaran30701:
### Description of the Change
This change fixes #589 and implements a future on saving of queries to ensure only one selection is needed to have a layer become activated.

LayerQueries list which used to be stored inside of StoreGraph was causing issues as the DualGraph managed two instances of StoreGraphs which were out of sync - causing odd behaviour like changing of attribute values. 
The fix stores the list of queries inside a GraphAttribute Object. The same migration process was done to LayerPreferences to fix an issue which arose from the above implementation. 

### Alternate Designs
Trigger a commit whenever one StoreGraph layerQueries list was updated, to force the other StoreGraph within the DualGraph to sync with the updated one. I did not get far into this idea before running into problems.

### Why Should This Be In Core?
Bug Fix - enhances the usability of layers view

### Benefits
layers now work on first click, and react normally because the two graphs are in sync

### Possible Drawbacks
None realised

### Verification Process
Testing Queries
Add color and other attribute queries to each layer, verify correct results
2.	selecting a section of the graph, right click context menu -> add to a vacant layer
a.	Ensure the selection is correct with what is added to the layer
3.	Select more elements and test remove functionality
a.	Ensure the elements are removed from that layer
4.	Test layers with short or long names for visuals and any errors.
Testing Draw mode
1.	Create a new graph
a.	Add a few elements and transactions on
i.	Should appear visible
b.	Open layers view
i.	Select layer 2 only
ii.	Add elements (Vertices and Transactions)
iii.	Elements should only appear on layer 2

### Applicable Issues
#589

PR #625, opened 2020-07-05
Convert PR #602 to use a DualGraph

arcturus2:
### Description of the Change

This is a suggestion to #602 for @aldebaran30701.

>I'd like to help out to see if I can give some guidance to get the graph switching part to work
@arcturus2 I think I'll take you up on that offer - Above is the branch I've been testing with.

So I noticed you've used a `StoreGraph` but what I think your after is testing a `DualGraph` and test the bug with syncing graphs. To do it yourself with a `StoreGraph` and mimic the sync will effectively be implementing the `DualGraph` in a unit test so I've just used a `DualGraph`.

As a result the unit test has been modified to deal with read and write locks but the code is pretty much as you had it.

So @aldebaran30701 your unit test was passing before my changes and they are still passing after my changes. I think your introduction of the queries list was to fix the bug with the graphs out of sync issue so are you able to keep this unit test and reproduce the bug to confirm this unit test can pick up the bug? Would it be a matter of running this unit test with the code in `master` and modify the unit test to undo some of the changes is PR #602?

Some notes:
* I've used `assertEquals()` instead
* I've added the NGTest setup methods

### Benefits

Graph bugs are something I never want to see introduced so spending the time to detect it in a unit test that may creep up in the future is going to be a well worth investment. Unit tests have proven really useful during v2 testing so I want to help try to catch this bug.

### DIFF
--- a/CoreLayersView/test/unit/src/au/gov/asd/tac/constellation/views/layers/utilities/LayersDualGraphSyncNGTest.java
+++ b/CoreLayersView/test/unit/src/au/gov/asd/tac/constellation/views/layers/utilities/LayersDualGraphSyncNGTest.java
@@ -18,8 +18,9 @@ package au.gov.asd.tac.constellation.views.layers.utilities;
 import au.gov.asd.tac.constellation.graph.Graph;
 import au.gov.asd.tac.constellation.graph.GraphWriteMethods;
 import au.gov.asd.tac.constellation.graph.LayersConcept;
-import au.gov.asd.tac.constellation.graph.StoreGraph;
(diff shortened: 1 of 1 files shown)


arcturus2:
> This looks to be causing the test to fail without the fix for layers included

Perfect!

>I'll be taking notes from this test to use in other test cases I create. Appreciate the time you took to look at the PR and help out developing the test

Anytime @aldebaran30701. Generally a `StoreGraph` is what we normally use because it makes it easier to read and write the unit test (no locking to deal with) but in this case we do want the `DualGraph`.

PR #627, opened 2020-07-06
Fix for schema view filter text turning invisible

formalhaut69:
Added to the css file to fix the colour unfocussing issue.

### Description of the Change
When the filter text box is unfocused the text colour would change to black making it nearly invisible. I added two css rules that always keep the text colour white. The !important tag was added to force the rules to override any other css rules in other files. I believe that the css rule that was causing this issue lies somewhere in an inbuilt css file that goes along with the Java textfield file.


arcturus2:
@formalhaut69 I had a test of this and for me the text is still coming up black and hard to see.

Auriga2:
Same here, I cannot see it working. But when I tried to debug and add a style in one tab, I could get it working on other tabs! Weird! The second time it didn't work either.

Issue #231, opened 2020-02-05
Random code cleanup tasks


arcturus2:
Here is a list of code quality improvements/checks that I had in my todo list. I've purposely kept them in one issue because we have a lot of issues already. Feel free to add to this or provide feedback.

* [ ] Verify that calls to `addListener` have a corresponding `removeListener`
* [x] Replace `Exceptions.printStackTrace` traces with `java.util.logger`
* [ ] Replace numbers with constants that call `Insets` method and perhaps make them available in a utility class so others can 
* [X] Review methods and arguement naming conventions in `StringUtilities` - having `String string` may be undesirable

* [X] Make a utility class to display `DialogDisplayer` errors instead of doing this for example:
```
final NotifyDescriptor nd = new NotifyDescriptor.Message("The provided URL is invalid.", NotifyDescriptor.ERROR_MESSAGE);
DialogDisplayer.getDefault().notify(nd);
```
* [X] Use `StringUtils.isBlank()` instead of the following:
```
!= null && !.isEmpty() everywhere.
isEmptyOrNull()
```
* [X] Search replace "," with SeperatorConstants.COMMA etc.
~~Annotate @nothreadsafe methods~~ - Not currently looking to pursue this throughout the whole codebase - new PRs can try to stick to this style if it helps dev.
~~Integrate JSR-305~~
~~* use @notnull etc ~~
 ~~SonarQube suggestions~~ - Removed from list as this is covered in other Sonar related tickets.

aldebaran30701:
>Verify that calls to addListener have a corresponding removeListener

I'm interested in the benefit of this. I had a quick google as I was curious as to java best practices, and found the following link which leads me to believe it isn't a garbage collection related code cleanup. 
https://coderanch.com/t/344003/java/call-remove-Listeners

Is it for another reason?

arcturus2:
> > Verify that calls to addListener have a corresponding removeListener
> 
> I'm interested in the benefit of this. I had a quick google as I was curious as to java best practices, and found the following link which leads me to believe it isn't a garbage collection related code cleanup.
> https://coderanch.com/t/344003/java/call-remove-Listeners
> 
> Is it for another reason?

[...]

PR #636, opened 2020-07-07
Code cleanup - replace null and empty checks with StringUtils

aldebaran30701:
Replace any check of the following
String != null && ! String.isEmpty()  ->  StringUtils.isNoneBlank(String)
String == null || String.isEmpty()  ->  StringUtils.isBlank(String)

Add utilities class for notifyDisplayer

This is code clean up as a part of #231


aldebaran30701:
> Replace `StringUtils.isNoneBlank` with `StringUtils.isNotBlank` (which I think is what you were after). Looks fine otherwise

Good catch! Lost in the documentation...

I've replaced those instances and it should be good to go now.

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/hierarchical/ArrangeInHierarchyAction.java
```diff
@@ -100,8 +101,7 @@ public void actionPerformed(final ActionEvent e) {
             }
 
             if (nsState == null) {
-                final NotifyDescriptor nd = new NotifyDescriptor.Message("There must be a named selection to specify the tree roots", NotifyDescriptor.WARNING_MESSAGE);
-                DialogDisplayer.getDefault().notify(nd);
+                NotifyDisplayer.display("There must be a named selection to specify the tree roots", NotifyDescriptor.WARNING_MESSAGE);
```
arcturus2, line 104: :heart:

CoreGraphNode/nbproject/project.xml
```diff
@@ -6,6 +6,14 @@
             <code-name-base>au.gov.asd.tac.constellation.graph.node</code-name-base>
             <suite-component/>
             <module-dependencies>
+                <dependency>
```
arcturus2, line 9: FYI an aspiration idea behind some of our modules like `CoreGraphNode` or `CorePluginFramework` is that it is portable. Adding more dependencies needs to be considered carefully. In this case I don't have an issue as it is 3rd party jars. `CoreGraphNode` is also heavily linked to NetBeans now anyway...

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/NotifyDisplayer.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Utility display method to show a dialog to the user.
+     * NotifyDescriptor.ERROR_MESSAGE will be used for errors
+     * NotifyDescriptor.INFORMATION_MESSAGE will be used for information
+     * NotifyDescriptor.WARNING_MESSAGE will be used for warnings
+     *
+     * @param message the String message to display on the prompt
+     * @param descriptorType the int value representative of the message type
+     */
+    public static void display(final String message, final int descriptorType) {
```
arcturus2, line 36: This is great. Can you please mention this in `CHANGELOG.md` as a way to advertise this? Thanks.

CoreGraphNode/nbproject/project.xml
aldebaran30701: I can't think of why it would have made it's way into the PR - Maybe when I added an import and just updated a dependency.

PR #637, opened 2020-07-07
Fix to the plugin environment

arcturus2:
### Description of the Change

* Returning null caused problems with context menu plugins not running
* Going to log this as FINE in case we want to know what's happening


aldebaran30701:
Was this the final fix added to the QCV PR where you close the graph and it null pointer errors? If it was, I think it's worth testing to see if it still exists and subsequently raise a ticket for it.

arcturus2:
> 
> 
> Was this the final fix added to the QCV PR where you close the graph and it null pointer errors? If it was, I think it's worth testing to see if it still exists and subsequently raise a ticket for it.

Nah. The null pointer when closing a graph with the QC view open had to do with how the QC Save State plugin was called. By not using the `getActiveGraph()` it fixed that problem.

PR #639, opened 2020-07-08
:bug: Synchronized renderTextAsLigatures method.

Nova-2119:
When a view is opened in a COnstellation instance with multiple graphs
multiple threads are trying to use the method at the same time. As the
drawing variable used comes from the static implementation of a class
they are using the same drawing causing issues.


arcturus2:
Thank you so much for fixing this bug @Nova-2119 ❤️

Issue #642, opened 2020-07-08
Remove Unused Attributes Plugin causes draw mode to error


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Running the Remove Unused Attributes Plugin on an empty graph, and then using draw mode to add a vertex causes an arrayindexoutofbounds error.
The error relates to an attribute not existing on a graph.
```
java.lang.ArrayIndexOutOfBoundsException: Index -1107 out of bounds for length 64
	at au.gov.asd.tac.constellation.graph.StoreGraph.setFloatValue(StoreGraph.java:2397)
	at au.gov.asd.tac.constellation.graph.interaction.plugins.draw.CreateVertexPlugin.edit(CreateVertexPlugin.java:89)
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:207)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:115)
Caused: java.lang.RuntimeException
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:135)
[catch] at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.lambda$executePluginLater$0(DefaultPluginEnvironment.java:122)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to Reproduce

1. New graph
2. Run Remove Unused Attributes Plugin 
3. Add element using draw mode

**Expected behaviour:** No error and an element appears

**Actual behaviour:** Error occurs 

[...]

PR #650, opened 2020-07-09
Resolved the error that the remove Unused Attributes Plugin caused

formalhaut69:
### Description of the Change
Added a check to stop the plugin from running if no nodes are on the active graph 

### Applicable Issues

https://github.com/constellation-app/constellation/issues/642


arcturus2:
Legend thanks for this @formalhaut69. Its a neat solution.

PR #652, opened 2020-07-09
:bug: Changed Arrange by Grid to use V1 logic.

Nova-2119:
I have changed the ArrangeInGridAction to call the GRID_COMPOSITE
plugin instead of the generic grid plugin. This changes the behaviour
so that the ArrangeInGrid action exhibitis the same behaviour as in
constellation V1. That is each subgraph gets its own grid.


arcturus2:
@serpens24 so in v1 singletons appear in a separate grid but in v2 its one big grid. We've opted to keen the existing functionality. Cool?

arcturus2:
Thanks @Nova-2119

PR #658, opened 2020-07-10
code cleanup-3

aldebaran30701:
### Description of the Change
This code cleanup PR makes use of the CollectionUtils methods to test for `col != null && !col.isEmpty()` and alike.

I've went through and removed:
* Unused imports
* unassigned variables

I replaced any case of string concatenation within a string builder to just use two `.append()` calls.

I've went through the whole Core Algorithm Plugins module and added finals to any that were missing.

### Alternate Designs
N/A

Improves code quality

### Benefits
Improves code quality

### Possible Drawbacks
Slight decrease in performance in exchange for readability (collection utils)

### Verification Process
Full clean and build, and static review of changes.

### Applicable Issues
#231


CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/hierarchical/HierarchicalState.java
```diff
@@ -22,19 +22,19 @@
...
+    public FastNewman.Group[] groups;
+    public boolean excludeSingleVertices;
+    public boolean excludedElementsDimmed;
+    public int[] clusterNumbers;
+    public int[] clusterSeenBefore;
+    public int redrawCount = 0;
+    public transient long modificationCounter = -1;
+    public transient long strucModificationCount = -1;
+    public boolean interactive = true;
+    public boolean colored = true;
```
antares1470, line 37: Is there a reason why these need to be made public? I would make these private and increase the accessibility as needed

aldebaran30701: Just an oversight on my end, thinking if private would not work without getters and setters, then I would just use public. Makes sense to have them protected. 👍

CoreGraphNode/nbproject/genfiles.properties
```diff
@@ -1,8 +1,8 @@
-build.xml.data.CRC32=ba351fce
+build.xml.data.CRC32=a437b10f
```
formalhaut69, line 1: Why is this committed?

CoreGraphNode/nbproject/genfiles.properties
```diff
@@ -1,8 +1,8 @@
...
 build.xml.script.CRC32=ab19df9f
 build.xml.stylesheet.CRC32=15ca8a54@2.79
 # This file is used by a NetBeans-based IDE to track changes in generated files such as build-impl.xml.
 # Do not edit this file. You may delete it but then the IDE will never regenerate such files for you.
-nbproject/build-impl.xml.data.CRC32=ba351fce
+nbproject/build-impl.xml.data.CRC32=a437b10f
```
formalhaut69, line 6: Same here. This shouldn't be committed.

CoreGraphNode/nbproject/genfiles.properties
formalhaut69, line 1: This file change shouldn't be committed

aldebaran30701: Fixed

arcturus2: Hey @formalhaut69 and @aldebaran30701 if the genfiles have changed then they should be committed. Normally genfiles change when you add new dependencies between modules.

PR #662, opened 2020-07-10
fix help page links

aldebaran30701:
### Description of the Change

Help pages over time have linked to the wrong module paths
This Pr fixes those so the help files link back up.
*bubble tree action - Does not have a proper documentation page, just a page with the title `bubble tree action`
*hierarchy action

Added a constant to hold the help location so it is easily seen within the classes in the need for future changes


CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/hierarchical/ArrangeInHierarchyAction.java
```diff
@@ -55,6 +55,7 @@
 public class ArrangeInHierarchyAction extends AbstractAction {
 
     private final GraphNode context;
+    private static final String HELP_LOCATION = "au.gov.asd.tac.constellation.plugins.arrangements.hierarchical.HierarchicalAction";
```
arcturus2, line 58: Perfect @aldebaran30701. Having "magic strings" taken out like this makes it way easier to know they exist so when we refactor code there is less chance they will be missed. Thank you!

Issue #668, opened 2020-07-13
DAV Split Nodes Plugin Null Pointer Exception


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Testing of 2.0 rc1 revealed that selecting all plugins from utility, all plugins from clean, and all plugins from hop and running them on a new graph with 3 drawn nodes will trigger a null pointer.

```
java.lang.NullPointerException
	at java.base/java.lang.String.contains(Unknown Source)
	at au.gov.asd.tac.constellation.views.dataaccess.plugins.clean.SplitNodesPlugin.query(SplitNodesPlugin.java:192)
[catch] at au.gov.asd.tac.constellation.views.dataaccess.templates.RecordStoreQueryPlugin.lambda$query$0(RecordStoreQueryPlugin.java:139)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### Steps to Reproduce
an exact reproducer has not been found, but none-the-less it will need to be fixed as it may occur elsewhere
1. DAV Open
2. new graph open
3. Draw mode and draw 3 vertices approx.
~4. Select all relevant plugins, no extra setup necessary~
4. Select split nodes based on identifier
5. hit 'run'

**Expected behaviour:** all plugins run and trigger warnings for unfilled parameters

**Actual behaviour:** null pointer on Split nodes, ~as well as triggering warning for unfilled parameters~

**Reproduces how often:** ~90%~ 10% with only split nodes selected. Have triggered it to occur with only split nodes selected, but it is seemingly random when it occurs. A null check wold fix this without much overhead.

### Additional Information
N/A

PR #709, opened 2020-07-16
:bug: switch default value to empty string

aldebaran30701:
### Description of the Change
Switches default value from null to empty string. Calling .contains(null) was resulting in a NPE.

### Alternate Designs
using StringUtils.EMPTY instead of ""
Unsure what is preferred in this scenario.

Bugfix

### Applicable Issues

#668

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
@@ -91,7 +91,7 @@ public class SplitNodesPlugin extends RecordStoreQueryPlugin implements DataAcce
         final PluginParameter<StringParameterValue> split = StringParameterType.build(SPLIT_PARAMETER_ID);
         split.setName("Split Character(s)");
         split.setDescription("A new term will be extracted from the first instance of this character(s) in the Identifier");
-        split.setStringValue("");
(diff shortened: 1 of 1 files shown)


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -91,7 +91,7 @@ public PluginParameters createParameters() {
         final PluginParameter<StringParameterValue> split = StringParameterType.build(SPLIT_PARAMETER_ID);
         split.setName("Split Character(s)");
         split.setDescription("A new term will be extracted from the first instance of this character(s) in the Identifier");
-        split.setStringValue(null);
+        split.setStringValue("");
```
arcturus2, line 94: My preference would be to do a null check further on rather than set this to an empty string. I say this because if via the ReST API we set it to null then we easily break the plugin. What do you think @aldebaran30701 ?

aldebaran30701: Good point, I'm adding a check to it instead. I'll test it and push the change in a moment. 👍

arcturus2: Thank you @aldebaran30701

Issue #718, opened 2020-07-16
Analytic view loses plugin results on view re-open


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Opening the Analytic view, running a question and reviewing the results happens all ok. The issue is when the user switches to another view like Attribute Editor and then back to the Analytic View. Doing so will require the user to re-run the question to view results. This only occurs when the user opens the other view over the top of Analytic View. It makes me think it is an issue regarding `componentShown(...)` and an unnecessary refresh.

### Steps to Reproduce

1. Open graph - draw mode w/a few nodes
2. Open Analytic View -> run betweenness centrality. or others.
3. Open Attribute Editor over the top
4. Switch back to Analytic View

**Expected behaviour:** State is kept

**Actual behaviour:** View is refreshed.

**Reproduces how often:** 100%

cygnus-x-1:
While individual analytic results are saved to the graph (degree centrality, cosine similarity, etc), the aggregated results displayed in the Analytic View are not. The intention was to save the aggregated result set for each tab to the Analytic View state... but first we need to give the Analytic View a state.

PR #719, opened 2020-07-17
:bug: Analytic View results refresh fix

aldebaran30701:
### Description of the Change
This change tracks the last current graph loaded with the view, and checks before reloading the view if a complete view reset is necessary.

### Alternate Designs
N/A

Bug fix

### Benefits
Keeps the results pane open when using the same graph and switching views

### Possible Drawbacks
N/A

### Verification Process
Test opening with a graph, test opening without a graph, test switching of graph when view is minimized, test results pane consistency with same graphs.

### Applicable Issues
#718

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
@@ -19,6 +19,7 @@ import au.gov.asd.tac.constellation.graph.Graph;
 import au.gov.asd.tac.constellation.graph.GraphElementType;
 import au.gov.asd.tac.constellation.graph.GraphWriteMethods;
 import au.gov.asd.tac.constellation.graph.ReadableGraph;
+import au.gov.asd.tac.constellation.graph.manager.GraphManager;
 import au.gov.asd.tac.constellation.graph.schema.attribute.SchemaAttribute;
 import au.gov.asd.tac.constellation.graph.schema.visual.concept.VisualConcept;
 import au.gov.asd.tac.constellation.plugins.PluginException;
@@ -74,6 +75,7 @@ public final class AnalyticViewTopComponent extends JavaFxTopComponent<AnalyticV
     private final AnalyticViewPane analyticViewPane;
     private final AnalyticController analyticController;
     private boolean suppressed = false;
+    private Graph currentGraph = null;
 
     public AnalyticViewTopComponent() {
         super();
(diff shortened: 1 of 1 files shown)


CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewTopComponent.java
```diff
@@ -74,6 +75,7 @@
     private final AnalyticViewPane analyticViewPane;
     private final AnalyticController analyticController;
     private boolean suppressed = false;
+    private Graph currentGraph = null;
```
arcturus2, line 78: There is potential for a memory leak to appear if the `currentGraph` is not set to `null` when the graph closes. I am speculating but I will do some testing to confirm @aldebaran30701.

aldebaran30701: How does it sound if I store the current graph's ID instead?

arcturus2: That sounds like a better idea though we haven't depended on the graph ID for much so I am not sure if it is the best approach. If it works then that does sound like a really good way forward and perhaps we can adopt it across more of the framework.

I remember seeing the graph ID change after plugins run (but I need to double check my memory). If that's the case (which I don't think so) would be a problem.

aldebaran30701: Hi @cygnus-x-1,  Would you be able to confirm for us if the graph id changes between plugin runs? Or subsequently offer a different way of tracking the current graph without memory leaks.

antares1470: I don't know whether or not it helps avoid memory leaks but having played with the rest api previously, I believe the graph Id remains the same between plugin runs (even between Constellation loads)

arcturus2: >...(even between Constellation loads)

Not sure on this bit @antares1470. The graph UUID is not stored in the star file so I believe a random one is issued on load.

cygnus-x-1: I believe a graph id should be reliable as long as that graph is in memory.

antares1470: Right. Like I said, it's been a while so I was only going off the top of my head

aldebaran30701: > If the graph is closed and opened again, it'll have a new id

I'd think that would cross over into a future state implementation. This (keeping the id) would work for the sake of not clearing the results table.

Issue #708, opened 2020-07-16
Hierarchical clustering Array index out of bounds


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Tested using V2.0 rc2
using the view showed this issue when following the reproducer loosely.
```
java.lang.ArrayIndexOutOfBoundsException: Index 64 out of bounds for length 64
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.hierarchical.HierarchicalControllerTopComponent$Update.edit(HierarchicalControllerTopComponent.java:726)
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:207)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:115)
Caused: java.lang.RuntimeException
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:135)
[catch] at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.lambda$executePluginLater$0(DefaultPluginEnvironment.java:129)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### Steps to Reproduce

1. Open new graph
2. Hierarchical clustering view opened
3. Draw mode, approx 7 vertices with 3 transactions 
4 cluster
5. enter interactive mode
6. run sphere graph builder
7. disable interactive

**Expected behaviour:** no error, disable colors/view

**Actual behaviour:** graph freezes up, and exception above

**Reproduces how often:** 100%

### Additional Information
Tested on .exe V2.0 rc2
[...]

PR #732, opened 2020-07-20
hierarchical clustering arrayindexoutofbounds fix

aldebaran30701:
### Description of the Change
Use the lower value out of vertex count, and group count. This will allow the clustering to run normally when the groups and vertexes are the same, but when the cluster button has not re-run, and made more groups within the state it will stop iteration over groups which do not exist.

### Alternate Designs
~check if the array position exists before every iteration~

bugfix

### Benefits
bugfix

### Possible Drawbacks
None realised

### Verification Process
Testing replication method and normal operations

### Applicable Issues
#708


CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/hierarchical/HierarchicalControllerTopComponent.java
```diff
@@ -719,8 +719,10 @@ public void edit(final GraphWriteMethods graph, final PluginInteraction interact
             final int transactionVisibilityAttribute = VisualConcept.TransactionAttribute.VISIBILITY.ensure(graph);
 
             int nextCluster = 0;
+            // Use the smaller of ints as to avoid going out of bounds.
+            // usually caused when cluster has not run with new elements added.
+            final int vertexCount = Math.min(graph.getVertexCount(), state.getCurrentNumOfClusters());
```
cygnus-x-1, line 724: The `vertexCount` is being used to retrieve vertex ids, so setting it to the number of clusters feels wrong, and like it could be error prone. What was you reasoning behind this?

aldebaran30701: @cygnus-x-1  So it grabs the group based on the current vertex position it iterates on. My thoughts based on this is that you never want it to grab a group by an iterator that is larger than the total number of groups. This happens when you cluster it once, then add elements to the graph. When you switch off interactive mode it iterates all nodes which were effected in the last cluster so it can revert the changes. Just writing this now it seems like there may be a possibility that it iterates the vertices, but does so under the assumption that the first x elements are the same as the ones which were clustered. I did not notice any issues with testing under normal circumstances. 
Making it the minimum of both makes it avoid checking the groups that does not exist, but still effectively returns the state to normal in the vertices that were effected under "interactive mode" 
Does this sound logical, or do you suggest revisiting it. I didn't want to just re-cluster the graph in case it messed with the behaviour a user would want to target.

cygnus-x-1: If I understand correctly, the underlying issue is that a graph could be updated (nodes added, removed, etc) which renders the clustering result invalid. If this is the case I feel like the clustering result should either be removed (forcing the user to re-cluster) or automatically updated (could have an impact on performance). @arcturus2, @sirius-1 what do you think?

aldebaran30701: You do understand correctly. Keen to hear their thoughts. 👍

arcturus2: If the cluster is considered invalid then I would prefer we force the user to re-cluster by updating the GUI to look like it wasn't clustered. Perhaps a label somewhere in red which says something like "You need to re-cluster because the graph has changed" would be a nice UX touch.

@cygnus-x-1 given our view of Constellation is to not do "black magic" (i.e. doing things behind the scenes without the user knowing), automatically updating the cluster I think would be going down this path. Would you agree?
Also if automatically updating clusters as the graph changes could be a performance hit then that's a turn off for me personally.

cygnus-x-1: While I agree with the "no black magic" principle, in this case the user had asked for the clustering to be done so i wouldn't have an issue with the view keeping that in sync with the graph.

aldebaran30701:
This PR now works as follows:
The use clusters a graph, then adds extra elements to it. 
This will not disable the interactivity of the view, and only allow the user to re-cluster. 
After re-clustering the user will be able to use the view as normal with full capability.

arcturus2:
Looks good.

image

PR #737, opened 2020-07-23
Upgrade Schema Vertex Types - Fixes #730

### LINKED ISSUE #730: Duplicate Types with v1 and v2
### Prerequisites

### Description

New nodes added in v2 from an existing graph from v1 will produce duplicates. After further debugging the issue seems to be related to the Types being different under the hood. We changed the detection and validation regex which is part of the type. The solution would be to upgrade the types using the Schema upgrade framework when opening old graphs.

### Steps to Reproduce

1. Open Constellation v1
2. Go into draw mode and add a node
3. Open the attribute editor and change the Type to "Ipv4 Address"
4. Press F5
5. Save and close
6. Open Constellation v2
7. Go into draw mode and add a node
8. Open the attribute editor and change the Type to "Ipv4 Address"
9. Press F5. You should have Vertex #0<Ipv4 Address>and Vertex #1<Ipv4 Address>. So far so good.
10. Select Vertex1 and in the attribute editor select Edit next to Identifier
11. Change it to Vertex #0 and click OK
12. Press F5

duplicate nodes

**Expected behaviour:** The nodes should merge into 1

**Actual behaviour:** There are 2 nodes on the graph!

### Additional Information

This should be fixed for all types.

arcturus2:
### Description of the Change

Adding certain nodes to a v1 graph opened in v2 will introduce duplicate types. This is because the Type have changed (which is the validation and detection regex).

We can use the Schema upgrade framework to detect old graphs and then upgrade them. Once this change goes through, the Analytic Schema version will be 4.

### Benefits

Supporting backwords compatibilty is something we always want to do in Constellation to make it seamless for users to continue working with their graphs and not worry about version upgrades breaking their graphs.

### Verification Process

1. I created a graph in v1.3.1 using [this script](https://gist.github.com/arcturus2/5d211d9f993f240564011ee42353fcdb) run in the Scripting View. There are 27 nodes.
2. Opened the graph in v2 and ran the script again. This showed 13 new nodes. These needed to be upgraded.

After adding `AnalyticSchemaV4UpdateProvider` I re-ran step 2 and observed that there was 27 nodes.

I decided to use the Scripting View so that I didn't need to write a plugin and have to checkout Constellation v1 to implement it there also. I guess there was only 27 nodes so you could create them by hand...but where is the fun in that?

### Applicable Issues

#730

### DIFF
--- a/CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/compatibility/AnalyticSchemaV4UpdateProvider.java
+++ b/CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/compatibility/AnalyticSchemaV4UpdateProvider.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2010-2020 Australian Signals Directorate
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.graph.schema.analytic.compatibility;
+
+import au.gov.asd.tac.constellation.graph.StoreGraph;
+import au.gov.asd.tac.constellation.graph.schema.SchemaFactory;
+import au.gov.asd.tac.constellation.graph.schema.SchemaFactoryUtilities;
+import au.gov.asd.tac.constellation.graph.schema.analytic.AnalyticSchemaFactory;
+import au.gov.asd.tac.constellation.graph.schema.analytic.concept.AnalyticConcept;
+import au.gov.asd.tac.constellation.graph.schema.type.SchemaVertexType;
+import au.gov.asd.tac.constellation.graph.versioning.SchemaUpdateProvider;
+import au.gov.asd.tac.constellation.graph.versioning.UpdateProvider;
+import java.util.HashMap;
+import java.util.Map;
+import org.openide.util.lookup.ServiceProvider;
+
+/**
+ * Upgrade Vertex Types that have changed.
+ * <p>
+ * The detection regex and validation regex was reviewed and improved.
+ *
+ * @author arcturus
+ */
+@ServiceProvider(service = UpdateProvider.class)
+public class AnalyticSchemaV4UpdateProvider extends SchemaUpdateProvider {
+
+    public static final int SCHEMA_VERSION_THIS_UPDATE = 4;
+
+    @Override
+    protected SchemaFactory getSchema() {
+        return SchemaFactoryUtilities.getSchemaFactory(AnalyticSchemaFactory.ANALYTIC_SCHEMA_ID);
+    }
+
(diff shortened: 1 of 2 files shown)


CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/compatibility/AnalyticSchemaV4UpdateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+                put(AnalyticConcept.VertexType.GEOHASH.getName(), AnalyticConcept.VertexType.GEOHASH);
+                put(AnalyticConcept.VertexType.MGRS.getName(), AnalyticConcept.VertexType.MGRS);
+                put(AnalyticConcept.VertexType.MD5.getName(), AnalyticConcept.VertexType.MD5);
+                put(AnalyticConcept.VertexType.IPV4.getName(), AnalyticConcept.VertexType.IPV4);
+                put(AnalyticConcept.VertexType.IPV6.getName(), AnalyticConcept.VertexType.IPV6);
+                put(AnalyticConcept.VertexType.EMAIL_ADDRESS.getName(), AnalyticConcept.VertexType.EMAIL_ADDRESS);
+                put(AnalyticConcept.VertexType.HOST_NAME.getName(), AnalyticConcept.VertexType.HOST_NAME);
+                put(AnalyticConcept.VertexType.URL.getName(), AnalyticConcept.VertexType.URL);
+                put(AnalyticConcept.VertexType.TELEPHONE_IDENTIFIER.getName(), AnalyticConcept.VertexType.TELEPHONE_IDENTIFIER);
+            }
```
antares1470, line 76: Sonar is identifying using double brace initialisation as a bug. Consequently, it would be good to remove the need to use it

arcturus2: Yep, easily fixed.

CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/compatibility/AnalyticSchemaV4UpdateProvider.java
```diff
@@ -0,0 +1,85 @@
...
+        typesToUpgrade.put(AnalyticConcept.VertexType.EMAIL_ADDRESS.getName(), AnalyticConcept.VertexType.EMAIL_ADDRESS);
+        typesToUpgrade.put(AnalyticConcept.VertexType.HOST_NAME.getName(), AnalyticConcept.VertexType.HOST_NAME);
+        typesToUpgrade.put(AnalyticConcept.VertexType.URL.getName(), AnalyticConcept.VertexType.URL);
+        typesToUpgrade.put(AnalyticConcept.VertexType.TELEPHONE_IDENTIFIER.getName(), AnalyticConcept.VertexType.TELEPHONE_IDENTIFIER);
+
+        for (int vertex = 0; vertex < graph.getVertexCount(); vertex++) {
+            final int vertexId = graph.getVertex(vertex);
+            final SchemaVertexType oldType = graph.getObjectValue(typeAttribute, vertexId);
+
+            if (typesToUpgrade.containsKey(oldType.getName())) {
```
cygnus-x-1, line 80: This will probably work just fine, but could be improved by comparing more of the type object than just the name.

arcturus2: Perhaps @cygnus-x-1 though remember that this is upgrading from v3 so we should be safe to assume what the type is at this point.

PR #738, opened 2020-07-23
Update documentation

arcturus2:
### Description of the Change

* Update with steps on how to build a standalone version
* Improve formatting
* Remove reference to not supporting MacOSX (because now we do!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,10 @@ To package Constellation in a zip bundle do the following:
 * In NetBeans, expand `Constellation` > `Important Files`
 * Right click on `Build Script` and run the `build-zip`
 * Navigate to the `dist` folder to get `constellation.zip` file
-* If you want to use a specific JRE then copy this to the same level as the 
-`bin` folder and call it `jre`. This is the default folder name Constellation 
-expects but you can change this from `etc\constellation.conf`.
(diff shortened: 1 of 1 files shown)


Auriga2:
Tried the below from constellation-app/constellation  (constellation-app/constellation-applications works)
> Right click on `Build Script` and run the `build-zip`
> * Navigate to the `dist` folder to get `constellation.zip` file

But still get the error when running the exe
image

If I click yes:
image

Should I follow this here? 
>* If you want to use a specific JRE then copy this to the same level as the `bin` folder and call it `jre`.

arcturus2:
@Auriga2 

If you try to run the .exe it will try to look for the local `'jre' folder but if it can't find it then it will look at the system wide version of Java. This is expected.


> Should I follow this here?
> If you want to use a specific JRE then copy this to the same level as the bin folder and call it jre.

Yep, follow this and you should be good to go.

Auriga2:
Had trouble figuring out how to find what to copy as 'jre'. I figured it out by comparing the folder content with the working one!
I guess specifying `zulu11.37.19-ca-fx-jdk11.0.6-win_x64 folder ` at least as an example, would help someone.

arcturus2:
Updated @Auriga2, thank you for the feedback.

PR #740, opened 2020-07-24
 Fix for forward slash being allowed in graphs name, causes errors if used

formalhaut69:
### Description of the Change
Added a regex check to restrict file names and removed functionality for making a new folder.

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
@@ -64,6 +64,7 @@ import au.gov.asd.tac.constellation.graph.node.GraphNode;
 import au.gov.asd.tac.constellation.graph.schema.SchemaFactoryUtilities;
 import au.gov.asd.tac.constellation.preferences.ApplicationPreferenceKeys;
 import au.gov.asd.tac.constellation.utilities.BrandingUtilities;
+import au.gov.asd.tac.constellation.utilities.gui.NotifyDisplayer;
 import java.awt.event.ActionEvent;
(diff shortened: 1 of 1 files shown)


arcturus2:
@formalhaut69 this is more an FYI but remember that the entire source code of NetBeans which we are using is available on GitHub. The latest version of this file which we copied in our Save As feature is available https://github.com/apache/netbeans/blob/master/platform/openide.loaders/src/org/openide/actions/SaveAsAction.java. I had a quick compare of the two files and I didn't see any bug or performance improvements we could include in our copy. Again this is more an FYI.

formalhaut69:
> 
> 
> @formalhaut69 this is more an FYI but remember that the entire source code of NetBeans which we are using is available on GitHub. The latest version of this file which we copied in our Save As feature is available https://github.com/apache/netbeans/blob/master/platform/openide.loaders/src/org/openide/actions/SaveAsAction.java. I had a quick compare of the two files and I didn't see any bug or performance improvements we could include in our copy. Again this is more an FYI.

Cool, thanks @arcturus2, I'll keep that in mind if I have to touch the saveAsAction again at some point.

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/io/SaveAsAction.java
```diff
@@ -293,6 +273,13 @@ private File getNewFileName() {
             if (newFile == null) {
                 break;
             }
+            while (!newFile.getName().matches("^[\\w-_.]*$") || !newFile.getParentFile().exists()) {
+                NotifyDisplayer.display("File names can only include alphanumeric characters, - or .", NotifyDescriptor.WARNING_MESSAGE);
```
arcturus2, line 277: Should this mentioned the "_" (underscore) as well?

Issue #711, opened 2020-07-16
The pop up message when starting Rest Server is hiding behind the graphs.


Auriga2:
### Description

When the `Tools` -> `Start Rest Server` is clicked, the below message pops up, but it's only visible if there are no graphs opened. When any graph is opened, only the bottom part is visible, although you can see the status by clicking on the orange triangle, which displays status of all the services.

![image](https://user-images.githubusercontent.com/64826928/87628638-90555f00-c770-11ea-9538-c87b4fc3b08b.png)

And maybe it should not allow new services to be started on the same port?
![image](https://user-images.githubusercontent.com/64826928/87628699-afec8780-c770-11ea-99d7-f2dfeb69a823.png)

### Steps to Reproduce

1. Open a graph, then click `Tools` -> `Start Rest Server`.
2. Check the notification at the bottom right corner. 

**Expected behaviour:** The pop up message should appear at the front

**Actual behaviour:** The pop up message should appear behind the graph

**Reproduces how often:** 100%

### Additional Information

Same behaviour in v2.0.0-beta5

Auriga2:
1. Recreatable in the v2 exe, only if the focus was on the graph (it works if the focus was on a view), but not in debug mode. 
2. It's not really starting multiple web servers, but displaying the same message pretends so. As  a fix we can display "Web server already started" instead of  "Web server started" to reduce confusion.

arcturus2:
> It's not really starting multiple web servers, but displaying the same message pretends so. As a fix we can display "Web server already started" instead of "Web server started" to reduce confusion.

Yep that would be a good idea.

Auriga2:
1. Clicking on the graph also brings the notification to the front. And this is the same behaviour in beta 5. It's hard to fix because we can't recreate in debug mode. @arcturus2 Should we put more effort to fix this, given it's a minor issue? If so who'd have the best advice?
2. Easy fix.

arcturus2:
1. @Auriga2 I support the 90/10 Principle. To fix this edge case I feel will take a long time and I do agree this is a low priority. Though not ideal from a user perspective; to fix this there is a chance you will need to get stuck into the NetBeans source code to fix this given the pop-ups are coming from NetBeans. I would suggest not spending the time to fix this; your time could be better spend helping with other bugs on the backlog.
2. Cool, if its an easy fix then that would be great to have.

Apologies for the delayed response.

aldebaran30701:
[...]

PR #744, opened 2020-07-27
Reword the pop up  notification when the REST server is already started

Auriga2:
### Description of the Change

The pop up message when starting the Rest Server saying `Web server started` is displayed for each time the `Tools` > `Start Rest Server` menu item is clicked. This gives a false impression of starting multiple servers.

### Alternate Designs
Disable menu item when the server is already running?

Less confusion for user

### Benefits
Better UX

### Possible Drawbacks

N/A

### Verification Process

1. Click `Tools` > `Start Rest Server` menu item. The pop up message should say `Web server started`
1. Click `Tools` > `Start Rest Server` menu item again. The pop up message should say `Web server already started`

### Applicable Issues

https://github.com/constellation-app/constellation/issues/711


Auriga2:
Ahh my bad! I tried to improve further by making the conditional check inline. It should be done before starting the web server!

PR #754, opened 2020-08-03
Feature/issue364 improve uncollide performance

Nova-2119:
# Description of the Change

### Benefits

Easier to make genuine use of large graphs.

### Possible Drawbacks

The max expansions parameter will have little meaning to analysts and could be misused.


Nova-2119:
Don't approve just yet, want to make a couple of minor improvements tomorrow morning.

Issue #378, opened 2020-03-27
Replace Quality Scores with Categories


cygnus-x-1:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Quality Control View is currently designed to run 'rules' across data in a graph in order to identify data quality issues, and to provide feedback on these issues in the form of a quality 'score' (0 to 100 - with 100 representing the worst quality) and a 'reason' for that score. Since these scores are essentially arbitrary (what does a score of 70 actually mean?), I propose we instead move to a system where we replace data quality scores with data quality categories, which I believe would be more interpretable. I imagine these categories might work much like a logging system, where there is a hierarchy of categories representing the severity or impact of the data quality issue (ie. FATAL, ERROR, WARNING, INFO). We could then assign each rule to a category - so the 'unknown type' rule might map to the 'INFO' category, whereas the 'blocked by compliance' rule would map to the 'FATAL' category. You might also add the ability to configure the rank of each category, so a user could specify which categories are important to them.

aldebaran30701:
I was thinking of implementing a menu which would allow the user to map any rule to a category like mentioned above. Possibly using a linking system like how the Map View uses any class registered with a provider annotation. It could be a dynamically loaded list where the drop down options allow mapping.

I was thinking of having them effect elements so that the "label" or "category" was added to it. instead of showing the quality in a score, it would append labels. ie. IdentifierInconsistentWithTypeRule + UnknownTypeRule would both appear.

This is just going by my understanding of the QCV, and assumed ability to have multiple rules broken or satisfied by a single element.

I'm thinking for ease of use, it might be better to have a set of categories predefined and not allow the user to change them - just map them differently.

I'd like to hear your input on this, @cygnus-x-1 @arcturus2 etc.

arcturus2:
> I was thinking of implementing a menu which would allow the user to map any rule to a category like mentioned above.

[...]

PR #760, opened 2020-08-04
:art: Implement quality categories

aldebaran30701:
### Description of the Change
This PR adds Quality categories to replace the visual representation of quality scores. 
It does not completely rid the code of "scores", as completely getting rid of the abstract
`getQuality()` method would cause other rules to break.

The QualityControlView table and the Data access view will now show categories of quality, as opposed to scores.

The categories map to int values which are representative of the old values. for example, the quality score previously used will map to a category. 
* MissingTypeRule RISK = 90 , maps to SEVERE.

DEFAULT = 1 - 29
INFO = 30 - 59
WARNING = 60 - 89
SEVERE = 90 - 94
FATAL = 95 +

An alert box now is implemented to map rules to each priority level. Allowing for manual overriding of specific rules or rules which are important to the user.
image

Highest priority rule in violation gets that label added.
image

Rule dialog shows in categories and also ranks highest priority category first.
image

Can map a rule to the default category, allowing that rule to be completely ignored.
image

Highest priority category in action gets added to the DAV button
image

### Alternate Designs
N/A

Feature allows user to easily read the quality of the data

### Benefits
Feature allows user to easily read the quality of the data

### Possible Drawbacks
None realised

### Verification Process
loading different data, testing sort order, testing any place score was used previously.
opening up dialog boxes, switching priorities.

### Applicable Issues
#378

### DIFF
--- a/CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewPane.java
+++ b/CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewPane.java
@@ -384,13 +384,13 @@ public final class QualityControlViewPane extends BorderPane {
     private static void showPriorityDialog() {
         final ScrollPane rulesScrollPane = new ScrollPane();
         rulesScrollPane.setPrefHeight(240);
-        rulesScrollPane.setPrefWidth(500);
+        rulesScrollPane.setPrefWidth(700);
         rulesScrollPane.setVbarPolicy(ScrollPane.ScrollBarPolicy.ALWAYS);
         rulesScrollPane.setHbarPolicy(ScrollPane.ScrollBarPolicy.NEVER);
 
         final GridPane buttonGrid = new GridPane();
         buttonGrid.prefWidthProperty().bind(rulesScrollPane.widthProperty());
-        buttonGrid.setPadding(Insets.EMPTY);
+        buttonGrid.setPadding(new Insets(5, 0, 5, 0));
 
         // Setting headings
         int rowCount = 0;
@@ -428,6 +428,28 @@ public final class QualityControlViewPane extends BorderPane {
             final RadioButton warningButton = new RadioButton();
             final RadioButton severeButton = new RadioButton();
             final RadioButton fatalButton = new RadioButton();
+            final Button resetButton = new Button("Reset");
+            resetButton.setOnAction(event -> {
+                switch (rule.getCategory(0)) {
+                    case DEFAULT:
+                        defaultButton.setSelected(true);
+                        break;
+                    case INFO:
+                        infoButton.setSelected(true);
+                        break;
+                    case WARNING:
+                        warningButton.setSelected(true);
+                        break;
+                    case SEVERE:
+                        severeButton.setSelected(true);
+                        break;
+                    case FATAL:
+                        fatalButton.setSelected(true);
+                        break;
+                    default:
+                        break;
+                }
+            });
 
             getPriorities().putIfAbsent(rule, rule.getCategory(0));
             // setting the selection based on the current priority
@@ -459,6 +481,7 @@ public final class QualityControlViewPane extends BorderPane {
             GridPane.setHalignment(warningButton, HPos.CENTER);
             GridPane.setHalignment(severeButton, HPos.CENTER);
             GridPane.setHalignment(fatalButton, HPos.CENTER);
+            GridPane.setHalignment(resetButton, HPos.CENTER);
 
             ruleName.setPadding(new Insets(0, 0, 5, 5));
             defaultButton.setPadding(new Insets(0, 10, 5, 10));
@@ -466,6 +489,7 @@ public final class QualityControlViewPane extends BorderPane {
             warningButton.setPadding(new Insets(0, 10, 5, 10));
             severeButton.setPadding(new Insets(0, 10, 5, 10));
             fatalButton.setPadding(new Insets(0, 10, 5, 10));
+            resetButton.setPadding(new Insets(5, 10, 5, 10));
 
             defaultButton.setUserData(QualityCategory.DEFAULT);
             infoButton.setUserData(QualityCategory.INFO);
@@ -487,6 +511,7 @@ public final class QualityControlViewPane extends BorderPane {
             buttonGrid.add(warningButton, 3, rowCount);
             buttonGrid.add(severeButton, 4, rowCount);
             buttonGrid.add(fatalButton, 5, rowCount);
+            buttonGrid.add(resetButton, 6, rowCount);
 
             rowCount++;
         }
@@ -496,7 +521,8 @@ public final class QualityControlViewPane extends BorderPane {
         final Alert alert = new Alert(Alert.AlertType.INFORMATION,
                 "Select Rule Priorities",
                 ButtonType.OK, ButtonType.CANCEL);
-        alert.setHeaderText("Priority defines category levels for each rule");
+        alert.setTitle("Select Rule Priorities");
+        alert.setHeaderText("Customise the priority of rules");
         alert.getDialogPane().setContent(rulesScrollPane);
         alert.setResizable(true);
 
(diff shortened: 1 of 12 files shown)


aldebaran30701:
This is ready for review, it appears that the 2 sonar warnings are false positives. Unless the issue is staring me in the face!

CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewPane.java
```diff
@@ -70,38 +79,44 @@
...
+    private final TableView<QualityControlEvent> qualityTable;
     private final FlowPane optionsPane;
 
-    public QualityControlViewPane(QualityControlViewTopComponent topComponent) {
-        this.parent = topComponent;
+    static {
+        for (final QualityControlRule rule : Lookup.getDefault().lookupAll(QualityControlRule.class)) {
+            rulePriorities.put(rule, rule.getCategory(0));
+        }
+    }
```
arcturus2, line 95: Running this in a static block like this means this bit of code must run when the class is loaded into memory (i.e when Constellation loads). This becomes problematic for a couple of reasons:
1) Constellation load times will take longer.
2) There is actually a runtime exception that Java throws when there are too many things run in a static block like this and a while back I moved them to be called lazily instead which you can do in `getPriorities()` with something like:

```java
private static Map<QualityControlRule, QualityCategory> rulePriorities = null;
...
...
public static Map<QualityControlRule, QualityCategory> getPriorities() {
    if(rulePriorities == null) {
        for (final QualityControlRule rule : Lookup.getDefault().lookupAll(QualityControlRule.class)) {
            rulePriorities.put(rule, rule.getCategory(0));
        }
    }
}

return rulePriorities;
```

FYI the call to `Lookup` isn't super efficient. For instance I have been able to dramatically shave seconds off the execution just by caching the object returned by `Lookup` where as it was being called many times (like in for loops in other parts of the app).

aldebaran30701: Hi @arcturus2 , Thanks for all the feedback! 👍 

>With the custom priorities radio option section, I don't believe this is being persisted into a Preferences object so it looks like changes will be lost after restarting Constellation.

You are correct, it doesn't currently.
I've been looking to store it into a preferences object but have hit a snag. I want to store a map into the preferences but have not found a method to set an object into the preferences. it only takes String or numeric values. 
From this I've pivoted to converting the map to a json String. I've looked through the utils and haven't found anything out the box. Do you have any suggestions re this or another possible solution?

>It didn't look like any of the "Rule" classes like UnknownTypeRule needed any changes; that's interesting and good actually.

Ideally this was what I was going for so I'm glad it all works without needing further extensions to those classes.


>Can you please write some unit tests to cover the logic? If you want a hand just let me know and happy to help writing uni tests.

>There should be a mention about this change to the users via the whatsnew.txt file informing them that scores are replaced by categories.

I can do these. 🍰 


>Running this in a static block like this means this bit of code must run when the class is loaded into memory (i.e when Constellation loads). 

Ohh no worries, I think that should be an easy fix.


>FYI the call to Lookup isn't super efficient. For instance I have been able to dramatically shave seconds off the execution just by caching the object returned by Lookup

I could store it into a class level variable, but when would i know to update it? only on instantiation of the view? 

I'll be a little while getting these changes in, and will re-request review once it's all wrapped up. 
Once again, cheers for taking such an in-depth look at this. 🙏

arcturus2: >...From this I've pivoted to converting the map to a json String. I've looked through the utils and haven't found anything out the box. Do you have any suggestions re this or another possible solution?

Yep that is how some other preferences work when you need to do more than a String. Have a look at your boostrap folder and then `build\testuserdir\config\Preferences\au\gov\asd\tac\constellation\plugins.properties`. Its a json string which `RecentParameterValues` loads and saves.

> I could store it into a class level variable, but when would i know to update it? only on instantiation of the view?
Yep that's how I would do it. You should be fine to put it in the class where `getPriorities` gets called.

aldebaran30701:
Hi @arcturus2 , 
Just a heads up this is what I've changed.

 *Implemented Preferences object for mapping of categories to rules
 *Added JSON Utilities for writing a Map to String and vice versa.
 *Added entry into `whatsnew.txt`
 *Lazy initialisation over rulePriorities Map
 *Lazy initialisation and caching of lookup object

Added unit tests for:
 *JSON Utilities
 *QualityControlEvent
 *Quality Control Rule

Should be all of the suggestions 👍

arcturus2:
Hi @aldebaran30701 so I got an NPE at line `QualityControlEvent.java:87`. I tried to continued testing by putting in 

```
category = QualityControlViewPane.getPriorities().get(rule) != null ?
                            QualityControlViewPane.getPriorities().get(rule):
                            QualityCategory.DEFAULT;
```

Next, for some reason I am not seeing your GUI changes like the category appearing either. I get this:

Screenshot from 2020-08-07 09-47-47

aldebaran30701:
>Hi @aldebaran30701 so I got an NPE at line QualityControlEvent.java:87.

Now that I read this, I'm thinking I missed the test case where the user does not have any preferences saved in memory. I'll verify all my logic and test this case to check if that's what's causing it. If that is the case, It would be why no GUI is updating. It defaults to Default category which is shown.

arcturus2:
I see @aldebaran30701. Again this is great and I want to merge this is in as soon as possible but I would like one more thing done. If the requirement is that a user has to set a preference for each rule then that will be a problem if they don't.

So what we generally do is have a default implementation. This could be made via a `ServiceProvider` annotation (meaning you use Lookup) to provide a default implementation for the preference. 

This would also allow other module suites to override the default preference if they don't agree with what the default implementation provided. And the user could also override it too. The user overriding vs the ServiceProvider is a new concept though.

@cygnus-x-1 what is your thoughts on all of this and this PR in general given you wanted the user to be able to customise the severity as well.

aldebaran30701:
@arcturus2 So I've fixed that bug, somewhere along the execution path it was initializing the array as empty, but not loading the look-ups. Now even if the array is empty, it re-checks the look-up.

>If the requirement is that a user has to set a preference for each rule then that will be a problem if they don't.

So straight out the box this Pr will use the default category linked with that rule. This is done via the rules "RISK" attribute.
If a preference is stored on that PC, then it uses the category defined by that rules preference.

>So what we generally do is have a default implementation

What do you mean by this? From my understanding my above comment is what a default implementation would mean. Let me know if I am reading it wrong 😆 

It could benefit from a "reset to default" button which allowed the user to use the default rule categories. Let me know if you agree.

arcturus2:
> > So what we generally do is have a default implementation
> 
> What do you mean by this? From my understanding my above comment is what a default implementation would mean. Let me know if I am reading it wrong 😆

All good. I was actually thinking you did this via the score but I went off on a tangent. What you have done is perfect. I will do more testing but all good.

arcturus2:
> It could benefit from a "reset to default" button which allowed the user to use the default rule categories. Let me know if you agree.

Based on testing and modifying my selection i think having a reset button would be extremely useful. I think debugging problems with users would be much easier by asking them to "reset to default" and try again first.

With regards to the placement of the button I'd vote to have it inside the dialog where you can customise the severity.

Also a few minor things.

1.On Linux the dialog window is a little smaller so the categories are elipsed. Could you look at extending the width if its not too much trouble?
2. Can you set a title for the dialog as it defaults to "Message".
3. Can we change the wording a bit from "Priority defines category levels for each rule" to "Customise the priority of rules" or something like that?

image

aldebaran30701:
I've gave the dialog a bit of extra space for other configurations. How does it look? 
image

arcturus2:
Looks much better. Thank you @aldebaran30701 

image

mimosa2:
Good work @aldebaran30701. From my testing it looks very good. Some minor fixes can be as follows;
1. The preference file entry (see below) can have a label such as "customRules=" . This is similar to recentValues
  ={"Missing type":"WARNING","Unknown type":"DEFAULT","Identifier inconsistent with type":"INFO"}
2. The alert box w/o any rules can have message such as 'There is no rules for this Identifier"

Rules_Alert

aldebaran30701:
> 1. The preference file entry (see below) can have a label such as "customRules=" . This is similar to recentValues
>    ={"Missing type":"WARNING","Unknown type":"DEFAULT","Identifier inconsistent with type":"INFO"}
> 2. The alert box w/o any rules can have message such as 'There is no rules for this Identifier"

Much appreciated for the feedback @mimosa2 
I've addressed these in a different ticket and PR as this has been since merged.  Please see #769 for the ticket and #770 for the PR.

PR #761, opened 2020-08-05
Fix for running query spinner disappears when saving graph

formalhaut69:
### Description of the Change
GraphNode keeps count of how many calls to turn off the loading spinner is made and only turns it off when it reaches zero. VisualGraphTopComponent just does it when it wants without checking. I've just added a simple change to make VisualGraphTopComponent call GraphNode's makeBusy method that keeps track of the count.


arcturus2:
Perfect @formalhaut69!

Issue #757, opened 2020-08-04
Allow <enter> to add new line in PluginParametersSwingDialog


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently, the swing dialog box created when executing a plugin interactively has the ok button focused by default. For most plugins this is ideal but there may be circumstances where this isn't the case. For example, if you have a string parameter that accepts multiple lines, a user will want to press enter to move to the next line. This won't be possible though with the ok button focused (as pressing enter will run the plugin).

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

PR #762, opened 2020-08-05
Unfocus Plugin Swing Dialog Ok button

antares1470:
### Description of the Change

This change adds the option when running a plugin interactively (i.e. through a swing dialog) to have the ok button unfocused in the dialog box. This enables the possibility of enter being able to be used to create a new line in a multi-line string parameter rather than simply executing the plugin.

This enables a new way to engage with plugins loaded up through a swing dialog.

### Benefits

Enables possibility of enter being used to create a new line

### Possible Drawbacks

Users used to hitting enter to execute the plugin may suffer some initial UX troubles if the ok button is unfocused. This is why I have made it an option rather than simply replacing the old way (so the ok button only needs to be unfocused as needed).

### Verification Process

1) Open Constellation and create a new graph
2) Go to build a sphere graph
3) Verify that the ok button is focused (and pressing enter runs the plugin)
4) Close Constellation
5) In the code, go to `SphereGraphBuilderAction.java` and at the PluginExecution line add `.withOkButtonFocused(false)`
6) Clean and build constellation
7) Repeat steps 1-2
8) Verify that ok is unfocused (and pressing enter doesn't run the plugin)

### Applicable Issues

#757

### DIFF
--- a/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginInteraction.java
+++ b/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginInteraction.java
@@ -236,7 +236,7 @@ public class DefaultPluginInteraction implements PluginInteraction, Cancellable 
     }
 
     @Override
-    public boolean prompt(final String promptName, final PluginParameters parameters) {
+    public boolean prompt(final String promptName, final PluginParameters parameters, final boolean okButtonFocused) {
         if (SwingUtilities.isEventDispatchThread()) {
             throw new IllegalStateException("Plugins should not be run on the EDT!");
         }
@@ -244,7 +244,11 @@ public class DefaultPluginInteraction implements PluginInteraction, Cancellable 
         boolean result = false;
 
         final PluginParametersSwingDialog dialog = new PluginParametersSwingDialog(promptName, parameters);
-        dialog.showAndWait();
+        if (okButtonFocused) {
+            dialog.showAndWait();
+        } else {
+            dialog.showAndWaitNoFocus();
+        }
         if (PluginParametersDialog.OK.equals(dialog.getResult())) {
             result = true;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Constellation Changes
 
+## 2020-08-01 Changes in August 2020
+* Updated a number of functions in Core Plugin Framework to include the option to unfocus the Ok button from the plugin swing dialog.
+
 ## 2020-07-01 Changes in July 2020
 * Added `AnalyticSchemaV4UpdateProvider` to upgrade `SchemaVertexType`s that have changed.
 * Added utility class `NotifyDisplayer` and static method `NotifyDisplayer#display` for use when displaying a `NotifyDescriptor` message box.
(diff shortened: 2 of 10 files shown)


arcturus2:
When you get a chance it would be great to get your thoughts on this @sirius-1 thank you.

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginInteraction.java
```diff
@@ -236,15 +236,19 @@ public boolean cancel() {
...
+    public boolean prompt(final String promptName, final PluginParameters parameters, final boolean okButtonFocused) {
         if (SwingUtilities.isEventDispatchThread()) {
             throw new IllegalStateException("Plugins should not be run on the EDT!");
         }
 
         boolean result = false;
 
         final PluginParametersSwingDialog dialog = new PluginParametersSwingDialog(promptName, parameters);
-        dialog.showAndWait();
+        if (okButtonFocused) {
```
cygnus-x-1, line 247: I hate this so much but have no better solution to offer :\

My only comment is that i would do away with the `okButtonFocused` parameter - you're always using a mutant JavaFX/Swing dialog here so the problem will always be a problem so just hard-code the no focus behaviour in.

antares1470: So the only way I can think of to remove the parameters is to make `okButtonFocused` in `PluginExecution` static. While it would be good to call the variable statically, we don't really want it to behave statically (as changing the value for one plugin would affect all subsequent plugin calls). That will mean that we need to either change the value back to true at the end of `showAndWaitNoFocus()` or having to always specify the value of `okButtonFocused` anytime a plugin is run interactively via the PluginExecution.

Curious as to whether you think this is better than the parameters option @cygnus-x-1? Personally I this isn't going to look clean either way

cygnus-x-1: I'm not sure I follow... What I was trying to get at is that any time you run a plugin interactively, it's going to create a `PluginParametersSwingDialog` (which is JavaFX in Swing) and you'll have this problem, so why not just hard-code it's default behaviour to unfocus the button rather than ask the developer to specify?

antares1470: The reason for having the option is a UX thing. While there are cases for having the ok button unfocused, I think these cases are exceptions rather than the new standard. I think having the ok button focused is still preferred for most plugins that are run interactively and so I don't want to take that option away (I personally tried some plugins like sphere graph builder without ok enabled and found myself still wanting to hit enter to execute the plugin)

cygnus-x-1: Well in that case what you've done is fine. My issue with all of this is that it's yet another thing developers need to remember when developing plugins, and it only exists because we're not handling Swing/JavaFX integration properly rather than because of a design decision. This whole thing is a stop-gap measure and should be fixed properly.

cygnus-x-1 (approved):

antares1470:
Going go with the alternate solution presented in #798 instead

PR #763, opened 2020-08-05
Bugfix for an NPE that occured when using the split nodes plugin

formalhaut69:
### Description of the Change
The split nodes plugin would throw an NPE 100% of the time when I used it on master. When updating the split nodes plugin with the new feature, @Auriga2 would experience this issue on occasion. The NPE occurs because when getting the splitParameters.get(SPLIT_PARAMETER_ID) we only checked if the parameter was null but not the string value so the string value could still return null. There is no associated ticket for this issue.

	at java.base/java.lang.String.contains(String.java:2036)
	at au.gov.asd.tac.constellation.views.dataaccess.plugins.clean.SplitNodesPlugin.query(SplitNodesPlugin.java:192)
[catch] at au.gov.asd.tac.constellation.views.dataaccess.templates.RecordStoreQueryPlugin.lambda$query$0(RecordStoreQueryPlugin.java:139)

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
@@ -181,7 +181,12 @@ public class SplitNodesPlugin extends RecordStoreQueryPlugin implements DataAcce
         final RecordStore result = new GraphRecordStore();
 
         final Map<String, PluginParameter<?>> splitParameters = parameters.getParameters();
-        final String character = splitParameters.get(SPLIT_PARAMETER_ID) != null ? splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() : "";
+        String character = "";
+        if (splitParameters.get(SPLIT_PARAMETER_ID) != null) {
+            if (splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() != null) {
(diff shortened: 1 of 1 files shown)


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -181,7 +181,12 @@ protected RecordStore query(final RecordStore query, final PluginInteraction int
         final RecordStore result = new GraphRecordStore();
 
         final Map<String, PluginParameter<?>> splitParameters = parameters.getParameters();
-        final String character = splitParameters.get(SPLIT_PARAMETER_ID) != null ? splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() : "";
+        String character = "";
+        if (splitParameters.get(SPLIT_PARAMETER_ID) != null) {
+            if (splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() != null) {
```
antares1470, line 186: I would just add this condition to the old tertiary operator.
i.e. `final String character = splitParameters.get(SPLIT_PARAMETER_ID) != null && splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() != null ? ...`

formalhaut69: Yeah, that looks cleaner. Thanks for that @antares1470

Issue #680, opened 2020-07-14
Levenshtein distance attribute parameter is broken


Nova-2119:
### Description

The attribute selection parameter for the Levenshtein Distance analytic plugin displays transaction attributes where it should display node attributes.

### Steps to Reproduce

None given

**Expected behaviour:** [What you expect to happen]
Node attributes should be displayed.

**Actual behaviour:** [What actually happens]
Transaction attributes are displayed.

**Reproduces how often:** [What percentage of the time does it reproduce?]
Unknown

serpens24:
The list of attirbutes that user can select is determined in CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/analytics/LevenshteinDistanceAnalytic.java https://github.com/constellation-app/constellation/blob/master/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/analytics/LevenshteinDistanceAnalytic.java, in the following block of code:
![image](https://user-images.githubusercontent.com/58677108/89149772-6be4e980-d59c-11ea-9bd7-f7b2aa9af9fd.png)

serpens24:
We discussed this in weekly standup meeting ...... the bullet point list of attributes is being constructed by getting list of attributes registered for the schema .. this is not what we want, what we want is the lost of VERTEX atrributes actually in use for actual nodes/veretexes ... ie - more in line with attribute editor list (but only those that are strings)

PR #764, opened 2020-08-06
Update to the Levenshtein Distance Analytic

formalhaut69:
### Description of the Change
The Levenshtein Distance Analytic will now load attributes that are relevant to the nodes that are currently in the graph. A flaw in this solution is that the analytic view will only load the attribute options once and will never update it. If anyone could suggest a way to manually refresh the Levenshtein distance analytic whenever the graph changes then that would completely resolve this issue. 

image

### Applicable Issues

https://github.com/constellation-app/constellation/issues/680


CoreAnalyticView/nbproject/project.xml
```diff
@@ -94,6 +94,14 @@
                         <specification-version>1.0</specification-version>
                     </run-dependency>
                 </dependency>
+                <dependency>
```
arcturus2, line 97: So we can't do this @formalhaut69 because it means that the Analytic View becomes tied to the Attribute Editor. This becomes a problem should we decide to replace the Attribute Editor with a different implementation module; we can't because it will want to remove the Analytic View too.

image

FYI all of the views now are not dependant on one another so we gotta keep it that way.

Perhaps look at the implementation in `RemoveUnusedAttributesPlugin` for inspiration and you might need to create a new utility method that could live somewhere in the `Core Graph Framework` module so it can be used by views and any plugin in the future.

CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/analytics/LevenshteinDistanceAnalytic.java
```diff
@@ -86,12 +86,10 @@ public void onPrerequisiteAttributeChange(final Graph graph, final PluginParamet
...
             try {
-                final Map<String, SchemaAttribute> attributes = graph.getSchema().getFactory().getRegisteredAttributes(GraphElementType.VERTEX);
-                for (final String attributeName : attributes.keySet()) {
-                    final SchemaAttribute attribute = attributes.get(attributeName);
-                    final String attributeType = attribute.getAttributeType();
-                    if (StringAttributeDescription.ATTRIBUTE_NAME.equals(attributeType)) {
-                        stringAttributes.add(attributeName);
+                final ArrayList<AttributeData> nodeAttributes = new AttributeReader(graph).refreshAttributes().getAttributeNames().get(GraphElementType.VERTEX);
+                for (AttributeData nodeAttribute : nodeAttributes) {
+                    if (nodeAttribute.getDataType().equals("string")) {
```
arcturus2, line 91: Please use `StringAttributeDescription.ATTRIBUTE_NAME` instead of "string".

CoreAnalyticView/nbproject/project.xml
formalhaut69: Okay, no worries @arcturus2 . I'll try and figure out an alternate way to do this without making any new dependencies.

formalhaut69:
@arcturus2, I figured out a way to get the node attributes without the dependency on the attribute editor. It still doesn't update the options though. This can be an issue because if the user opens the analytic view with an empty graph, the only option will be "identifier" and it will be permanently stuck like that (unless the user reloads the application with a populated graph). You also mentioned that it might be a good idea to make a method in the core graph framework that does this. Perhaps if you believe that there is a use case here I could make a simple class that can get attributes based on it's graphElementType.

arcturus2:
>It still doesn't update the options though. This can be an issue because if the user opens the analytic view with an empty graph...

Yep this is still a problem. Perhaps, so we don't forget can you either create a new ticket for this issue? If so, please reference this ticket for history.

> Perhaps if you believe that there is a use case here I could make a simple class that can get attributes based on it's graphElementType and then update this analytic to use it

Yeah I think that would be great. Name the class with *Utilities method perhaps. One of the reasons we have less code re-use is because common things are not written in a place they can be called like a utility method. You saw that when you had to use a method from the Attribute Editor.

It is a time consuming thing to go through the code and refactor it for all cases but I've tried to stay on top of new instances (like this one). If you have time then please go ahead and make a utility method and can you see if perhaps the Attribute Editor could also use this? So that would be 2 uses.

Issue #628, opened 2020-07-06
Analytic View Null Pointer


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

I opened the analytic view and the following null pointer exception occurred:

```
java.lang.NullPointerException
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.lambda$updateSelectablePluginsParameters$29(AnalyticConfigurationPane.java:534)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.updateSelectablePluginsParameters(AnalyticConfigurationPane.java:530)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.setPluginsFromSelectedQuestion(AnalyticConfigurationPane.java:515)
	at au.gov.asd.tac.constellation.views.analyticview.AnalyticConfigurationPane.lambda$new$13(AnalyticConfigurationPane.java:209)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
[catch] at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to Reproduce

1. Open a new graph
2. Open the analytic view

**Expected behaviour:** The view should open without any errors

**Actual behaviour:** A popup with a null pointer error was displaued

[...]

PR #765, opened 2020-08-07
:bug: add lock to access of question - Null pointer

aldebaran30701:
### Description of the Change
Add a lock to the access of `currentQuestion`
Initially reproduced by making a new graph on fresh constellation boot. 
using sphere graph builder
launching analytic view

NPE would then occur 100% of the time.
#628

### Alternate Designs
N/A

bugfix

### Benefits
bugfix

### Possible Drawbacks
none- realised

### Verification Process
Test opening and closing of analytic view + reproducer

### Applicable Issues
#628

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticConfigurationPane.java
@@ -522,25 +522,20 @@ public class AnalyticConfigurationPane extends VBox {
     }
 
     public final void updateSelectablePluginsParameters() {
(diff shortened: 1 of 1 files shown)


formalhaut69:
I got the NPE again so @aldebaran30701 made another change to try and resolve it. After the change I could no longer reproduce the NPE but based on how this issue has gone I wouldn't get my hopes up 😆

arcturus2:
We'll get there!

arcturus2:
Thank you got this.

However, these `lock` implementations is a work around solution and I only implemented this in the Map View a while back because the code that wasn't thread safe was with a 3rd party code (which we can't change).

The analytic view code is ours and we should look at how we can make it all thread safe. I've created #767 to address the long term solution.

Issue #743, opened 2020-07-27
Split Nodes plugin enhancement


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently when you run the split nodes plugin, any transactions hanging off the old node will be attached to one of the new nodes formed and the other node(s) will have a transaction to this node. There are situations though where this way of graphing the split isn't the desired behaviour.

In this new way to graph the split nodes, transactions hanging off the old node would go to all the new nodes created (not just one of them) with no transactions between the new nodes. In the plugin, there should be a parameter to decide between the old and new way of graphing the split.

e.g. A sample graph
![graph1](https://user-images.githubusercontent.com/62311243/88501753-acc68680-d00f-11ea-8b05-bcc5f212305b.png)

The current behaviour of Split Nodes
![graph2](https://user-images.githubusercontent.com/62311243/88501781-c2d44700-d00f-11ea-8f3c-7685c8773b30.png)

The suggested alternative behaviour
![graph3](https://user-images.githubusercontent.com/62311243/88501783-c5cf3780-d00f-11ea-9bc5-d2963492b946.png)

### Additional Information

If a user selects the new way to display the split, the Transaction Type parameter should be disabled as this won't be needed

Auriga2:
I need to use `edit` method instead of the `query` because the `RecordStoreQueryPlugin` only sends details of the "selected" nodes (transactions or all) in the`RecordStore`. To implement the new functionality I need to know all the transactions adjacent to the selected node.

I guess if we choose to  split the below node highlighted in the left, it should result the graph at the right:

![image](https://user-images.githubusercontent.com/64826928/89359156-470b8600-d704-11ea-8c4c-a68e95dca92a.png)

arcturus2:
@Auriga2 agreed, using the `edit` method is what I would use too. Thank you for the diagram, it makes it so much easier to understand; and that diagram to the right is what this issue is expecting to have.

[...]

PR #773, opened 2020-08-11
Split nodes plugin enhancement - Split nodes into same level

Auriga2:
### Description of the Change
A new option `Split Nodes into same level` is added in to the Split nodes plugin, enabling which will split nodes such that the new nodes are hanging off from the same parent node rather than from one of the new nodes.

### Alternate Designs
N/A

Better UX

### Benefits
Better UX

### Possible Drawbacks
N/A

### Verification Process
- Open a graph and open Attribute Editor 
- Renamed some nodes to a delimited string  such as 'abc;def;ghi'
- Open up Data Access View and go to Split Nodes plugin
- Set the split character as ';' and select the  `Split Nodes into same level` option
- Ensure the renamed node/s is/are selected and run the plugin
- Ensure all the new nodes have the same nodes connected via similar transactions as the selected node

### Applicable Issues
https://github.com/constellation-app/constellation/issues/743


Auriga2:
@arcturus2 I noticed in V2.0.0 that if you do not select the `Split on All Occurrences` and run the plugin multiple timed on the same node to split further, OR on different nodes to split them, all the new nodes are added on top of each other.  This versions also does the same thing if  `Split on All Occurrences`  is not selected, irrespective of the new option. Should we fix it, or does it not matter?

arcturus2:
> 
> 
> @arcturus2 I noticed in V2.0.0 that if you do not select the `Split on All Occurrences` and run the plugin multiple timed on the same node to split further, OR on different nodes to split them, all the new nodes are added on top of each other. This versions also does the same thing if `Split on All Occurrences` is not selected, irrespective of the new option. Should we fix it, or does it not matter?

So @Auriga2 from a UX perspective it should be fixed but this is a problem outside the scope of this plugin. The scope of the a plugin execution should be limited to what data it has and when it finishes its execution and returns the new recordstore, the framework should be placing the nodes in a location not ontop of each other.

I'd vote to create a new ticket to discuss this problem and see how we can solve it for all situations.

Auriga2:
@arcturus2: I'll create a new ticket. But with the new change, I'm no longer using the recordstore, as I have access to the graph in the edit method.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
-                //as the first node
+        for (int vertexAttributePosition = 0; vertexAttributePosition < vertexAttributeCount; vertexAttributePosition++) {
+            int vertexAttributeId = graph.getAttribute(GraphElementType.VERTEX, vertexAttributePosition);
+            String vertexAttributeName = graph.getAttributeName(vertexAttributeId);
+
+            if (vertexAttributeName.equals(VisualConcept.VertexAttribute.IDENTIFIER.getName())) {
+                graph.setStringValue(vertexIdentifierAttributeId, newVertexId, newNodeIdentifier);
+            } else if ((VisualConcept.VertexAttribute.X.getName()).equals(vertexAttributeName)) {
+                //The X coordinate is skipped so that the second node is not created in the exact same location
+                //as the first node. Copiying the Y & Z for a better arrangement as it looks less cluttered.
```
arcturus2, line 168: A minor spelling typo, should be "Copying"

arcturus2:
@Auriga2 please also note the build is failing. See https://travis-ci.com/github/constellation-app/constellation/builds/179228723

Auriga2:
@arcturus2 I don't get any build errors locally. How could I troubleshoot?  Could it be related to the IndexOutOfBoundsException mentioned above?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -94,6 +95,12 @@ public PluginParameters createParameters() {
         split.setStringValue(null);
         params.addParameter(split);
 
+        final PluginParameter<BooleanParameterValue> splitIntoSameLevelParameter = BooleanParameterType.build(SPLIT_INTO_SAME_LEVEL_PARAMETER_ID);
+        splitIntoSameLevelParameter.setName("Split Nodes into same level");
+        splitIntoSameLevelParameter.setDescription("Choose to split Nodes hanging off from the same parent node rather than from one of the new nodes");
```
antares1470, line 100: I'm wondering if there is a better name and description for this? I'm not sure whether this makes sense to a user without already knowing what it does. I can't think of anything off the top of my head though

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
-            types.put(SchemaVertexTypeUtilities.matchVertexTypes(left).get(0), left);
-        }
-        if (rightVertexTypesMatches.size() > 0) {
-            types.put(SchemaVertexTypeUtilities.matchVertexTypes(right).get(0), right);
-        }
-        final List<SchemaVertexType> ordered_types = new ArrayList<>(types.keySet());
+    private int createNewNode(final GraphWriteMethods graph, final int selectedNode, final String newNodeIdentifier, final String linkType, final boolean splitIntoSameLevel) {
+        int vertexIdentifierAttributeId = graph.getAttribute(GraphElementType.VERTEX, VisualConcept.VertexAttribute.IDENTIFIER.getName());
+        int transactionTypeAttributeId = graph.getAttribute(GraphElementType.TRANSACTION, AnalyticConcept.TransactionAttribute.TYPE.getName());
+        int vertexAttributeCount = graph.getAttributeCount(GraphElementType.VERTEX);
```
antares1470, line 154: these 3 can be made final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
 
-        result.add();
-        result.set(GraphRecordStoreUtilities.SOURCE + GraphRecordStoreUtilities.ID, query.get(SOURCE_ID));
-        result.set(GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.IDENTIFIER, left);
-        if (ordered_types.size() > 0 && leftVertexTypesMatches.size() > 0) {
-            result.set(GraphRecordStoreUtilities.SOURCE + AnalyticConcept.VertexAttribute.TYPE, leftVertexTypesMatches.get(0));
-        }
-        result.set(GraphRecordStoreUtilities.DESTINATION + VisualConcept.VertexAttribute.IDENTIFIER, right);
+        // Add new node
+        int newVertexId = graph.addVertex();
```
antares1470, line 157: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
-            if (key.endsWith(".[id]") || SOURCE_IDENTIFIER.equals(key)) {
-                //Skips the id and Identifier to make the new node unique
-            } else if ((GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.X).equals(key)
-                    || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Y).equals(key)
-                    || (GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.Z).equals(key)) {
-                //The coordinates are also skipped so that the second node is not created in the exact same location
-                //as the first node
+        for (int vertexAttributePosition = 0; vertexAttributePosition < vertexAttributeCount; vertexAttributePosition++) {
+            int vertexAttributeId = graph.getAttribute(GraphElementType.VERTEX, vertexAttributePosition);
+            String vertexAttributeName = graph.getAttributeName(vertexAttributeId);
```
antares1470, line 162: these 2 can be final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
             } else {
-                result.set(GraphRecordStoreUtilities.DESTINATION + key.replace(GraphRecordStoreUtilities.SOURCE, ""), query.get(key));
+                graph.getNativeAttributeType(vertexAttributeId).copyAttributeValue(graph, vertexAttributeId, selectedNode, newVertexId);
             }
         }
 
-        if (ordered_types.size() > 1 && rightVertexTypesMatches.size() > 0) {
-            result.set(GraphRecordStoreUtilities.DESTINATION + AnalyticConcept.VertexAttribute.TYPE, rightVertexTypesMatches.get(0));
+        // Transactions
+        final boolean directed = false;
```
antares1470, line 175: if directed can't ever be true, it's probably better to remove this variable and just simply add false into the relevant function calls

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+        if (splitIntoSameLevel) {
+            int transactionCount = graph.getVertexTransactionCount(selectedNode);
```
antares1470, line 177: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+            for (int transactionPosition = 0; transactionPosition < transactionCount; transactionPosition++) {
+                int originalTransactionId = graph.getVertexTransaction(selectedNode, transactionPosition);
+
+                int sourceVertex = graph.getTransactionSourceVertex(originalTransactionId);
+                int destinationVertex = graph.getTransactionDestinationVertex(originalTransactionId);
```
antares1470, line 182: these 3 can be final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+                if (sourceVertex == selectedNode && destinationVertex == selectedNode) {
+                    newTransactionId = graph.addTransaction(newVertexId, newVertexId, directed);
+                } else if (sourceVertex == selectedNode) {
+                    newTransactionId = graph.addTransaction(newVertexId, destinationVertex, directed);
+                } else if (destinationVertex == selectedNode) {
+                    newTransactionId = graph.addTransaction(sourceVertex, newVertexId, directed);
+                }
+
+                //Loops through all the transaction attributes and copies them to the new transaction
+                int transactionAttributeCount = graph.getAttributeCount(GraphElementType.TRANSACTION);
```
antares1470, line 193: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+                for (int transactionAttributePosition = 0; transactionAttributePosition < transactionAttributeCount; transactionAttributePosition++) {
+                    int transactionAttributeId = graph.getAttribute(GraphElementType.TRANSACTION, transactionAttributePosition);
+                    String transactionAttributeName = graph.getAttributeName(transactionAttributeId);
```
antares1470, line 196: these 2 can be final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+                    if (transactionAttributeName.equals(VisualConcept.TransactionAttribute.IDENTIFIER.getName())) {
+                        graph.setStringValue(transactionAttributeId, newTransactionId, Integer.toString(newTransactionId));
+                    } else {
+                        graph.getNativeAttributeType(transactionAttributeId).copyAttributeValue(graph, transactionAttributeId, originalTransactionId, newTransactionId);
+                    }
+                }
+            }
+        } else {
+            //Add a transaction of the type 'linkType' between the selected node and new node
+            int newTransactionId = graph.addTransaction(selectedNode, newVertexId, directed);
```
antares1470, line 206: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
         final Map<String, PluginParameter<?>> splitParameters = parameters.getParameters();
         final String character = splitParameters.get(SPLIT_PARAMETER_ID) != null && splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() != null ? splitParameters.get(SPLIT_PARAMETER_ID).getStringValue() : "";
         final ParameterValue transactionTypeChoice = splitParameters.get(TRANSACTION_TYPE_PARAMETER_ID).getSingleChoice();
         final String linkType = transactionTypeChoice != null ? transactionTypeChoice.toString() : AnalyticConcept.TransactionType.CORRELATION.getName();
         final boolean allOccurrences = splitParameters.get(ALL_OCCURRENCES_PARAMETER_ID).getBooleanValue();
+        final boolean splitIntoSameLevel = splitParameters.get(SPLIT_INTO_SAME_LEVEL_PARAMETER_ID).getBooleanValue();
+
+        int vertexSelectedAttributeId = graph.getAttribute(GraphElementType.VERTEX, VisualConcept.VertexAttribute.SELECTED.getName());
+        int vertexIdentifierAttributeId = graph.getAttribute(GraphElementType.VERTEX, VisualConcept.VertexAttribute.IDENTIFIER.getName());
+        List<Integer> newVertices = new ArrayList<>();
```
antares1470, line 223: these 3 can be final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+
+        final int graphVertexCount = graph.getVertexCount();
+        for (int position = 0; position < graphVertexCount; position++) {
+            int currentVertexId = graph.getVertex(position);
```
antares1470, line 227: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+
+            if (graph.getBooleanValue(vertexSelectedAttributeId, currentVertexId)) {
+                final String identifier = graph.getStringValue(vertexIdentifierAttributeId, currentVertexId);
+
+                if (identifier != null && identifier.contains(character) && identifier.indexOf(character) < identifier.length() - character.length()) {
+                    String leftNodeIdentifier = "";
+                    if (allOccurrences) {
+                        final List<String> substrings = new ArrayList<>(Arrays.asList(identifier.split(character)));
```
antares1470, line 235: given the use of `substrings`, I would probably just leave it as an array rather than converting it to a list

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
Auriga2: Agree, but I couldn't think of anything better either.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
Auriga2: Whatever the value added here will be overwritten when copying the attributes. Just wanted a default value here. I vaguely remember Sanarqube throwing a code smell  when I added "false" like value in few lines without using a variable.

antares1470: @Auriga2 Ok I think I understand what you are doing here.

So what you have done works, but it's not optimal. The directed parameter of addTransaction() is actually the direction under the hood (e.g. Graph.UPHILL, Graph.UNDIRECTED) and not the value of the directed transaction attribute. Copying the value of the transaction parameter works because when you complete schema, the function checks for a mismatch in transaction attribute value vs under the hood value (e.g. transaction attribute is true but the under the hood value is Graph.UNDIRECTED) and creates a copy of the transaction with the correct under the hood value (removing the old one) if a mismatch exists. This means that with what you currently have, any directed transactions you are copying are effectively being created twice (once by your function and once on complete schema).

I would suggest then grabbing the value of the directed attribute beforehand and passing that value in to addTransaction() to ensure the transaction only gets added once

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
cygnus-x-1: I've looked over the code and still can't understand what this is supposed to be doing - could you explain it?

arcturus2: Have you looked at the actual ticket #743. There are images too!

Auriga2:
@arcturus2 : I managed to recreate the index error. It happens if there's no resulting sub-string. E.g. I had a node `vv` and I spitted with `v` .
Should we leave node `vv` in this case? May be with a warning message?
If there's a node `vwx`, should we just rename it to `wx` when splitting with `v`?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
-                    final int i = identifier.indexOf(character);
-                    final String left = identifier.substring(0, i);
-                    final String right = identifier.substring(i + 1);
-                    editResultStore(result, left, right, query, linkType);
+                    // Rename the selected node
+                    graph.setStringValue(vertexIdentifierAttributeId, currentVertexId, leftNodeIdentifier);
                 }
             }
         }
+        if (!newVertices.isEmpty()) {
```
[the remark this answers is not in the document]
Auriga2: Added it to avoid resetting the view when there's no matching node to split. I can remove it if you think it's okay to just flicker the graph sometimes.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
Auriga2: @antares1470  I don't quite get it. If it's `Graph.UPHILL`, `Graph.UNDIRECTED` etc, and not `true` or `false`, then it's still the same thing when I grab and use the directed attribute (which is` true` or `false`) when creating transaction.

Auriga2: image
 I don't get the relation of this boolean to the `GraphConstants`.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -130,82 +144,132 @@ public void updateParameters(Graph graph, PluginParameters parameters) {
...
+                arrangement.executeNow(vlGraph.getInclusionGraph());
+                vlGraph.retrieveCoords();
+            }
+
+            PluginExecution.withPlugin(VisualSchemaPluginRegistry.COMPLETE_SCHEMA).executeNow(graph);
+            PluginExecutor.startWith(InteractiveGraphPluginRegistry.RESET_VIEW).executeNow(graph);
+        }
     }
+
+    public PluginExecutor completionArrangement() {
```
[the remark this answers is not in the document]
Auriga2: @cygnus-x-1 : I tried to move `completionArrangement()` into the `SimpleEditPlugin` but addding the `Core Arrangement Plugins` there throws this warning. `RecordStoreQueryPlugin`  is different as it's in `Core Data Access View` module.
image

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
antares1470: @Auriga2 so when you add a transaction, the directed parameter determines the under-the-hood direction. The value of the directed transaction attribute is then determined from this under-the-hood value. The main issue is that this under-the-hood value can't actually be changed on a transaction (and so changing the value of the directed transaction attribute after the fact doesn't actually change this value). Consequently when you do change the directed transaction value, a new transaction needs to created with the correct under-the-hood value. This is the relevant code in `completeTransaction()` in `AnalyticSchema` that does that:
```
final Boolean directed = graph.getBooleanValue(transactionDirectedAttribute, transactionId);
final boolean transactionIsDirected = graph.getTransactionDirection(transactionId) != Graph.FLAT;
if (directed != transactionIsDirected) {
   <code to create clone of transaction>
}
```
What this means is that when adding a transaction, it's better to pass through the right value for the directed parameter rather than leaving it to `completeSchema` to fix it after the fact.

Does that make sense?

Auriga2: It makes sense to provide the correct value beforehand, and verified that it does no longer attempt to remove and add the transactions again by the above code.

antares1470 (changes requested):
Missed a couple of finals from my previous review.

Also would be good if you are able to add a whats new entry for this

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,9 @@
 == 3030-12-31 Getting Started
 <p>If you're new to CONSTELLATION, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.introduction">introduction</a>.</p>
 
+== 2020-08-28 Split Nodes Plugin
+<p>A new option is added to split nodes with duplicate transactions.</p>
```
antares1470, line 5: The first interpretation that came to mind when reading this was that it splits nodes that have duplicate transactions rather than duplicating the transactions of the split node for all the newly created nodes. Might be worth slightly altering the entry just to make the latter (correct) interpretation of the functionality clearer.

antares1470 (changes requested):
If you are also able to handle the merge conflicts that would be great.

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/whatsnew/whatsnew.txt
Auriga2: Fixed

antares1470 (approved):

arcturus2 (approved):

Issue #768, opened 2020-08-10
Levenshtein Distance Analytic doesn't update 


formalhaut69:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Levenshtein distance analytic doesn't update when changes to the graph are made. It's onPrerequisiteAttributeChange method should be getting called to update the analytic along with it's UI. 

### Steps to Reproduce

1. Make a change to the graph
2. Open the Levenshtein distance analytic's options

**Expected behaviour:** The options will update to reflect what is currently in the graph

**Actual behaviour:** The options are permanently stuck as whatever was loaded into the UI originally

**Reproduces how often:** 100%

### Additional Information
There are several other classes that use this method so this issue may not be exclusive to the Levenshtein distance analytic

formalhaut69:
https://github.com/constellation-app/constellation/pull/764 PR for progress with fixing the Levenshtein distance analytic

antares1470:
Have created #842 which was found while testing this

PR #781, opened 2020-08-14
single choice parameter type not updating fix

aldebaran30701:
### Description of the Change
This change allows the get and set of object value within `SingleChoiceParameterType` to be used to update the `PluginParameter`. Previously, on updating of the plugin parameters through a graph update would cause only the current selection to be updated. Now, it will update all possible choices and the current selection.
This change also includes a null check which was found through testing.

### Alternate Designs
Add an `updateParameters()` method within all `ParameterTypes` which is called on updateParameters calls. This would allow any current usages of `getObjectValue()` to remain unchanged.

bugfix

### Benefits
Fixes the issue where plugin values are not updated on graph changes

### Possible Drawbacks
Possible usage of `getObjectValue()` elsewhere

### Verification Process
Tested opening and closing graphs, selecting and running plugin.

### Applicable Issues
#768

### DIFF
--- a/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/sna/similarity/LevenshteinDistancePlugin.java
+++ b/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/sna/similarity/LevenshteinDistancePlugin.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.commons.lang3.StringUtils;
 import org.openide.DialogDisplayer;
 import org.openide.NotifyDescriptor;
 import org.openide.util.NbBundle;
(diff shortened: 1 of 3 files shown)


Nova-2119:
@arcturus2 @aldebaran30701 
The changes to SingleChoiceParameterType have created a couple of bugs. I think this is the cause of the bug https://github.com/constellation-app/constellation/issues/793 as well as a bug in the histogram view. 

Tomorrow I will see if I can fix the bugs whilst keeping the intention of these changes, @aldebaran30701 I may need to hit you up for some help if I am struggling or need to make alterations to these changes :)

arcturus2:
@Nova-2119 @aldebaran30701 

Yeah just hit a similar problem and I will confess that I didn't test this at all and just merged it in.

The `getObjectValue` is now returning the `SingleChoiceParameterType` whereas it used to return the `ParameterValue`. I'm OK with the change given it has made it consistent with `MultiChoiceParameterType`. What we will need to do is go through instances and chain a call to `getChoiceData`.

The CHANGELOG.md will also need to be updated given this was an API change. Can I ask for a favor and leave this one with both of you to work out please?

aldebaran30701:
@arcturus2 Yep completely understand, We are looking at it now. 👍

Issue #766, opened 2020-08-07
Table View bugs when spamming Selected Only mode button


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

If you spam click the selected only button in table view, you occassionally see rows of data duplicating and eventually, you get a null pointer which renders the table view useless.

### Steps to Reproduce

1. Open a new graph
2. Draw a node and have the node selected
3. Open table view and change it to node mode
4. Spam the selected only button until you get a null pointer. Observe you occassionally see a duplicate row further down the table along the way

**Expected behaviour:** One row for each table item, no null pointers

**Actual behaviour:** Occassional duplicate entry, null pointers rendering table view useless

**Reproduces how often:** 100% (eventually)

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

aldebaran30701:
Closing #777 as this ticket encapsulates the problem.

Other reproducer:
1.Open table view
2.Open new Graph
3.Sphere graph build (1000 vertices, 500 transactions)
4.Drag cell headings left of right quickly to replicate the null pointer

```
java.lang.NullPointerException
	at javafx.controls/javafx.scene.control.skin.CellSkinBase.access$100(CellSkinBase.java:56)
	at javafx.controls/javafx.scene.control.skin.CellSkinBase$StyleableProperties$1.isSettable(CellSkinBase.java:166)
	at javafx.controls/javafx.scene.control.skin.CellSkinBase$StyleableProperties$1.isSettable(CellSkinBase.java:161)
	at javafx.graphics/javafx.scene.CssStyleHelper.transitionToState(CssStyleHelper.java:666)
	at javafx.graphics/javafx.scene.Node.doProcessCSS(Node.java:9647)
	at javafx.graphics/javafx.scene.Node.access$900(Node.java:398)
	at javafx.graphics/javafx.scene.Node$1.doProcessCSS(Node.java:471)
[...]

Issue #580, opened 2020-06-16
Delimited File Import: Preview table issues


serpens24:
### Description

When importing a small set of data the preview window shows duplicate entires

### Steps to Reproduce

1. Create a file containing the follwoing, and call it tst.csv
`
a,b,c,d
1,2,3,4
5,6,7,8`
2. Select 'Import' -> 'From Delimited File' and ensure CSV is selected.
3. Press the + button, locate the file and click 'Open'
4. Duplicate entries are seen:
![image](https://user-images.githubusercontent.com/58677108/84725990-02e4ea80-afcb-11ea-8593-e7dbc9d6fba4.png)

**Expected behaviour:** 
no duplicate entries

**Actual behaviour:**
As per image above, duplicates of early rows are seen in reverse order from the bottom of the table. i've seen this sort of thing before in tableview, so suspect theres something systematic.

**Reproduces how often:**
100% I think

Auriga2:
@serpens24 I cannot recreate this. Tried debugging the latest in master and the released v2.0.0-beta5 exe.
![image](https://user-images.githubusercontent.com/64826928/85099724-52325180-b23d-11ea-879c-8a9e3a722ea7.png)

Auriga2:
serpens24 means to use the 'Enter' instead of the 'spaces' in the csv file. 
But still cannot recreate if you just try the `+ button` once. You need to add one more file, or Use `- button` and add it again, to recreate the issue.
When there are multiple files added, the mirror image even appear/disappear randomly when clicking between the files.

GCHQDeveloper601:
I had a play around with this issue today. Not sure of the root cause of it but calling refresh on the panel after setting the data seems to resolve the issue. I'll create a PR for it now.

PR #783, opened 2020-08-14
Bug 580 fix - refresh TableView

GCHQDeveloper601:
### Description of the Change
A fix for issue 580. Stops the TableView from having duplicate (reversed) values further down the table.

### Alternate Designs
None considered, found the cause and implemented the solution

This is a fix to an issue affecting Core.

### Benefits
Makes importing data look nicer.

### Possible Drawbacks
None found in testing.

### Verification Process

Tested locally, replicated actions to create the issue several times and error did not recur. 
To replicate 
*Open the "Import from Delimited File" Window
--Start
*Clicked Green plus
*Found sample file and clicked Open.
*Selected File in Files panel
*Clicked Red Minus
*Clicked Green plus
*Found sample file and clicked Open.
*Repeat from START with several different files

### Applicable Issues

N/A


aldebaran30701:
Linking this PR to #580 
This also seems similar to #766 in the fact that when it loads it sometimes has duplicate entries mirrored.

Issue #752, opened 2020-08-03
Graph dependant drop downs in DAV not updating when loading or switching to another graph


Auriga2:
### Description

Sometimes the below drop downs do not refresh if you open the graph after the Data Access View. Also when you switch between graphs these drop downs do not refresh even though `updateParameters` is called. As a workaround, typing a missing item adds it to the list, but cannot get rid of it by switching to the other graph. 
If you have no idea about the missing items, the workaround to refresh the list is `close the Data Access view, then close Constellation and then open Constellation again and then load the graph and then open DAV`. 
- `Transaction Type` in `Split Nodes Plugin`
- `Words to Extract` in `Extract words from text plugin`

### Steps to Reproduce
As above

**Expected behaviour:** The drop downs should refresh when loading a graph or switching to another graph

**Actual behaviour:** The drop downs do not refresh

**Reproduces how often:** 90%

### Additional Information

arcturus2:
The way it's meant to work is that `updateParameters` should be called when you open a graph or switch graphs. This method gets a chance to update the parameters it has in context of the graph you have. The Analytic schema has registered transaction types defined and the Split Nodes Plugin drop down should update. If you then go to File -> New Graph -> Visual Graph (which has no transaction types registered) it should call `updateParameters` and set the drop down list to empty.

I noticed whilst testing just now that I had the same issue you described above (which means it is broken) but then ran it in debug mode and it seemed to work. That is, the drop down list did populate after opening up an Analytic graph with the DAV already open.

I then tried to open a Visual Graph and noticed that though `updateParameters` was called, the drop down list did not go back to being empty. This is a bug.

In general, any time a graph is open or you switch to another graph all views need to update. For example, the DAV updates its drop down lists. Another example is the Schema View which interrogates the schema on the graph and updates its registered attributes, node and transaction types in the tabs.

You should not need to open the views or graph in a specific way for them to update. That is another bug then @Auriga2.

aldebaran30701:
I can change the code to allow the DAV to load the backing list for the dropdowns.

***Without SuppressEvent***
There is a piece of code which sets a default to 'CORRELATION' - this without the `suppressEvent(...)` code will trigger the `Split Nodes Plugin` to become 'checked' in the DAV. It will then tell the SingleChoiceInput to update through it's listeners. 

***With SuppressEvent***
The default value will be set, but not update the UI as the event would not bubble to the UI to update it. This will set the value correctly and also not trigger the plugin to become 'checked' in the DAV

[...]

PR #788, opened 2020-08-18
DAV plugins parameter options not loading fix

aldebaran30701:
### Description of the Change
This change allows the updating of plugin parameters to be in effect when loading a graph.
Previously, it did not load when a graph changed. 

See alternate designs for implications

### Alternate Designs
This design does not set the default option as "CORRELATION" for `SplitNodesPlugin` as the `suppressEvent()` call stops the triggering of a value change event. 

If you would like the DAV to preselect default options on graph load - the `suppressEvent()` call will have to be removed. Removing these calls will then trigger the DAV to select that plugin as 'checked'. This will be odd for the user and is the main reason the `suppressEvent()` call was implemented. 

It appears we may have to choose between functionality.

Loads parameter list options

### Benefits
Loads parameter list options

### Possible Drawbacks
See alternate designs

### Verification Process
Testing of open and closed graphs, checked all other instances of `updatePatameters(...)`

### Applicable Issues
#752

### DIFF
--- a/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/sna/similarity/LevenshteinDistancePlugin.java
+++ b/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/sna/similarity/LevenshteinDistancePlugin.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


mimosa2:
@aldebaran30701, I have tested this and noticed that the drop down refreshes when switching the graphs.

antares1470 (approved):

cygnus-x-1 (approved):

mimosa2 (approved):

PR #790, opened 2020-08-18
Add transactions into existing hashmod plugin

CrucisGamma:
### Description of the Change

Modifies HashMod to be able to specify that a certain field should be matched to another field if the values match.  Essentially a 
foreign key type arrangement (ManagerId -> EmployeeId for example)

### Alternate Designs

To do things manually 

Hashmod is already in the core and this is extending it to allow transactions to be put into the mix as well as vertices.

### Benefits

Users will be able to specify that a field in a vertex should match another field's value and then a transaction will be created.

### Possible Drawbacks

None foreseen (as the user controls what gets linked in)

### Verification Process

Have implemented it twice actually and have checked it against a few different data sets that I can think of (floor plans and a made up manager-employee structure)

### Applicable Issues

Nil known.


CoreGraphUtilities/src/au/gov/asd/tac/constellation/graph/utilities/hashmod/Hashmod.java
```diff
@@ -74,7 +75,7 @@ public void setCSVFileStr(final String csvFileStr) {
     }
 
     public String[] getCSVFileHeaders() {
-        if (CollectionUtils.isNotEmpty(data)) {
+        if (data != null && !data.isEmpty()) {
```
arcturus2, line 78: @CrucisGamma did the CollectionUtils not work as expected? If so what was the issue?

PR #791, opened 2020-08-18
Add transactions into existing hashmod plugin

CrucisGamma:
### Description of the Change

Modifies HashMod to be able to specify that a certain field in a vertix should be matched to another field in a vertex (and hence draw a transaction between them) if the values match.  Essentially a foreign key type arrangement (ManagerId -> EmployeeId for example)

### Alternate Designs

To do things manually 

Hashmod is already in the core and this is extending it to allow transactions to be put into the mix as well as vertices.

### Benefits

Users will be able to specify that a field in a vertex should match another field's value and then a transaction will be created.

### Possible Drawbacks

None foreseen (as the user controls what gets linked in)

### Verification Process

Have implemented it twice actually and have checked it against a few different data sets that I can think of (floor plans and a made up manager-employee structure)

### Applicable Issues

Nil known.


arcturus2:
@CrucisGamma I merged this in for you given you have been testing this also and its still in experimental.

spjspj:
> @CrucisGamma I merged this in for you given you have been testing this also and its still in experimental.
Thanks @arcturus2!  Have been meaning to put in the pull request for ages but didn't get around to it.

PR #798, opened 2020-08-20
Unfocus Plugin Swing Dialog Ok Button v2

antares1470:
### Description of the Change

When a plugin is run interactively, checks whether a multi-line string parameter exists and unfocuses ok button of swing dialog if so.

### Alternate Designs
#762 

### Benefits

Allows users to hit enter when typing things into a multi-line string parameter via the swing dialog box 

### Applicable Issues

#757

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Constellation Changes
 
+## 2020-08-01 Changes in August 2020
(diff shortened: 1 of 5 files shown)


aldebaran30701:
I like this change better than #762.

arcturus2 (approved):
I'd prefer this over #762

cygnus-x-1:
I think this implementation is much neater, and should be easy to update to the more flexible solution we previously discussed if we decide to go down that route.

antares1470:
Based on feedback, I'm going to keep this one and close #762

Issue #713, opened 2020-07-16
Tool Tips should reflect button states


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Toggling certain actions does not change the tool tip to explain which mode the button is in.
the noticeable ones are: (all in the vertical graph toolbar)
* Toggle draw/selection mode
* Toggle draw directed transactions
* 2D/3D modes
* Label displays
* Blaze visibility
* Graph threshold
Ideally each of these will switch tool tips to better represent the button's current state.

### Steps to Reproduce

1. Select draw mode
2. deselect draw mode

**Expected behaviour:** tool tip should change to show state

**Actual behaviour:** tool tip is the same for both states.

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

PR #799, opened 2020-08-20
Tool Tips should reflect button states

formalhaut69:
### Description of the Change
Added more short descriptions that will inform the user of the current state these buttons are in.
newToggleDescriptions

### Applicable Issues

https://github.com/constellation-app/constellation/issues/713


formalhaut69:
Let me know if you guys think there's anymore buttons that fall into this category that require this change.

Issue #794, opened 2020-08-19
Scatter plot does not update


Nova-2119:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The scatterplot view plot does not update in a large proportion of instances. Example when changing one of the axis. The user must close the view and then re-open it for the change to take effect.

### Steps to Reproduce

1. Open new Graph
2. Create Sphere Graph
3. Open ScatterPlot View
4. Change the X axis variable to X
5. Change the Y axis variable to Y
---- The below proves that changes also don't change the graph once a graph is visible
[Optional 1]  Close ScatterPlot View
[Optional 2] Open ScatterPlot View
[Optional 3] Change X or Y axis variable

**Expected behaviour:** [What you expect to happen]
The scatterplot should generate when the Y variable is selected. After it has generated the scatterplot should regenerate when options are changed

**Actual behaviour:** [What actually happens]
User must close and open view to generate initial plot and before seeing any changes to the plot.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

Nova-2119:
The scatter plot can be made to work by performing the following steps.
1. Open new graph
2. Create spher graph
3.Open scatterplot view
4. Set axis to X and Y attrbutes
5. Open map view
6. Close scatterplot view
7. Open scatterplot view

After doing the above the scatterplot should update in real time.

Nova-2119:
Okay so my description of the problem wasn't quite accurate/detailed enough and I thin I have a fix.

The problem looks like it is specific to when the scatterPlot is opened after a graph has been opened. If you open a graph after the scatterPlot has opened it will act normally.
[...]

PR #802, opened 2020-08-23
:bug: Delete handleNewGraph from ComponentShowing

Nova-2119:
Previously when the Scatterplot view was opened no listener would be
created because componentShowing would update the currentGraph prior to
the code that check if the currentGraph has changed and creates a
listener if it has. This led to the scatterPlot view being 
unresponsive.

### Benefits

Scatterplot updates as expected

### Possible Drawbacks

As of yet unseen consequence from code removal

### Applicable Issues

https://github.com/constellation-app/constellation/issues/794


aldebaran30701 (approved):
This seems to be an oversight in my testing from stopping refreshes on views which are minimized.
Good spotting!

arcturus2 (changes requested):
The issue with removing the `onShowing` is that when the view is docked and you hover over it, it no longer picks up the graph that you are on. See the gif.

pr 802 issue in docked mode

Secondly @Nova-2119 can you try to sort out your LF with git settings as files your modifying come up as all lines being changed. Thanks!

Nova-2119:
@arcturus2 apologies, I only just saw your comment. I will get onto this next week.

Nova-2119:
> 
> 
> The issue with removing the `onShowing` is that when the view is docked and you hover over it, it no longer picks up the graph that you are on. See the gif.
> 
> pr 802 issue in docked mode
> 
> Secondly @Nova-2119 can you try to sort out your LF with git settings as files your modifying come up as all lines being changed. Thanks!

Awesome, so Im fairly confident I have fixed this. I added the onShowing method back in and got it to trigger a refresh rather than handle new graph. This way it updates in the scenario above but doesn't update the activeGraph (the original problem). 
When you open a graph and then open the scatterplot it will perform a refresh with a null graph (which refresh is fine with) before then updating the active graph and refreshing the plot again. The first refresh is therefore redundant, but I don't see a way around it, we need that refresh for the scenario above, and we have no way of distinguishing the two scenarios.

When the view is docked and you switch graphs ListeningTopComponent.newActiveGraph() is called which will update ScatterPlotTopComponent.currentGraph but wont refresh the graph. This means that everything is setup and you just need to refresh the view when the view is shown. Previously we called handleNewGraph() which was overkill because that is already called when the graph is switched, its also what caused the original issue by updating currentGraph prior to calling ListeningTopComponent.newActiveGraph() and hence stopping listeners being created.

arcturus2 (approved):
Nice work @Nova-2119

antares1470 (approved):

PR #807, opened 2020-08-24
Resolved saved layers exception

formalhaut69:
### Description of the Change
The exception occurs because the LAYER_PREFERENCES attribute hadn't been initialised yet in the recalculateLayerVisibilities method.

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/StoreGraph.java
@@ -1971,7 +1971,7 @@ public class StoreGraph extends LockingTarget implements GraphWriteMethods, Seri
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


antares1470 (approved):

arcturus2:
@formalhaut69 with this change, if you open a graph without any layers (and perhaps have the layers view closed) is this going to add the layers preference attribute to the graph regardless?

aldebaran30701:
@arcturus2  So I'm thinking it won't - I've done a little code tracing and found that it initially gets called here 
```
if (attribute == layerMaskSelectedAttributeId && layerMaskSelectedAttributeId != Graph.NOT_FOUND) {
```
Which is only calling the `updateAllLayerMasks()` when the layer mask attribute is changed. 

Did you manage to create the attribute on a new graph @arcturus2 - I was also under the same assumption as you until I traced the code.

Issue #805, opened 2020-08-24
Attribute scripts on imported data will cause the importer to hang.


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
This was first found through the PR introducing the delimited importer into the view framework.
Add a csv file to the import window, drag in an attribute to a column and right click -> script -> enter an incorrect script.
No error is raised, the application just hangs. 

### Steps to Reproduce

1. open importer
2. add a csv file
3. drag an attribute to a column
4. right click the attribute
5. select script
6. enter jibberish to break the script.
7. click ok.

**Expected behaviour:** error dialog and cancelling of script.

**Actual behaviour:** no error raised, application hangs in attempt to re- try?

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

formalhaut69:
Followed the instructions to reproduce, consty didn't hang but it did throw this exception when trying to import with a broken script.

java.lang.IllegalArgumentException: Children: duplicate children added: parent = RunPane@15fb83af
	at javafx.graphics/javafx.scene.Parent$3.onProposedChange(Parent.java:558)
	at javafx.base/com.sun.javafx.collections.VetoableListDecorator.add(VetoableListDecorator.java:206)
	at au.gov.asd.tac.constellation.plugins.importexport.delimited.AttributeList.lambda$createAttribute$0(AttributeList.java:102)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
	at javafx.base/com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
[...]

PR #813, opened 2020-08-26
bugfix/Attribute scripts on imported data will cause the importer to hang

formalhaut69:
### Description of the Change
The importer would recognise if there was a syntax error in the script but would still pass it through so an exception would eventually occur when it would try to run the broken script. I've made a change so that a error popup will appear with the error message and set the script back to it's default value. 

### Applicable Issues

https://github.com/constellation-app/constellation/issues/805


antares1470 (approved):

aldebaran30701 (approved):
For myself it loaded the dialog above the graph but below the import dialog. This should be fine if the importer PR  which changes it to a view is combined with it.

arcturus2 (approved):

Issue #816, opened 2020-08-26
Layer by time bug


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

I get the following exception:

```java
Layer by Time: Action failed: au.gov.asd.tac.constellation.graph.locking.DualGraph@d0787c1; Layer by Time
Index -1107 out of bounds for length 128
WARNING [au.gov.asd.tac.constellation.plugins.templates.SimpleReadPlugin]: Action failed: au.gov.asd.tac.constellation.graph.locking.DualGraph@d0787c1; Layer by Time
java.lang.ArrayIndexOutOfBoundsException: Index -1107 out of bounds for length 128
	at au.gov.asd.tac.constellation.graph.StoreGraph.setIntValue(StoreGraph.java:2324)
	at au.gov.asd.tac.constellation.plugins.arrangements.time.LayerByTimePlugin.buildBins(LayerByTimePlugin.java:648)
	at au.gov.asd.tac.constellation.plugins.arrangements.time.LayerByTimePlugin.read(LayerByTimePlugin.java:377)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleReadPlugin.run(SimpleReadPlugin.java:81)
	at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.lambda$executePluginLater$0(DefaultPluginEnvironment.java:129)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to Reproduce

1. Build Sphere Graph
2. Arrange -> Layer by time
3. Select the date time attribute to Datetime 
4. Set Layer date-times by to Bin
5. Click OK

[...]

PR #817, opened 2020-08-26
layer by time add attribute

aldebaran30701:
### Description of the Change
This change adds an attribute which was mistakenly removed previously.

### Applicable Issues
#816


arcturus2 (approved):
Thank you for your assistance to fix this so quickly @aldebaran30701 ❤️

Issue #815, opened 2020-08-26
Scatterplot selected log out of memory exemption


Nova-2119:
### Description

An __ error is created whenever you try to have a scatterplot with a logarithmic axis display only selected nodes when no nodes are selected on the graph.

### Steps to Reproduce

1. Open scatterplot view
2. Open new graph
3. in Scatterplot view select: x, y, selected, log scale x axis.

**Expected behaviour:** [What you expect to happen]
An empty plot should be shown, behaviour should be analogous to what a happens when "og scale x axis" is not selected.

**Actual behaviour:** [What actually happens]
An out of memory exception is thrown.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

PR #820, opened 2020-08-27
:bug: Alternate minimal value for empty graph in logarithmic axis.

Nova-2119:
Add an if statement which will change the default minimum value from 0
to 1 if the logarithmic axis is called for an empty graph.

### Benefits

Stops an out of memory exception from occuring when a logarithmic axis is selected for a plot with no nodes.

### Possible Drawbacks

Unknown

### Applicable Issues

https://github.com/constellation-app/constellation/issues/815


antares1470 (approved):

arcturus2 (approved):
Nice.

PR #827, opened 2020-09-01
HOTFIX: altered duplicate transaction identifiers from split nodes

antares1470:
### Description of the Change

The new split nodes functionality duplicated transactions attached to the split node for all newly created nodes but the identifier of these transactions wasn't the original identifier meaning that if 2 nodes merged (because they had the same identifier) the transactions attached to these nodes (identical in every way bar the identifier) wouldn't merge

### Applicable Issues

#743


arcturus2 (approved):

Auriga2 (commented):
I did this intentionally, to make the new transaction identifier unique.  If that's not needed, we can just remove this if block, and used the same code in the else block to copy the identifier as well.

Auriga2 (changes requested):
I did this intentionally, to make the new transaction identifier unique. If that's not needed, we can just remove this if block, and used the same code in the else block to copy the identifier as well.

antares1470:
> I did this intentionally, to make the new transaction identifier unique. If that's not needed, we can just remove this if block, and used the same code in the else block to copy the identifier as well.

Good point, will do that. When cloning transactions, we ideally want to retain all the attribute values we can which includes identifier. This becomes especially important if a user has given a meaningful name to a transaction.

Auriga2 (approved):

Issue #822, opened 2020-08-28
Improve the split logic in the Split Nodes plugin


Auriga2:
### Description

Current split logic is inconsistent, for instance, when the `Split Characters` present consecutively in the middle or at the beginning, it duplicates the selected node in addition to the other new nodes splitted from the rest of the identifier. If the `Split Characters` present consecutively in the end, it just split the rest of the identifier.

This is the proposed criteria:

Split on first occurrence: (ALL_OCCURRENCES = False)
,,, split by , => create one node with ,,
vvv split by v = create one node with vv
v,,v split by , = create 2 nodes (v and ,v) as between ,, is no identifier
,,v split by , = create 1 node (,v) as nothing between the first two ,,s
v,,  split by , = create 2 nodes (v and ,) 

Split on all occurrences:: (ALL_OCCURRENCES = True)
,,, split by , = not create anything because of nothing between delimiters.
vvv split by v = the same as above.
v,,v split by , = create 2 nodes (v and v) as between ,, is no identifier
,,v split by , = create 1 node (v) as nothing between the first two ,,s
v,,  split by , = create 1 node (v) 

### Steps to Reproduce
Example scenario:
Create below nodes and split `a;;b` node with `;` in the `Split Characters` and with `Split on All Occurrence` set to true.
![image](https://user-images.githubusercontent.com/64826928/91514982-43e07000-e927-11ea-9f5c-c94a3301f0b7.png)

**Expected behaviour:** A new node with the same identifier `a;;b` should not be created among the other new nodes.

**Actual behaviour:**  A new node `a;;b` is created

**Reproduces how often:** 100%

Auriga2:
`v,,v split by , = create 2 nodes (v and v) `
This should produce 2 nodes with same identifier, so only one is added in the graph, with a transaction to itself. In the unit test it appears to add a new node with the same identifier. Is this behaviour expected?

antares1470:
> `v,,v split by , = create 2 nodes (v and v) `
> This should produce 2 nodes with same identifier, so only one is added in the graph, with a transaction to itself. In the unit test it appears to add a new node with the same identifier. Is this behaviour expected?

[...]

PR #828, opened 2020-09-01
Improve the split logic in the Split Nodes plugin

Auriga2:
### Description of the Change
Current split logic is inconsistent, for instance, when the Split Characters present consecutively in the middle or at the beginning, it duplicates the selected node in addition to the other new nodes splitted from the rest of the identifier. If the Split Characters present consecutively in the end, it just split the rest of the identifier.

### Alternate Designs
N/A

Better UX

### Benefits
Better UX

### Possible Drawbacks
N/A

### Verification Process
This is the proposed criteria:

Split on first occurrence: (ALL_OCCURRENCES = False)
,,, split by , => create one node with ,,
vvv split by v = create one node with vv
v,,v split by , = create 2 nodes (v and ,v) as between ,, is no identifier
,,v split by , = create 1 node (,v) as nothing between the first two ,,s
v,, split by , = create 2 nodes (v and ,)

Split on all occurrences:: (ALL_OCCURRENCES = True)
,,, split by , = not create anything because of nothing between delimiters.
vvv split by v = the same as above.
v,,v split by , = create 2 nodes (v and v) as between ,, is no identifier
,,v split by , = create 1 node (v) as nothing between the first two ,,s
v,, split by , = create 1 node (v)

### Applicable Issues
#822


antares1470 (approved):

mimosa2:
@Auriga2 I have tested this one and looks good. The split identifier logic works as per the proposed criteria. 
Only one thing I noticed was the Raw attribute also get splitted which doesn't happen in the current application. I am not sure if this is an expected change.

Auriga2:
@mimosa2 From what I can see, the `Raw` and even the `Label` field changes when  we change the Identifier, similar to V2.0.0
Great if you can attach a screen capture if they do not behave the same way for you.
image

mimosa2:
Ok. Thanks @Auriga2. I will check this out again.

arcturus2 (changes requested):
@Auriga2 it looks like the build failed because the test `testQueryWithNoResultingNodesToSplitTo_AllOccurancesNotSelected_WithoutDuplicateTransactions` failed. Please check when you can.

mimosa2:
@Auriga2, you are right, the Constellation - v2..0.0 version behaves the same way as you mentioned. Sorry I must have tested a different version on our side.

Auriga2:
> @Auriga2 it looks like the build failed because the test `testQueryWithNoResultingNodesToSplitTo_AllOccurancesNotSelected_WithoutDuplicateTransactions` failed. Please check when you can.

Fixed

arcturus2:
Awesome work on the unit testing @Auriga2, loving it!

arcturus2 (approved):

Issue #189, opened 2020-01-08
Copy to new graph default name issue


aldebaran30701:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Creating a new graph, and selecting "Copy to New Graph" will create a new graph successfully, but rename the graph as the next graph number appended. This does not appear to be intuitive that it is a copy.
![image](https://user-images.githubusercontent.com/58677759/72025279-8b968680-32c7-11ea-9431-373f1eef1279.png)

### Steps to Reproduce

1. Open Constellation
2. Open new graph
3. Copy to new graph

**Expected behaviour:** 
New graph is opened and named incrementally.
analyticgraph1 is copied to new graph named analyticgraph2

**Actual behaviour:** 
New graph is opened and new graph number is appended.
analyticgraph1 is copied to new graph named analyticgraph12

**Reproduces how often:** 
100%

### Additional Information

N/A

formalhaut69:
What we could do here is add " - copy" in the same way that windows does instead of the additional numbers. Although, it seems that the numbers are being added somewhere outside of the CopyToNewGraphPlugin file so if this enhancement was made, it could cause issues elsewhere.

formalhaut69:
This is the code block responsible for adding the numbers to the end of the graph names. Changing it could result in issues elsewhere, any suggestions? 
https://github.com/constellation-app/constellation/blob/c968184e866a76049d50df69b27b6304f7b739b5/CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java#L62-L75

aldebaran30701:
Ah yes, that looks like trouble to fix, considering it is utilized for other graph creation methods. Perhaps that is why the previous implementation was structured this way.

arcturus2:
@aldebaran30701 I'm probably used to this feature how it is so I will sound biased when I say I don't think there is a problem with how it currently works 😆. I've also not had any complaints from users about this naming convention so not too keen changing it unless there is feedback from users you talk to or feedback from UX folk.

[...]

PR #832, opened 2020-09-01
Bugfix/issue189 copy to new graph

GCHQDeveloper601:
# Description of the Change
Bugfix for issue #189 
Added logic as described in the ticket and some unit tests to cover the changes

### Alternate Designs
Design discussed in ticket #189 

Bugfix

### Possible Drawbacks

Have not added code to test for long filenames and the prompt the user may encounter.

### Verification Process
Added unit test

### Applicable Issues
#189


CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
```diff
@@ -29,11 +36,28 @@
  */
 public class GraphObjectUtilities {
 
+    private static final GraphObjectUtilities graphObjectUtilities = new GraphObjectUtilities();
```
antares1470, line 39: Based on what I could see from the usage, I think it would be better to make relevant functions and variables being called static rather than creating a static instance of the class within itself (doing a bit of testing, it seems to work the same).

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
```diff
@@ -61,15 +85,54 @@
...
+
+    private String getNewFileName(final String name, final boolean numbered, final FileObject root) {
+        final List<FileObject> files = Arrays.stream(root.getChildren()).filter(file -> file.getName().equals(name)).collect(Collectors.toList());
+        if (files.size() > 0) {
+            if (name.endsWith(COPY_STRING)) {
+                return String.format("%s (%d)", name, 1);
+            }
+
+            final Matcher matcher = Pattern.compile(COPY_STRING_PATTERN).matcher(name);
+            //final Matcher matcher =  COPY_NAME_MATCHER.matcher(name);
```
antares1470, line 121: if you move the instantiation of `COPY_NAME_MATCHER` out of the constructor, you should be able to uncomment this line and remove the line above it

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
```diff
@@ -61,15 +85,54 @@
...
+            return String.format("%s - Copy", name);
+        }
+        return numbered ? String.format("%s%d", name, ++fileCounter) : String.format("%s", name);
+    }
+
+    private boolean isFileNameDuplicateInMemory(final String name, final FileObject root) {
+
+        final String tempName = FilenameUtils.getBaseName(name);
+        final List<FileObject> files = Arrays.stream(root.getChildren()).filter(file -> file.getName().equals(tempName)).collect(Collectors.toList());
+        return (files.size() > 0);
```
antares1470, line 136: Should be able to remove the brackets from here

antares1470 (changes requested):

aldebaran30701 (commented):
It seems to handle the currently logic great, but doesn't seem correct to me.
I'm not sure on others' opinions on this..
It seems a little confusing to myself.

analyticgraph3 is the graph we start with.
We make a copy of **analyticgraph3**  named: **analyticgraph3 - Copy** - Correct
We make a copy of **analyticgraph3**  named: **analyticgraph3 - Copy (1)** - Correct
We make a copy of **analyticgraph3 - Copy** named: **analyticgraph3 - Copy (2)** - Doesn't seem right?

In that situation - I feel it may be best to have it named: **analyticgraph3 - Copy - Copy**
Simply because it is a copy of a copy. The naming lends itself to it being copy number 2 of **analyticgraph3** 

Please sanity check me if this shouldn't be the way it is looked at - Or if there is a particular reason it is done this way.

antares1470:
> It seems to handle the currently logic great, but doesn't seem correct to me.
> I'm not sure on others' opinions on this..
> It seems a little confusing to myself.
> 
> analyticgraph3 is the graph we start with.
> We make a copy of **analyticgraph3** named: **analyticgraph3 - Copy** - Correct
> We make a copy of **analyticgraph3** named: **analyticgraph3 - Copy (1)** - Correct
> We make a copy of **analyticgraph3 - Copy** named: **analyticgraph3 - Copy (2)** - Doesn't seem right?
> 
> In that situation - I feel it may be best to have it named: **analyticgraph3 - Copy - Copy**
> Simply because it is a copy of a copy. The naming lends itself to it being copy number 2 of **analyticgraph3**
> 
> Please sanity check me if this shouldn't be the way it is looked at - Or if there is a particular reason it is done this way.

Doing a bit of testing in File Explorer, what you have said @aldebaran30701 appears to be correct. Whether the new name is **analyticgraph3 - Copy (2)** or **analyticgraph3 - Copy - Copy** depends of whether we are making a copy of **analyticgraph3** (the former) or **analyticgraph3 - Copy** (the latter).

There is one thing worth noting though which I just thought of is that when copying over to a new graph. Adding **- Copy** to the graph name implies an exact copy of graph being made (i.e. the entire graph is being copied over) when actually, only a portion of the graph may be copied over so this naming convention may confuse users in that sense.

aldebaran30701:
>There is one thing worth noting though which I just thought of is that when copying over to a new graph. Adding - Copy to the graph name implies an exact copy of graph being made (i.e. the entire graph is being copied over) when actually, only a portion of the graph may be copied over so this naming convention may confuse users in that sense.

 Would that then lend it to having the name like: **subset of analyticgraph3** When the graph has been created through "copy selection to new graph" ?

@GCHQDeveloper601 How difficult do you think something like this would be to implement if we took this path?

GCHQDeveloper601:
> > There is one thing worth noting though which I just thought of is that when copying over to a new graph. Adding - Copy to the graph name implies an exact copy of graph being made (i.e. the entire graph is being copied over) when actually, only a portion of the graph may be copied over so this naming convention may confuse users in that sense.
> 
> Would that then lend it to having the name like: **subset of analyticgraph3** When the graph has been created through "copy selection to new graph" ?
> 
> @GCHQDeveloper601 How difficult do you think something like this would be to implement if we took this path?

This would be possible but would need to implement a different approach. This attempt was to use the existing method and base the naming on if there were other graphs with the same name. Anything more complicated would probably best be done with a new method, one that is context specific and let the function call the method that is needed. For instance a blank new graph would not have an issue with the existing method but having a "create a copy/subset name" method for subselection or duplication. 

Alternatively, the question of naming could be left to the context of where it is being called. With this approach the copy to new graph plugin should be responsible for creating a name based on the parent graph and that the child is a subset. Any other plugins which name a graph would be unaffected unless they needed to add specificity of context to the name.

This was an attempt of trying to build the name with the context available and keeping the changes within the one file. I'd be happy to take a wider look at how to implement this by expanding the scope.

aldebaran30701:
>This was an attempt of trying to build the name with the context available and keeping the changes within the one file. I'd be happy to take a wider look at how to implement this by expanding the scope.

It does seem to be moving out of scope, although I feel that if we addressed the wider issue it would do a better job as a whole. I'd have to agree about waiting and getting the users in on some discussion. It could be completely fine to the users without having a custom name for the copy selection to new graph.

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
```diff
@@ -29,11 +36,28 @@
  */
 public class GraphObjectUtilities {
 
+    private static final GraphObjectUtilities graphObjectUtilities = new GraphObjectUtilities();
+
+    private GraphObjectUtilities() {
+        //Adding private to remove code smell
+        COPY_NAME_MATCHER = Pattern.compile(COPY_STRING_PATTERN);
```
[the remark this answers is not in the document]
GCHQDeveloper601: I changed the COPY_NAME_MATCHER to static and removed the now empty constructor

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
GCHQDeveloper601: Changed the other methods to static so this isn't needed.

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
GCHQDeveloper601: My bad, I already had it compiled. Bad debug left in the code. Removed :)

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
GCHQDeveloper601: Done

GCHQDeveloper601:
I have pushed the changes related to the comments just to make sure I remembered to do them.

CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/GraphObjectUtilities.java
```diff
@@ -129,10 +121,10 @@ private String getNewFileName(final String name, final boolean numbered, final F
         return numbered ? String.format("%s%d", name, ++fileCounter) : String.format("%s", name);
     }
 
-    private boolean isFileNameDuplicateInMemory(final String name, final FileObject root) {
+    private static boolean isFileNameDuplicateInMemory(final String name, final FileObject root) {
 
         final String tempName = FilenameUtils.getBaseName(name);
         final List<FileObject> files = Arrays.stream(root.getChildren()).filter(file -> file.getName().equals(tempName)).collect(Collectors.toList());
-        return (files.size() > 0);
+        return files.isEmpty();
```
antares1470, line 128: will want `!files.isEmpty()` here

antares1470 (changes requested):

antares1470:
> @aldebaran30701, @GCHQDeveloper601 and @antares1470 I just wanted to let you know that it is amazing to see so much energy and thought being put into getting this right. One thing I will also say though is that we haven't really shared this new naming convention with the user community to get a thumbs up from them for this. If you have run this past some regular users and they are happy then that is good; however if this idea hasn't been socialised I'd like to do so and see what they say given this is a change in functionality they are otherwise used to for years.

We've chatted to some users about this and the reception has been positive so I'm happy for this work to continue.

arcturus2:
During testing I managed to lock up Constellation when it created the 3rd graph but I was not able to reproduce it. Something I just wanted to mention. The only thing I was doing was creating new graphs. Has anyone else seen this during testing?

antares1470 (approved):

arcturus2 (approved):

Issue #585, opened 2020-06-17
Schema View filter text invisible when out of focus


Auriga2:
When you go out of focus from the input text field in the filter, what you typed is not shown until you give focus back to the input text field. Same behaviour on the filters on Attributes, Node Types and Transaction Types tabs.

![testing schema view search](https://user-images.githubusercontent.com/39325530/84681771-fbe0bc80-af77-11ea-93e0-7bf3a9290f95.gif)

arcturus2: My guess is the bug is with the style sheet (css) file which is a pain to debug ...

aldebaran30701:
I can't recreate this in the current master running on debug mode locally. 
I can however recreate it with V2 Beta 5 running from the .EXE
Something to look out for to see if it works in the next release or even try to build a .EXE locally using current master to test.

Auriga2:
I could recreate in debug mode when I created the issue, but it worked abruptly then. Maybe it's fixed by a commit since?

formalhaut69:
@arcturus2. The fix I made in the pull request https://github.com/constellation-app/constellation/pull/627 appears to only work on my machine. Strange issue. 
![schemaViewFix](https://user-images.githubusercontent.com/58677762/86673947-503b0180-c037-11ea-8904-3dbcf06f383d.gif)

Auriga2:
This is really a  weird issue,  because after a while the bug disappears (it stops turning black) when debugging. I had to clean and build again to recreate the issue. Otherwise it'll pretend it's fixed, until you rebuild. Also the issue happens only when you click on the graph area and `treeView` section, or another view. It doesn't happen if you click on other areas and even the `detailsView` section in Node Types and Transaction Types tabs.
Using Bootstrap doesn't apply the css changes.

arcturus2:
> Using Bootstrap doesn't apply the css changes.

Using the bootstrap saves time running Constellation. You still need to compile the changes but you should be able to set the main project to the bootsrap (Run -> Set as main project) and be able to use F6 to launch it for testing. Let me know if that is not the case.

Auriga2:
Ah yes that part works although I cannot use F6 other times when I need to "Apply code changes". (Not really needed here.) May be there's another short cut key for that?

Auriga2:
I did a workaround until someone figures this out, by making the filter text black and filter background grey, so that when the text is changed to black (in the mysterious place), it' not visible to the user. 
[...]

PR #836, opened 2020-09-03
Schema View filter text invisible when out of focus

Auriga2:
### Description of the Change
When you go out of focus from the input text field in the filter, what you typed is invisible in the black background, until you click back in the filter text field. A temporary fix here by changing the filter background to grey and filter text to black so that it's always visible.
Same behaviour on the filters on Attributes, Node Types and Transaction Types tabs.

### Alternate Designs
- Find the line of code which sets the text colour to black and set it to white or a similar colour. 
- Set a priority style other than using `!important` which didn't work

Better UX

### Benefits
Filter text no longer go invisible.

### Possible Drawbacks
N/A

### Verification Process
- Open a graph and open the Schema View. 
- Type something in the `Attributes` tab and click in the graph section. The filter text colour or the background colour should not change.
- Repeat with the filters in `Node Types` and `Transaction Types` tabs.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/585

### DIFF
--- a/CoreSchemaView/src/au/gov/asd/tac/constellation/views/schemaview/resources/schema-view.css
+++ b/CoreSchemaView/src/au/gov/asd/tac/constellation/views/schemaview/resources/schema-view.css
@@ -2,3 +2,8 @@
 .schemaview-highlight-blue {
     -fx-background-color: rgb(30,144,255); -fx-padding: 2 5 2 5;
 }
+
+.text-field{    
+    -fx-text-inner-color: black;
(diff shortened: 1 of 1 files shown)


antares1470 (approved):

arcturus2 (approved):

arcturus2:
Thanks for all the testing and persistence with this @Auriga2 and others that have tried this. I created https://github.com/constellation-app/constellation/issues/841 to look at this problem sometime later.

Issue #830, opened 2020-09-01
Make all nodes visible in the layer by time plugin


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The layer by time plugin makes some nodes invisible in the graph it generates. The reason it does this is to pre-compute the graph into layers so you can use it with the "Dim unselected" feature in the Timeline view. However having all nodes visible does not impact this feature.

Having invisible nodes on the graph is confusing so we should remove this logic throughout the plugin.

### Steps to Reproduce

1. Create a Sphere graph
2. Layer by time

**Expected behaviour:** [What you expect to happen]
All nodes should be visible

**Actual behaviour:** [What actually happens]
Some nodes are invisible

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Additional Information

If you remove LayerByTimePlugin.java:141 (see line below) and then work through cleaning up the plugin that might just be all that is required.

```
private static final String VISIBILITY = "visibility";
```

PR #844, opened 2020-09-14
Bugfix for Layer By Time Visibility 

Polaris5514:
### Description of the Change
Visibility of Duplicate nodes was being set to 0, changed to 1

### Alternate Designs
N/A

Bugfix

### Benefits
All nodes Visible 

### Possible Drawbacks
None realised

### Applicable Issues
#830

### DIFF
--- a/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/time/LayerByTimePlugin.java
+++ b/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/time/LayerByTimePlugin.java
@@ -717,13 +717,13 @@ public class LayerByTimePlugin extends SimpleReadPlugin {
         graph.setFloatValue(x2Attr, sNodeId, graph.getFloatValue(xAttr, sNodeId));
         graph.setFloatValue(y2Attr, sNodeId, graph.getFloatValue(yAttr, sNodeId));
         graph.setFloatValue(z2Attr, sNodeId, graph.getFloatValue(zAttr, sNodeId));
-        graph.setFloatValue(vxVisibilityAttr, sNodeId, 0);
+        graph.setFloatValue(vxVisibilityAttr, sNodeId, 1);
 
         final int dNodeId = graph.getTransactionDestinationVertex(txId);
         graph.setFloatValue(x2Attr, dNodeId, graph.getFloatValue(xAttr, dNodeId));
         graph.setFloatValue(y2Attr, dNodeId, graph.getFloatValue(yAttr, dNodeId));
         graph.setFloatValue(z2Attr, dNodeId, graph.getFloatValue(zAttr, dNodeId));
-        graph.setFloatValue(vxVisibilityAttr, dNodeId, 0);
+        graph.setFloatValue(vxVisibilityAttr, dNodeId, 1);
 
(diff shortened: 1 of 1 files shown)


arcturus2 (changes requested):
@Polaris5514 thanks for this and what you have proposed will work. However I was also thinking about just not setting the `vxVisibilityAttr` attribute because it does default to `1`. This would mean that we could delete the definition of `vxVisibilityAttr` and any references to it. If you have tried this already and if that wasn't working then let me know and I'll be happy to approve this.

Polaris5514:
@arcturus2 Apologies, I had attempted this yesterday but Node 0 was still invisible. Have now found that Node 0 is set to invisible on creation of the sphere graph, not within the Layer By Time plugin. 
Requested changes have now been properly implemented

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/time/LayerByTimePlugin.java
```diff
@@ -441,7 +437,6 @@ public void read(final GraphReadMethods rg, final PluginInteraction interaction,
                 //Tag on the back of this loop and update the visibility of each node to match it's z value
                 if (isTransactionLayers) {
                     float vxZ = wgcopy.getFloatValue(zAttr, vertexId);
-                    wgcopy.setFloatValue(vxVisibilityAttr, vertexId, vxZ / (z + 1));
                 }
```
arcturus2, line 440: @Polaris5514 from my reading of the code I think this entire if block could be removed. What do you think?

arcturus2 (changes requested):
All good @Polaris5514, you have nothing to apologies for. I think I've found a bit more code that could be deleted so let me know when you have a chance to review.

The advantage of removing code is that it potentially can uncover additional redundant code and takes a step towards removing technical debt which is a good thing 😄

Polaris5514:
Thanks @arcturus2, I have removed the suggested if block. Have also looked through and found two more redundant sections of code and removed them as well.

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/time/LayerByTimePlugin.java
```diff
@@ -356,8 +355,6 @@ public void read(final GraphReadMethods rg, final PluginInteraction interaction,
             final int timeLayerAttr = wgcopy.addAttribute(GraphElementType.TRANSACTION, LayerNameAttributeDescription.ATTRIBUTE_NAME, LAYER_NAME, "time layer", defaultName, null);
             wgcopy.addAttribute(GraphElementType.GRAPH, IntegerAttributeDescription.ATTRIBUTE_NAME, NLAYERS, "The number of layers to layer by time", 1, null);
             final int zAttr = wgcopy.addAttribute(GraphElementType.VERTEX, FloatAttributeDescription.ATTRIBUTE_NAME, "z", "z", 0, null);
```
antares1470, line 357: this variable is apparently no longer being used. If you're able to remove it that would be great.

antares1470 (changes requested):

antares1470 (approved):

Issue #712, opened 2020-07-16
Constellation - Tutorial Page/Quick Start Guide - Font to small


WCWarrior1:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Constellation-v2.0.0-rc2 - Tutorial/Quick Start Guide - the font on the right side is way to small and is very hard to read, this needs to be made bigger. 

Check snip attached.
![Tutorial Tab and fonts example](https://user-images.githubusercontent.com/66282538/87627265-add6f880-c771-11ea-965f-604d1b733e08.jpg)

### Steps to Reproduce

1. **Not required**
2. [Second Step]
3. [and so on...]

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

arcturus2:
Thanks @WCWarrior1 will have a look. I felt it looked small too.

PR #855, opened 2020-10-16
Update to 'Whats New' Fontsize

Polaris5514:
### Description of the Change
Changed the What's New font size to a set size instead of using preferred font size

### Alternate Designs
Different Font Size

Bugfix

### Benefits
What's New text easier to read

### Possible Drawbacks
None Realised 

### Applicable Issues
#712


arcturus2 (changes requested):
Thanks for this @Polaris5514 but hard coding the font size isn't the right approach. We've taken steps to have our views dynamically update their font size when you change it from Setup -> Options -> Misc because different people will have different screen resolutions or preferences. 

Can you work with `FontUtilities.getOutputFontSize()` and see if you can maybe increase it by a factor like 1.5 to get it aligned with the rest of the views?

antares1470 (approved):

arcturus2 (approved):

aldebaran30701 (approved):

Issue #56, opened 2019-07-26
Notes View - A history of actions and states


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

I think Constellation needs a new view which simply allows users to record notes. These notes would be saved along with the graph and would help the user track their analytical workflow, as well as provide context to other users about how conclusions have been drawn from a graph. 

It would also be great to have some automated notes, such as notes indicating when important plugins are run (such as data access plugins). The view should provide a way to show and hide these automated notes at the users request.

That is, it should provide the ability to review what activities have been undertaken on items in a Constellation graph. Some graphs may be used for extended periods of time and the ability to see what queries/searches/enrichments have been done on them and when, would be useful.

For example it will help answer questions like:
 * have I run query XYZ on node ABC?
 * have I already enriched with datasource Y?
 * have I searched event or content already?

In addition, once imports and enrichments are run, the system generated note when selected should highlight the nodes and transactions that were involved with that execution.

sol695510:
Notes View has been made with ability to store user notes and generate automatic notes on plugin activation so far.

Currently there exists a bug where the automatic notes do not refresh automatically on graph unless you switch to another open graph and come back to the original graph.

See current functionality on this branch:
https://github.com/constellation-app/constellation/tree/feature/Notes-View-A-History-of-actions-and-states-%2356

arcturus2:
@sol695510 many apologies for not being able to get onto testing this for a long time. Its looking great for the POC.

![image](https://user-images.githubusercontent.com/39325530/91810383-1f5efd80-ec71-11ea-9e2d-a59c4452139b.png)

I'm not sure what notes tools you like but I'm a fan of Google Keep and like how that shows the heading when you click on it. See the gif below for some idea. This is probably something that come later though.

![keep add note](https://user-images.githubusercontent.com/39325530/91810731-bdeb5e80-ec71-11ea-92a1-c36390e92213.gif)

cygnus-x-1:
Some thoughts:
[...]

PR #877, opened 2020-10-26
Feature/notes view a history of actions and states #56

sol695510:
### Description of the Change
New view that allows users to keep a record of self written notes. Also lists "auto notes" that track plugin actions.

### Alternate Designs
N/A

It is a new view for Constellation.

### Benefits
Self written notes are unique to graphs saved respectively so if a single graph is accessed by multiple users, the notes can be useful documentation that can be easily seen over other documentation like code comments.

### Possible Drawbacks
N/A

### Verification Process
N/A

### Applicable Issues
#56


sol695510:
Further development related to this PR can be found at https://github.com/constellation-app/constellation/pull/880#issue-510516462. As such this PR is being closed.

Issue #862, opened 2020-10-21
Table view export limit


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The table view now has pagination but a side effect of this is that the export function (to CSV and Excel) only exports the results shown on that page, rather than all selected results across pages.

### Steps to Reproduce

1. Create Sphere graph with more than 500 transactions
2. Open Table view
3. Export graph to Excel.
4. Open the resulting file and observe only 500 items have been exported (for page limit = 500)

**Expected behaviour:** All desired nodes/transactions to export are exported

**Actual behaviour:** Only desired nodes/transactions to export on the current page are exported

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

PR #885, opened 2020-11-02
Table view export

antares1470:
### Description of the Change

Have fixed the issue where copying and exporting data from the table view only was applied to the current page.

Bug fix in Core

### Benefits

Users can copy and export the entire table instead of just the page they are currently on.

### Possible Drawbacks

Getting the selection variants working properly meant a cost in performance for export to excel

### Applicable Issues

#862

### DIFF
--- a/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
+++ b/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
@@ -1272,7 +1272,7 @@ public final class TableViewPane extends BorderPane {
                 if (!state.isSelectedOnly()) {
                     final List<Integer> selectedIds = new ArrayList<>();
                     final int[][] selectedIndices = new int[1][1];
-                    final Thread t = new Thread() {
+                    final Thread selectedIdsThread = new Thread("Update Selection from FX Thread: Get Selected Ids") {
                         @Override
                         public void run() {
                             final ReadableGraph readableGraph = graph.getReadableGraph();
@@ -1283,9 +1283,9 @@ public final class TableViewPane extends BorderPane {
                                     .map(row -> table.getItems().indexOf(row)).mapToInt(i -> i).toArray();
                         }
                     };
-                    t.start();
+                    selectedIdsThread.start();
                     try {
-                        t.join();
+                        selectedIdsThread.join();
                     } catch (InterruptedException ex) {
                         Thread.currentThread().interrupt();
                     }
--- a/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
+++ b/CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
@@ -1041,7 +1041,7 @@ public final class TableViewPane extends BorderPane {
             }
             
             sortedRowList.comparatorProperty().bind(table.comparatorProperty());
-            updateSelectionFromJAT(parent.getCurrentGraph(), parent.getCurrentState());
+            updateSelectionFromFXThread(parent.getCurrentGraph(), parent.getCurrentState());
             sortedRowList.comparatorProperty().addListener(tableComparatorListener);
             selectedProperty.addListener(tableSelectionListener);
         }
@@ -1260,7 +1260,7 @@ public final class TableViewPane extends BorderPane {
      * @param graph the graph to read selection from.
      * @param state the current table view state.
      */
-    private void updateSelectionFromJAT(final Graph graph, final TableViewState state) {
+    private void updateSelectionFromFXThread(final Graph graph, final TableViewState state) {
         synchronized (LOCK) {
             if (graph != null && state != null) {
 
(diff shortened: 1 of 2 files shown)


CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -1239,44 +1234,99 @@ public void updateSelection(final Graph graph, final TableViewState state) {
...
     }
+       
+    /**
+     * A version of the updateSelection(Graph, TableViewState) function which is
+     * to be run on the JavaFX Application Thread
+     * 
+     * @param graph the graph to read selection from.
+     * @param state the current table view state.
+     */
+    private void updateSelectionFromJAT(final Graph graph, final TableViewState state) {
```
arcturus2, line 1263: A real minor one but how about calling it `updateSelectionFromFxThread`?

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
```diff
@@ -1239,44 +1234,99 @@ public void updateSelection(final Graph graph, final TableViewState state) {
...
+        synchronized (LOCK) {
+            if (graph != null && state != null) {
+
+                if (!Platform.isFxApplicationThread()) {
```
arcturus2, line 1267: So if it is meant to be run on the FX thread, how is this bit passing?

antares1470: `Platform.isFxApplicationThread()` is a boolean function that returns true if we are running in the fx application thread. So the condition here returns true if we are NOT running on the fx application thread

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
antares1470: done

CoreTableView/src/au/gov/asd/tac/constellation/views/tableview2/TableViewPane.java
arcturus2: Doh! Thanks, I totally misread it.

arcturus2 (changes requested):
Hi @antares1470 I got an `IllegalStateException` trying to export a graph with just 6 transactions to csv. I got a similar error trying to export to Excel.

```
java.lang.IllegalStateException: Not processing on the JavaFX Application Thread
	at au.gov.asd.tac.constellation.views.tableview2.TableViewPane.updateSelectionFromFXThread(TableViewPane.java:1268)
	at au.gov.asd.tac.constellation.views.tableview2.TableViewPane.createPage(TableViewPane.java:1044)
	at au.gov.asd.tac.constellation.views.tableview2.TableViewPane.lambda$paginate$62(TableViewPane.java:1054)
	at au.gov.asd.tac.constellation.views.tableview2.TableViewUtilities.getTableData(TableViewUtilities.java:117)
	at au.gov.asd.tac.constellation.views.tableview2.TableViewUtilities$ExportToCsvFilePlugin.execute(TableViewUtilities.java:258)
	at au.gov.asd.tac.constellation.plugins.templates.SimplePlugin.run(SimplePlugin.java:68)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

arcturus2:
FYI the first time I started Constellation, created a sphere graph and then loaded the Table View it got stuck and looked like this:

image

I wasn't able to reproduce it though.

arcturus2 (approved):

antares1470:
> 
> 
> FYI the first time I started Constellation, created a sphere graph and then loaded the Table View it got stuck and looked like this:
> 
> image
> 
> I wasn't able to reproduce it though.

Thanks @arcturus2 for testing. I did a little digging into this and when I tested, I too encountered the issue of table view not loading correctly (table view loaded but the data was either missing or in the wrong columns). Closing and loading table view a few times produced differing results before eventually loading correctly. 

I tried replicating in master and got no success so I went back to my branch to confirm whether it was my branch and got no success replicating the first time. I did get it to replicate the next time I ran it though. 

This makes me think that the issue you've identified may be something in my code so I'll do some more digging and see if I can resolve this.

antares1470:
> Thanks @arcturus2 for testing. I did a little digging into this and when I tested, I too encountered the issue of table view not loading correctly (table view loaded but the data was either missing or in the wrong columns). Closing and loading table view a few times produced differing results before eventually loading correctly.
> 
> I tried replicating in master and got no success so I went back to my branch to confirm whether it was my branch and got no success replicating the first time. I did get it to replicate the next time I ran it though.
> 
> This makes me think that the issue you've identified may be something in my code so I'll do some more digging and see if I can resolve this.

Alright I think I have found the source of the problem and resolved it. My new update selection function had a synchronized despite the fact it would only be running on the fx thread (something I missed having initially copy pasted the original update selection function as a starting point for the new one).

PR #900, opened 2020-11-27
add help documentation for layers view

aldebaran30701:
### Description of the Change
Add help pages for layers view.
This pr adds the help set for layers view. This includes manual and query layers, adding and removing to layers. General usage of layers view.

### Alternate Designs
Waiting on the help set overhaul has proven to take longer than expected - so the idea to get at least some documentation into the project was the next step.

doco

### Benefits
doco

### Possible Drawbacks
N/A

### Verification Process
Open help pages via search


aldebaran30701:
@arcturus2 
I just did some extra testing, as well as added a help button. It seems to work fine on my machine - I did however correct a mismatch between class IDs which might have prevented the help from opening as intended.

image

arcturus2 (approved):
Thank you - works well

antares1470 (approved):

PR #903, opened 2020-12-02
updated whats new and CHANGELOG with some missing entries

antares1470:
### Description of the Change

Updated the whatsnew and CHANGELOG with some missing entries.

Related to changes made in Core

### Benefits

Fills in holes on changes made for developers and users.

### Possible Drawbacks

Nil


arcturus2 (approved):
This is awesome @antares1470. Many thanks for doing this!

Issue #902, opened 2020-12-01
Table View issues


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Some issues have been identified with the table view:
1) When in Selection only mode, Export to CSV/Excel (Selection) only exports the last selected row
2) When selecting rows in the table view, you lose the selection you have as soon as you try select another row on another page (even when holding down Ctrl key)

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** Export selections from table view as expected and selections persist across pages

**Actual behaviour:** Above issues

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

cygnus-x-1:
I've tested the pagination issues mentioned during our last catch-up, and had a general play with the Table View. I didn't find any bugs and really like how pagination has been implemented!

PR #905, opened 2020-12-04
Bugfix/table view issues

antares1470:
### Description of the Change

Fix some bugs where exporting in selected only mode didn't export properly (due to not retaining selection when switching pages) and selecting in the table deselected any selected items on other pages.

Fixes issues in table view

### Benefits

See above

### Applicable Issues

#902


mimosa2:
Tested this. Works fine as expected. The selections on different pages are retained.

arcturus2 (approved):

mimosa2 (approved):

cygnus-x-1 (approved):

Issue #932, opened 2021-01-12
Welcome Page doesn't resize properly


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

With no views open, you can see everything on the welcome page as expected. It appears that the page doesn't resize properly though when views are opened (and docked). A view opened on the bottom of the screen forces the user to scroll to see everything and a view opened (and docked) on the left hand side results in being unable to see everything properly (even with scrolling).

### Steps to Reproduce

1. Open the Welcome tab
2. Open the Attribute Editor (or any other view that when opened is docked on the left hand side)
3. Try scrolling on the welcome page to see any all of the content

**Expected behaviour:** As views are added/removed and docked/undocked, the welcome page resizes such that you are able to see all the important features on the page

**Actual behaviour:** Welcome page doesn't resize properly with an inability to see everything on the page in some instances

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

PR #933, opened 2021-01-12
Welcome Page Resize Issue

Delphinus8821:
### Description of the Change
Added a scroll pane to the welcome page so that the content on the page is able to be seen without any buttons overlapping. The contents of the page now fill to the height and width of the current page and if the page size gets smaller than the layout of the contents of the page, scroll bars are added. This allows for the page keeps its formatting and allow all buttons to continue to be viewed using the scroll bar.

### Alternate Designs
None realised

The current welcome page does not resize correctly so some of the buttons are unable to be reached. This change ensures that all of the content on the page is able to be viewed either by default or by using the scroll bars. 

### Benefits
Allows for all of the buttons on the welcome page to be reachable when the page is resized.

### Possible Drawbacks
None realised

### Verification Process
Tested by changing the size of the window and by adding additional views in constellation both to the side and bottom of the window. 

### Applicable Issues
#932 

Example of the change to the welcome page with views added to the side and bottom of the window.

ScrollingWelcomePage

### DIFF
--- a/CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
+++ b/CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
@@ -78,8 +78,7 @@ public class WelcomeViewPane extends BorderPane {
             splitPane.setOrientation(Orientation.HORIZONTAL);
             splitPane.setStyle("-fx-background-color: transparent;");
             final ScrollPane scrollPane = new ScrollPane(splitPane);
-            
-            Rectangle2D visualBounds = Screen.getPrimary().getVisualBounds();
+            final Rectangle2D visualBounds = Screen.getPrimary().getVisualBounds();
             
             splitPane.setPrefHeight(visualBounds.getHeight());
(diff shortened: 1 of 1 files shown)


CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
```diff
@@ -71,12 +73,19 @@ public WelcomeViewPane() {
...
-
+            
             final SplitPane splitPane = new SplitPane();
             splitPane.setOrientation(Orientation.HORIZONTAL);
             splitPane.setStyle("-fx-background-color: transparent;");
-            welcomeViewPane.setCenter(splitPane);
-
+            final ScrollPane scrollPane = new ScrollPane(splitPane);
+            
+            Rectangle2D visualBounds = Screen.getPrimary().getVisualBounds();
```
antares1470, line 82: final

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):

PR #759, opened 2020-08-04
remember_dropdown_list_options

mimosa2:
### Description of the Change

Remember Drop Down List Options

### Why Should This Be In Core?

This is for Data Access view

### Benefits

Users don't have to remember previous selections.

### Possible Drawbacks

Currently the number inputs are not remembered with this changed. Default values are overwritten.

### Verification Process

I have only used the Test Parameters Plugin to test my change. I have commented out the default values within the plugin.

Issue #25, opened 2019-07-26
Remember drop down list options


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a user,
I want the Data Access View to remember my selections between sessions,
so that I don't have to spend time clicking the same options every time i launch the application.

Every time Constellation is opened, the Data Access View retains the default drop down list option rather than the option set by the user. Having the last drop down list option set instead would save a lot of time.

### Steps to Reproduce

1. Open a graph
2. Open the Data Access View
3. Expand the Merge Nodes plugin and select "Supported Type"
4. Click Run
5. Close Constellation
6. Open Constellation

**Expected behaviour:** The Merge Nodes plugin's drop down list selection should be "Supported Type"

**Actual behaviour:** The drop down list resets to nothing being selected

**Reproduces how often:** 100%

mimosa2:
As I went through this I found the current setup seems to be as follows;
1. A preference file with recent values of plugin parameters are written in the user directory whenever the App is closed .
2. When opening the App the instance PluginParametersPane loads the recent values from the file via ValueInputPane and RecentParameterValues.
3. Currently the ValueInputPane loads the previous text field values in a separate dropdown and it doesn't load the recent value on the actual field.
4. Curently there is no setup for SingleCoiceInputPane and MultiChoiceInputPane to load the dropdown or checkbox parameters.

ValueInputPane
![ValueInputPane](https://user-images.githubusercontent.com/54261086/83868975-24dea180-a76f-11ea-99e7-270099382461.PNG)

RecentParameterValues
![RecentParameterValues](https://user-images.githubusercontent.com/54261086/83869477-fb724580-a76f-11ea-9232-94cfa91ed372.PNG)

I tried with SingleChoiceInputPane first to add to the LISTENERS list (same as the ValueInputPane)  but not sure that is the way to go. (see below)

SingleChoiceInputPane
![SingleChoiceInputPane](https://user-images.githubusercontent.com/54261086/83869012-33c55400-a76f-11ea-8df4-5d6d6707afe4.PNG)

[...]

PR #940, opened 2021-01-18
HOTFIX: added missing case in MultiValueInputPane

antares1470:
### Description of the Change

For Single Choice and Multi Choice parameter types, the list of choices can be Strings or they can be other parameter types. When work was done to remember Data Access plugin parameter value between sessions (#759), the work done on the single choice input pane accounted for all parameter type choices but work done on the multi choice input pane only accounted for string parameter type choices.

This change mimics in `MultiChoiceInputPane` what is done in the `SingleChoiceInputPane` for remembering parameter values

Affects Core code

### Benefits

Fixes potential bug in case where there is a multichoice parameter whose options aren't string parameters

### Applicable Issues

#25


mimosa2 (approved):
Works correctly. Good work, thanks.

Issue #896, opened 2020-11-23
Split Nodes plugin resetting attributes


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Split Nodes by Identifier plugin runs a "complete with schema" at the end. This is a problem because if I have customised my nodes with different colours etc, splitting them will reset them (because the complete with schema plugin runs).

### Steps to Reproduce

1. Change some icons and colours of nodes
2. Run the split nodes by identifier plugin on a character

**Expected behaviour:** [What you expect to happen]
Nodes should be split preserving how they look (icons and colours)

**Actual behaviour:** [What actually happens]
Nodes are split by the icons and colours have reverted

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Additional Information

Is it possible the achieve what we need to in the Split nodes plugin without running the "Complete With Schema" plugin?

Delphinus8821:
@arcturus2 @antares1470  It appears that Complete with Schema can be useful in specific scenarios 

The below examples are using the same generated sphere graph where using Complete with Schema removes the colours and icons. 
![RandomSphereGraph](https://user-images.githubusercontent.com/73083035/104539513-9e2b6480-566d-11eb-8f1e-dd60c9e03311.png)

Attribute editor view on the same node before running the plugin and after running the plugin with and without the schema plugin:
![Attribute Editor Split Nodes](https://user-images.githubusercontent.com/73083035/104530927-f443dc00-565c-11eb-83a8-2d6e9249b08e.png)

Below is the example of a sphere graph where the type of all nodes is changed to Person, so Complete with Schema updates all of the nodes to match the Person schema.
![Split Node Person Node Type Comparison](https://user-images.githubusercontent.com/73083035/104538854-94edc800-566c-11eb-9155-f0646402ca65.png)

Delphinus8821:
An option is to create a checkbox for the Split Node Plugin to allow the user to choose whether to apply the complete with schema plugin or not. 

E.g.
[...]

PR #942, opened 2021-01-19
Split Nodes Reset Attributes

Delphinus8821:
### Description of the Change
The split nodes plugin when run would change the icons and colours of nodes due to the complete with schema plugin being run by default. A checkbox has been added to the split nodes plugin to allow for the user to decide whether they wish for complete with schema to be run on the selected nodes. The selected nodes which the split nodes plugin runs on are placed into a subgraph so that if the user chooses to use the complete with schema plugin it only affects the nodes in that subgraph instead of altering the entire graph. 

### Alternate Designs
None realised

Allows for the user to make more decisions about how the graph is affected when the split nodes plugin is run. 

### Benefits
Allows for users to choose whether they want to keep the changes they have made to nodes (colours and icons) when running the plugin.

### Possible Drawbacks
None realised

### Verification Process
Tested by running the split nodes plugin alongside the other plugins in data access view. Also ran the split nodes plugin with and without the complete with schema checkbox ticked and on both selections on a graph and the full graph.

### Applicable Issues
#896


antares1470 (changes requested):
Doing some testing and it doesn't appear that my selection is completing with schema if I have it ticked. Pressing F5 afterwards confirms this.

antares1470 (approved):

arcturus2 (approved):
Awesome! Thank you for this @Delphinus8821 .

PR #947, opened 2021-01-27
Data Access Plugins bug fix

antares1470:
### Description of the Change

Fixed the issue where plugins in the data access view would be enabled upon expanding the parameters (if recent values had been applied).

The cause of the issue was a parameter listener that was being added before the parameter pane was created (which would be triggered if recent parameter values were applied to the parameter). Since the listener isn't actually needed before the pane is created, this listener has now been added after the pane is created.

Fixes a bug in core

### Benefits

Plugins aren't unexpectedly enabled

### Possible Drawbacks

Nil

### Applicable Issues

#25


arcturus2 (approved):
Thanks for this!

Issue #866, opened 2020-10-21
Fix rendering of buttons and menu options


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Menu options in views are not obvious because they don't look like something you can click on. For example refer to the "Options" menu on the top left of the Data Access View. Same thing for menu options in the Map view.

These are JavaFX CSS type fixes.

cygnus-x-1:
An example of this is the Delimited File Importer Options Menu, which contains options for saving and loading templates, which we've seen many users miss altogether.

serpens24:
The only real way to address this issue is to roll through different views and look for buttons that are not distinguishable. The examples cited by @antares1470  and @cygnus-x-1 are just two examples.

Its also worth noting that both the 'Data Access View' and 'Delimited File Import' view have an 'Options' menu at the top left that is not obvious - but its owrth noting the actual shading of these two views is different, see image below where the 'Delimited File Importer' window has a darker top bar than the Data Access View.

I don't propose to make this consistant at this point unless it falls out in ensuring the button colorations are unique.

![image](https://user-images.githubusercontent.com/58677108/105116406-a0cd0480-5b1a-11eb-8387-ce16e9910a74.png)

serpens24:
Ehile the colors below are not a reccomendation, they were used to validate the reason the Option buttons were not easily identifed - ultimately, the reason is that these buttons are menubar buttons which by default share the same look as the underlying menubar.

Hence the approach needs to be to change the colour of the .menu-button item in the css styles.

![image](https://user-images.githubusercontent.com/58677108/105131358-ad605580-5b38-11eb-8ae5-2f91cb70227e.png)

PR #948, opened 2021-01-27
Modified menu display to ensure it is visible.

serpens24:
### Description of the Change

Modified menu display settings for several views to give a consistant feel as well as shading any menu options to make the menu identifyable.

### Alternate Designs

nil

Requested ticket highlighted that in some cases this top menu wasnt easily identifed as cselectable.

### Benefits

Ensure menu is selectable - and identifiable by the user. Note - the issue in this case is that the top bar is a menubar, which normally doesn't have shading for each menu option, this however is having other buttons overlaid so it requires a slightly different look and feel.

### Possible Drawbacks

None identified

### Verification Process

Go through all views and ensure any with menus have the same look and feel - in particualr any menu (ie Options menu for Data
Access View) is shaded differently to identify it as a button.

### Applicable Issues

#866

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewPane.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewPane.java
@@ -15,6 +15,7 @@
  */
 package au.gov.asd.tac.constellation.views.analyticview;
 
+import au.gov.asd.tac.constellation.utilities.color.ConstellationColor;
 import au.gov.asd.tac.constellation.utilities.icon.UserInterfaceIconProvider;
 import au.gov.asd.tac.constellation.views.analyticview.AnalyticViewTopComponent.AnalyticController;
 import au.gov.asd.tac.constellation.views.analyticview.questions.AnalyticQuestion;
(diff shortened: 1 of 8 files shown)


antares1470 (changes requested):
Looking at the changes, I think the menu button could really benefit from having a border around it (even a thin one). I can tell that the Options buttons are a different colour, but only because I'm looking for it. I think this may be because the menu button colour itself isn't that much different from the bar it sits on (especially for the Delimited File Importer). I think the colour scheme can be kept, just needs a outline.

Also, I noticed the help button in the Delimited File Importer is still a darker colour to the one you've been applying to some of the other views. Not sure if that was intended or was simply missed but I think the colour change should be applied there also.

arcturus2 (approved):
Thanks for going through this and trying to make it more consistent. It is definitely a step in the right direction. As @antares1470 mentioned, it would be interesting to see what an outline would look like.

serpens24:
Thanks both. I had missed the delimited importer help color so will fix.

The color I used for menus was based on the button color of the other overlay buttons - ie the + and * buttons for Data Access View. These do stand out better because they have borders. I cannot replicate those border styles AFAIK, but I can add a plain rectangular border. Doing so certainly makes it more obvious .... I will experiment with colour, I suspect a light silver rather than white may be the best option.

image

antares1470 (approved):

Issue #941, opened 2021-01-19
Graph selection from the Histogram no longer works.


cygnus-x-1:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Selecting bins/bars on the Histogram no longer updates selection on the graph. This is a fundamental issue which prevents intended use of the Histogram as you can no longer find information on the graph that you've filtered down to in the Histogram.

### Steps to Reproduce

1. Open a graph and the Histogram.
2. Select an attribute in the Histogram and select a bin.

**Expected behaviour:** Graph elements represented by the selected bin should also be selected on the graph.

**Actual behaviour:** Nothing is selected on the graph (and have fun finding it in the graph via other means).

**Reproduces how often:** 100%

cygnus-x-1:
Histogram bars are drawn using Graphics2D, so this is likely unrelated to the similar Conversation View issues around graph selection.

PR #951, opened 2021-01-28
fixed histogram selection issue

antares1470:
### Description of the Change

A change was made previously to fix the issue of dragging in the histogram not working. The fix to this though introduced a new bug in which histogram selection wasn't applied to the graph. This change was reverted to fix the selection issue.

Reverting this change meant the dragging bug was again an issue. After some digging, I identified the cause of this to be a boolean condition that was introduced to `mouseDragged()` which meant the most of the code in the function was never executed (as this condition was always false). This condition was modified to correctly add the intended behaviour of introducing the condition.

Fixes a bug in core

### Benefits

Dragging and selection in the histogram now work again as intended

### Possible Drawbacks

Nil

### Applicable Issues

#941


arcturus2 (approved):
Perfect works great.

Issue #952, opened 2021-01-28
Histogram filter on selecting in Link mode


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Filtering in the histogram in **Link** mode causes a null pointer. Refer to gif for a reproducer.

![histogram filter on selection null pointer](https://user-images.githubusercontent.com/39325530/106090797-1e2df000-617f-11eb-9535-4717e7540ae7.gif)

```
java.lang.NullPointerException
	at au.gov.asd.tac.constellation.views.histogram.HistogramFilterOnSelectionPlugin.edit(HistogramFilterOnSelectionPlugin.java:69)
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:207)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:115)
Caused: java.lang.RuntimeException
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:135)
[catch] at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.lambda$executePluginLater$0(DefaultPluginEnvironment.java:129)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to Reproduce

1. Open a graph with a transaction
2. Histogram -> Link mode
3. Select an attribute (e.g. Source Node Attribute)
4. Select the bar in the histogram
5. Click on the Filter on Selection button

**Expected behaviour:** [What you expect to happen]
The histogram should show the highlighted bars

[...]

PR #960, opened 2021-02-02
Added EDGE and LINK element types to saupport filterinf by them.

serpens24:
Modified Historgram filter setup to include EDGE and LINK as potential element types in line with GUI.
This is done to ensure code is able to successfully filter by element type  rather than retuning null and triggering exceptions.

### Alternate Designs

Nil - this is a bug fix due to missing element types

Bug fix

### Benefits

Can now filter by Edge/Link Grapg Elements rather than seeing an exception

### Possible Drawbacks

none known

### Verification Process

Repeated steps shown in issue description. 
Performed filtered-by and clear opeorations for all graph element types.

### Applicable Issues

#952


antares1470 (approved):

arcturus2 (approved):
Thank you!

Issue #926, opened 2020-12-18
Analytic View Aggregation method initially has incorrect list of options to pick from


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

By default Analytic View opens up with the "Connects the Network Best?" question selected. When you first open it up, the Aggregation method drop-down is incorrectly populated with a number of options that aren't available for the question. This is corrected after another question is selected (with only one option) but really this should be correctly populated from the start.

### Steps to Reproduce

1. Open Analytic View
2. Observe Aggregation method drop-down list (unexpected list)
3. Select a different question and back again and observe only one option available in drop-down (expected list)

**Expected behaviour:** Aggregation method drop-down is populated with only one option for "Connects the Network Best?" question

**Actual behaviour:** Initially, the aggregation method drop-down is populated with other options that aren't available for the question

**Reproduces how often:** 100%

aldebaran30701:
I've done some digging into this, but have not been able to get to the bottom of it. 
What I have found is that it happens some time during the first or second selection of a different question. The most recent thing I was investigating was the UpdateParameters() method - which I had a suspicion of it clearing all of the other possible options to select.

I've decided to switch gears and take a look at another issue in the meantime.

arcturus2:
Does this happen in v1?

Delphinus8821:
In version 1.3.1 the aggregator method drop down is greyed out for all questions. 

![Version 1 3 1 Aggregator](https://user-images.githubusercontent.com/73083035/106834957-533dc380-66e6-11eb-8402-1b285551db22.PNG)

Delphinus8821:
Version 2.0.0 only has one option available in the aggregation method drop down for all of the questions.

antares1470:
Doing a quick dig I found this in `AnalyticConfigurationPane.updateGlobalParameters()`:
```
final List<AnalyticAggregatorParameterValue> aggregators = new ArrayList<>();
...
} else if (questionListPane.isExpanded() && currentQuestion != null) {
[...]

PR #966, opened 2021-02-08
:bug: Fix incorrect initial population of analytic aggregators.

Nova-2119:
Incorrect initial display of analytic aggregators was due to the 
categoryListPane.expanded value being initially set to true when it 
actually starts hidden, with the questionListPane expanded. Added a line
after initialisation to set categoryListPane.expanded to false.

### Benefits

Analytic views aggregators panel now initialises with the correct choices.

### Possible Drawbacks

NA

### Applicable Issues

#926


antares1470 (approved):

aldebaran30701 (approved):
I only seemed to get the expected behaviour.
I've tested the opening and closing of graphs, selection of graph elements all while clicking different questions, and expanding categories. all tested options seem consistent with the below behaviour.
**Questions**: One main aggregation method available (Mean | Append)
**Categories**: Global - append score only. Metrics, Similarity and Centrality - vary (multiple options available for each Analytic) 

I'm fairly confident this is working as intended and fixes the issue.

Issue #852, opened 2020-10-07
Find View hangs applications with large graphs open 


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When several large graphs are open in Constellation, opening find view causes the application to hang and lag. The issue may be to do with updating the find view state.

### Steps to Reproduce

1. Open 4 sphere graphs with 50k nodes each
2. Open Find View
3. Try using Constellation

**Expected behaviour:** Find View doesn't cause application to hang

**Actual behaviour:** Find view causes application to hang with several large graphs open

**Reproduces how often:** 100%

### Additional Information

Was found while testing #328

serpens24:
I beleive arcturus mentioned that the issue occurs when switching between open graphs weith find view (and other views) all open.

Nova-2119:
So I wasn't able to get the application to hang but I did consistently get an error. I've fixed the source of the error and I'm hoping that will fix the hanging issue described.

It feels a bit like putting a band-aid on when a proper bandage is required though. Its still acting a bit odd, for instance when you switch graphs the view might not appear, particularly if you are switching to that graph for the first time.
I think ideally it needs re-writing or atleast examining in depth, but I think that about the analytic view aswell so maybe I just don't like views!

PR #967, opened 2021-02-11
:bug: Stop error triggered when switching graphs with find open

Nova-2119:
Previously an error could occur when switching between graphs because
a FindRules attribute and type would be modified without its attributes
being updated to reflect the new type.
The line added causes the attributes value to be corrected prior to
accessing the attributes in the method that the error occurs in.

### Benefits

Stops the occurrence  of an error when switching graphs with the find view open.

### Possible Drawbacks

I was unable to reproduce the slowness described in the ticket so this may actually address a different issue. However this error occurs when performing the steps described in the reproducer and I have seen errors remain hidden and cause application slowdown before.

### Applicable Issues

#852

### DIFF
--- a/CoreFindView/src/au/gov/asd/tac/constellation/views/find/gui/FindCriteriaPanel.java
+++ b/CoreFindView/src/au/gov/asd/tac/constellation/views/find/gui/FindCriteriaPanel.java
@@ -112,6 +112,7 @@ public class FindCriteriaPanel extends JPanel implements DateTimeListenerInterfa
         createCriteriaPanels();
 
         this.localState = localState;
+        this.localState = this.getState();
 
         panelCriteriaHolder.removeAll();
 


CoreFindView/src/au/gov/asd/tac/constellation/views/find/gui/FindCriteriaPanel.java
```diff
@@ -112,6 +112,7 @@ public FindCriteriaPanel(final FindTopComponent parent, final FindRule localStat
         createCriteriaPanels();
 
         this.localState = localState;
+        this.localState = this.getState();
```
antares1470, line 115: This certainly isn't your fault but seeing these 2 lines makes me think there's some function logic that needs to be modified as this looks like it is instantiating the same field twice which looks a little pointless (in reality the 2nd call is actually just updating the existing value)

Nova-2119: 100%, working with it some stuff to just feels wrong? Like things seem to get called more often then they should and things like that. But it's the same with the analytic view so I don't know if its indicative of an issue or just a quirk of how views work, I haven't spent enough time with them.

I'd like to spend some proper time going through it but that will take weeks.

aldebaran30701:
I could not replicate the errors nor could I trigger the behaviour Nova mentioned re: switching graphs and the find view erroring or not appearing.
I did check out the PR and it appears that I also don't get any error or described behaviour.

arcturus2 (approved):

antares1470 (approved):

Issue #949, opened 2021-01-27
Exception opening graph without Plugin Reporter


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

If you close the plugin reporter, restart Constellation and try to open a graph I always get a `ConcurrentModificationException`.

```
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1042)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:996)
	at au.gov.asd.tac.constellation.graph.node.plugins.FrameworkGraphManager.graphOpened(FrameworkGraphManager.java:102)
	at au.gov.asd.tac.constellation.graph.node.GraphNode.<init>(GraphNode.java:162)
	at au.gov.asd.tac.constellation.graph.node.GraphNode.<init>(GraphNode.java:135)
	at au.gov.asd.tac.constellation.graph.interaction.gui.VisualGraphTopComponent.<init>(VisualGraphTopComponent.java:451)
	at au.gov.asd.tac.constellation.graph.interaction.gui.VisualGraphOpener$GraphFileOpener.done(VisualGraphOpener.java:222)
	at java.desktop/javax.swing.SwingWorker$5.run(SwingWorker.java:750)
	at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.run(SwingWorker.java:847)
	at java.desktop/sun.swing.AccumulativeRunnable.run(AccumulativeRunnable.java:112)
	at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.actionPerformed(SwingWorker.java:857)
	at java.desktop/javax.swing.Timer.fireActionPerformed(Timer.java:317)
	at java.desktop/javax.swing.Timer$DoPostEvent.run(Timer.java:249)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
[...]

PR #971, opened 2021-02-16
Bugfix/949 quality control view

Nova-2119:
### Applicable Issues

#949


arcturus2 (approved):
Seems to do the trick. The QC view seems to work fine with new graphs too.

aldebaran30701:
I can't seem to get the original bug to reproduce, but for myself it was only on the .exe version. I tested opening other graphs and actually found a bug with the current implementation in master.
I don't believe it is linked to this PR so I'll be making another ticket. #973

antares1470 (approved):

PR #974, opened 2021-02-16
:racecar: Converted displayPos's to int

Nova-2119:
### Applicable Issues

#407


aldebaran30701 (commented):
Just noting that I can't find any downsides to this PR during testing.
I could not trigger an out of memory error with a 50k transaction graph or a 100k transaction graph.

antares1470 (approved):

antares1470:
> Just noting that I can't find any downsides to this PR during testing.
> I could not trigger an out of memory error with a 50k transaction graph or a 100k transaction graph.

It can be quite laggy if there are lots of transactions with datetimes in close proximity to one another. But I'm no longer seeing out of memory exceptions (I think this was the case before the PR but like @aldebaran30701, I don't see any downsides to the PR)

arcturus2 (approved):

Issue #976, opened 2021-02-18
Consolidate Plugin Reporter Drop-downs into one list


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently in the plugin reporter, there is a drop-down list for selecting plugins to display (filter) and one selecting plugins to not display (ignore). It makes sense to have these drop-down lists but having them together presents some issues. Namely that the logic for each list is the inverse of the other and so can cause conflicts. e.g. What should happen if a tag is ticked in both lists? What should happen if a tag is unticked in both lists?

A solution could be to add a listener to each list to listen for changes in the other list (if I tick a tag in one list, untick it in the other) but I think a better solution would be to simply consolidate this all into one list with an option to either include or exclude all the selected tags in the list.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** User is able to clearly state which tags they want to include or exclude

**Actual behaviour:** User decides what to include and exclude from separate lists which can cause logic issues

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Issue #930, opened 2021-01-12
Plugin Reporter Null Pointer


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

In the plugin reporter if you click on the drop down list of filters, open a new graph and then close the graph it always results in a null pointer. The NPE seems to only happen if you click on the drop down list which is really strange.

![plugin reporter null pointer](https://user-images.githubusercontent.com/39325530/104276858-d316aa80-54f9-11eb-92ef-2b8a6b178734.gif)

### Steps to Reproduce

1. open the plugin reporter
2. build a sphere graph
3. click on show all in the plugin reporter
4. click on the drop down list filter (**only happens with this step**)
5. open a new graph
6. close the new graph

**Expected behaviour:** [What you expect to happen]
No errors

**Actual behaviour:** [What actually happens]
You get an NPE
**Reproduces how often:** [What percentage of the time does it reproduce?]

PR #979, opened 2021-02-23
Consolidate Plugin Reporter Drop-Down Lists

Delphinus8821:
### Description of the Change
The Plugin Reporter currently has 2 drop-downs lists in its menu bar, one for filtering the plugins shown and the other for selecting plugins to be ignored from the list shown. This change consolidates them into one drop-down list for filtering what is shown in the reporter. 

### Alternate Designs
To create a listener for each of the existing lists to have them update when the other is changed.

### Benefits
The current solution of the two drop-down lists is confusing for users as the drop-down lists appear to do the opposite of each other but do not always behave this way. Consolidating it to one list is more user friendly and understandable. 

### Possible Drawbacks
None realised

### Applicable Issues
#976

### DIFF
--- a/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
+++ b/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
@@ -21,14 +21,12 @@ import au.gov.asd.tac.constellation.graph.manager.GraphManagerListener;
 import au.gov.asd.tac.constellation.utilities.text.SeparatorConstants;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.prefs.Preferences;
 import org.openide.util.NbPreferences;
 
@@ -47,15 +45,15 @@ public class GraphReportManager {
     /**
      * Tags that will be ignored (not recorded) by GraphReport instances.
      */
-    private static final Set<String> IGNORED_TAGS;
+    private static final Set<String> TAGS;
 
     static {
         // Determine the ignored tags *before* we set up the graph reports.
         final Preferences prefs = NbPreferences.forModule(GraphReportManager.class);
         final String filteredTagString = prefs.get(IGNORED_TAGS_KEY, "LOW LEVEL").trim();
-        IGNORED_TAGS = new HashSet<>();
+        TAGS = new HashSet<>();
         if (!filteredTagString.isEmpty()) {
-            IGNORED_TAGS.addAll(Arrays.asList(filteredTagString.split(SeparatorConstants.TAB, 0)));
+            TAGS.addAll(Arrays.asList(filteredTagString.split(SeparatorConstants.TAB, 0)));
         }
 
         GraphManager.getDefault().addGraphManagerListener(new GraphManagerListener() {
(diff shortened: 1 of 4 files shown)


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
```diff
@@ -47,15 +45,15 @@
     /**
      * Tags that will be ignored (not recorded) by GraphReport instances.
      */
-    private static final Set<String> IGNORED_TAGS;
+    private static final Set<String> TAGS;
```
antares1470, line 48: Given the comment here, I'm not this field is required (given the code that used this field has been removed) so I think this can be removed

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
```diff
@@ -47,15 +45,15 @@
...
     static {
         // Determine the ignored tags *before* we set up the graph reports.
         final Preferences prefs = NbPreferences.forModule(GraphReportManager.class);
         final String filteredTagString = prefs.get(IGNORED_TAGS_KEY, "LOW LEVEL").trim();
-        IGNORED_TAGS = new HashSet<>();
+        TAGS = new HashSet<>();
         if (!filteredTagString.isEmpty()) {
-            IGNORED_TAGS.addAll(Arrays.asList(filteredTagString.split(SeparatorConstants.TAB, 0)));
+            TAGS.addAll(Arrays.asList(filteredTagString.split(SeparatorConstants.TAB, 0)));
         }
```
antares1470, line 57: Similarly, this bit of code can also be removed

antares1470 (changes requested):
If you're able to add a whatsnew entry for this as well, that would be great

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
```diff
@@ -18,46 +18,24 @@
...
 
-    /**
-     * Tags that will be ignored (not recorded) by GraphReport instances.
-     */
-    private static final Set<String> IGNORED_TAGS;
 
     static {
-        // Determine the ignored tags *before* we set up the graph reports.
-        final Preferences prefs = NbPreferences.forModule(GraphReportManager.class);
-        final String filteredTagString = prefs.get(IGNORED_TAGS_KEY, "LOW LEVEL").trim();
```
arcturus2, line 55: I'm wondering if we can retain the filtered tags logic code in the Manager class? I see that this code has now moved to the `PluginReporterPane` which is part of the Views. Given we have the Notes View also responding to the `GraphReportManager` I feel that in terms of responsibility it should be the manager that gets to initialise the filter list and let the views know. What do you think?

arcturus2 (changes requested):
I'm also able to get `PluginReporterPane` line 244 to have a `NullPointerException` occur by:

1. have two graphs open with a couple of nodes
2. running say Grid arrangement many times on each
3. selecting the drop down list once
4. switching graphs
5. running the Grid arrangement again multiple times.

Delphinus8821:
@arcturus2 This null pointer is the same as the one recorded in #930 which I almost have a fix for

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/reporting/GraphReportManager.java
arcturus2: Ignore me, you haven't moved it. I think something worth investigating later on is if more control should be given to the `GraphReportManager`.

arcturus2 (approved):

arcturus2:
> 
> 
> @arcturus2 This null pointer is the same as the one recorded in #930 which I almost have a fix for

Sweet.

antares1470 (approved):

Issue #837, opened 2020-09-03
Memory leak related to the graph


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

During testing and profiling of the application, it was noticed that the graph now has a memory leak. 
Garbage collection does not clean up the object.

### Steps to Reproduce

1. Open Constellation
2. Open new graph
3. Close new graph

**Expected behaviour:** No graphs held within memory

**Actual behaviour:** All graphs stay in memory and cannot be garbage-collected.

**Reproduces how often:** 100%

### Additional Information

N/A

Nova-2119:
So I think I have found the leak.
The visualGraphTopComponent doesnt get cleaned when a graph is closed. Instead it gets cleaned when you open the next graph. This also stops the VisualManager from getting cleaned. Most of the resources associated with the graph are tied to this somehow, so I think this explains why there is very little memory recovery when you close graph.
It also explains why the affect doesn't "stack" if you open multiple graphs, and then close them all, the memory retained is the same as when the last graph was open. As long as another graph is open the finalizer is called. If no other graph is open the finalizer is called next time you open a graph.

After some reading about memory leaks in Java my suspicions is that this due to all the static variables in the class. I'm not entirely convinced yet, but I think this is stopping the finalizer getting called on the last instance of the class? The class itself is being used by Netbeans code, so we can't really do anything up the stack to try and convince it to Garbage it.

Hence i'm thinking I will will write a method that clears all of the non-static variables and call this when componentClosed is called. I think this will minimise its memory usage and might release references that allow other classes like the visualManager to finalise. Will comment hear on how it goes!

Nova-2119:
For the record, the profiler did not help at all and if anything hindered me! After doing some reading I saw that overriding finalizer can cause memory leaks. So I did search for finalizer and found that VisualTopComponent, VisualManager etc all overwrite it.
This cause me to stumble upon MemoryManager, a class in Constellation specifically for developers to help them debug memory leaks 🕺.
[...]

PR #990, opened 2021-03-11
Performance/837 memory leak

Nova-2119:
### Benefits

Reduces retained memory usage after all graphs have been closed.

### Possible Drawbacks

Opening a graph after all graphs have been closed may be slower.

### Applicable Issues

#837


CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphTopComponent.java
```diff
@@ -549,6 +551,11 @@ public void componentClosed() {
             System.gc();
         }
 
+
+        if (GraphManager.getDefault().getAllGraphs().isEmpty()) {
```
arcturus2, line 555: Happy to approve as is but curios to know if this would be better placed before we call the `System.gc()` ?

Nova-2119: Ive made this change, logically I think it makes sense. But I didn't see any difference in behaviour when I tested this in the profiler.

MY guess is that because GC is being called from within this class the stuff that it later garbage collects if you click the button isn't yet available for GC because this class refers to it.

antares1470 (approved):

arcturus2 (approved):

Issue #984, opened 2021-03-03
Improve renderTextAsLigatures speed.


Nova-2119:
### Description

The method GlyphManagerBI.renderTextAsLigatures is much slower than the implementation prior to V1.3. In this upgrade it underwent a complete refactor to expand Constellation's comparability with different fonts.

However this had led to a significance performance impact in loading and interacting with graphs as it slows down refreshing the graph on the canvas. Performance improvements to renderTextAsLigatures will hence make an obvious impact to Constellation performance and should be explored.

Note: There is no guarantee that performance improvements are there to be had and any improvements will likely require a complete refactor rather than just tidying things up and doing things a bit more efficiently.

### Steps to Reproduce

1. Open a 100,000 node graph in the current version
2. Observe how long it takes (with multithreading)
3. Open a 100,00 graph in the [original version of Constellation uploaded to github](https://github.com/constellation-app/constellation/releases/tag/v1.20190929.144916)
4. Notice the speed (without multithreading)
5. Alternately instead of opening the current version open [original V2 without multithreading improvements](https://github.com/constellation-app/constellation/releases/tag/v2.0.0) for a more even comparison.

**Expected behaviour:** [What you expect to happen]
 Current version should be comparable to V1 in speed.

**Actual behaviour:** [What actually happens]
Current version is significantly slower than V1 for core graph features should as deleting nodes, opening graphs etc.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Additional Information

This ticket is likely to be quite involved, it would really suit someone looking to learn about rendering and/or after a substantial project to get stuck into.

arcturus2:
Just to show you how crazy it is, `renderTextAsLigatures` compared in v1.2 back in 2018 with `master` on the 16th of March.

* in v1 the method is called 3x as much and is 10x faster.

![renderTextAsLigatures v1 2 vs master 16 Mar 2021](https://user-images.githubusercontent.com/39325530/111241328-02a87700-8651-11eb-9592-1a898efa017d.png)

![renderTextAsLigatures v1 2 vs master 16 Mar 2021 telemetry](https://user-images.githubusercontent.com/39325530/111241703-c0cc0080-8651-11eb-9490-53d0ae83805d.png)

arcturus2:
Interesting to note that just by removing all the threading, the `renderTextAsLigatures` runs nearly twice as fast on a 1k node and 1k transaction graph.

[...]

Issue #980, opened 2021-02-24
High memory use deleting nodes


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The recent performance improvements in v2 are great allowing graphs to be opened so much quicker. The implementation involved multi-threading and a conscious decision to use more memory to allow for real parallelism.

It looks like deleting nodes of a graph is causing the graph labels code to run which is spiking up memory use (as a side affect of the above performance changes).

Why should deleting nodes require a complete re-render? Ideally deleting nodes should not have to call the code creating the graph top/bottom labels to be called.

Could some caching be used?

![memory increasing when deleting nodes](https://user-images.githubusercontent.com/39325530/108937385-c1045c00-76a2-11eb-9ad2-317a9b5fc5ab.gif)

Note that if I call the GC to run it will reduce memory used. Until the memory is released however, Constellation becomes a bit sluggish as the JVM deals with the memory its trying to manage with these quick successions of graph interactions.

In the gif, the response time to delete nodes took a while as well. This might be due to the slower machine I'm on and running profiling as well.

### Steps to Reproduce

1. New Sphere Graph of 10,000 nodes and 5000 transactions
2. Tree arrange
3. Observed the memory was idling at 413 MB / 500MB
4. Select some nodes with transactions and delete them
5. Repeat the process
6. Observed the memory jumps to 3.5 GB / 4 GB

**Expected behaviour:** [What you expect to happen]
Memory use should not spike very high when deleting nodes. Certainly the undo/redo stuff will take up some memory and running the delete plugin should too but nothing crazy.

**Actual behaviour:** [What actually happens]
Memory spikes and causes application to become slow/sluggish.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

Nova-2119:
High memory is due to the graph being rendered to the screen after deleting nodes.

It may seem excessive at first glance to rerender the whole graph after deleting some nodes, however the current change framework and implementation of NodeLabelBatcher do not really allow us to avoid this fate.
[...]

PR #994, opened 2021-03-21
Improve overall performance for renderTextAsLigatures

arcturus2:
### Description of the Change

With a traditional Map as a cache, the performance of renderTextAsLigatures has improved greatly and the slowest part next was adding the glyphs to the buffer. This part was synchronized due to threading so removing the synchronized calls and therefore threading improves the performance further. As a result all the recent threading added by @Nova-2119 was needed to be removed. Tried adding Guava caching with a purge timeout but it was the slowest part and negated the performance improvements being made.

With caching we are having subsequent graph opens and graph changes like deletes etc working much faster.

The following is an overall performance comparing v1.2 with `master` and this branch. This is testing on a 30k node and transaction graph.

The performance improvements was gained by caching the bulk of the work done by `renderTextAsLigatures` and removing calls to `synchronize` which and therefore required all the threading to be removed. As a result the memory spikes are a lot less meaning GC work is less intense and the CPU spikes are less too.

image

**Important Note - Note that it looks like it is taking 73 seconds to open in current `master`. This is misleading and inaccurate and apologies in advance. Remember that `master` is the only threaded version so this is the total CPU time for all threads. In practice it took around 35 seconds to render.**

As you can see though the performance is not as fast as the original v1.2, it is significantly faster than v2. Note that with caching, deleting nodes etc is faster than the original v1.2 by 100ms so interacting with the graph should be better now.

v1.2 opening graph 1st time:
v1 2 30k nodes and links 1st open

v1.2 opening graph 2nd time:
v1 2 30k nodes and links 2nd open

v1.2 deleting half of the graph:
v1 2 30k nodes and links delete half

master opening graph 1st time:
master 30k nodes and links 1st open

master opening graph 2nd time:
master 30k nodes and links 2nd open

master deleting half of the graph:
master 30k nodes and links delete half

latest opening graph 1st time:
latest 30k nodes and links 1st open

latest opening graph 2nd time:
latest 30k nodes and links 2nd open

latest deleting half of the graph:
latest 30k nodes and links delete half

**Memory and Threads Performance Usage**

This is current `master` opening the 30k nodes and links and then deleting half the graph:

master 30k nodes open and delete

This is the current branch doing the same thing. You will notice the memory usage is low again and threads are less. We knew that adding threads was going to increase memory usage but practically this is causing problems with GC called a lot more and peaking memory usage, causing an out of memory to occur sometimes.

latest 30k nodes open and delete

### Alternate Designs

I looked at using Guava caching to purge items in the cache that was written x minutes ago but it was slowing things down greatly compared to using a simple Map and the performance gains was less. Changing the cache to clear after x hours didn't make a difference either.

### Benefits

There is a reduction in cpu and memory spikes and this means the GC works a bit less. Subsequent graph opens are so much quicker now thanks to caching.

### Possible Drawbacks

The caching won't purge anything so over time the memory use will increase but I feel this is a worthwhile trade off with the current version as the memory retained does greatly help with graph interaction which is what you do most once your in Constellation.

### Verification Process

Compared v1.2, `master` and this branch running if profiling. In v1.2 I profiled `au.gov.asd.tac.constellation.visual.opengl.utilities.GlyphManager.renderTextAsLigatures`. In `master` I profiled `au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs.GlyphManagerBI.renderTextAsLigatures`. In this branch I also profiled `au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs.GlyphManagerBI.buildLigature`

### Applicable Issues

#984 and #980

### DIFF
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/batcher/ConnectionLabelBatcher.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/renderer/batcher/ConnectionLabelBatcher.java
@@ -26,17 +26,12 @@ import au.gov.asd.tac.constellation.visual.opengl.utilities.LabelUtilities;
 import au.gov.asd.tac.constellation.visual.opengl.utilities.SharedDrawable;
 import au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs.ConnectionGlyphStream;
 import au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs.ConnectionGlyphStreamContext;
-import au.gov.asd.tac.constellation.visual.opengl.utilities.glyphs.GlyphManagerBI;
 import com.jogamp.opengl.GL;
 import com.jogamp.opengl.GL3;
 import java.io.IOException;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.util.ArrayList;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Logger;
 
 /**
  *
@@ -81,10 +76,6 @@ public class ConnectionLabelBatcher implements SceneBatcher {
 
     private static final int FLOAT_BUFFER_WIDTH = 4;
     private static final int INT_BUFFER_WIDTH = 4;
-
-//    private static final int NUM_CORES = Runtime.getRuntime().availableProcessors();
-//    private static final int NUM_CORES = 1;
-//    final ExecutorService pool = Executors.newFixedThreadPool(NUM_CORES);
 
     public ConnectionLabelBatcher() {
         // Create the batch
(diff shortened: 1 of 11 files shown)


antares1470 (approved):

arcturus2:
@antares1470 noting that you approved this I wanted to make the following statement clear as not to mislead the perfomance results. I have updated the description witrh the following...

**Important Note - Note that it looks like it is taking 73 seconds to open in current `master`. This is misleading and inaccurate and apologies in advance. Remember that `master` is the only threaded version so this is the total CPU time for all threads. In practice it took around 35 seconds to render.**

aldebaran30701 (approved):
Been testing for a little while now, and it seems to be fine. no errors. For a second I thought i found an issue but it was a 160k sphere graph and it just took longer than usual to load. Still a seemingly faster load than current master.

arcturus2:
> Been testing for a little while now, and it seems to be fine. no errors. For a second I thought i found an issue but it was a 160k sphere graph and it just took longer than usual to load. Still a seemingly faster load than current master.

Cool. Deleting parts of the graph should be quicker than master as well. Also it should be using a lot less memory overall. Closing and opening that graph should be a fraction of current master.

ben-manes:
FYI: Guava defaults to a conservative setting (4) and the concurrencyLevel has to be increased in performance sensitive code (e.g. 64). Caffeine cache, a rewrite of Guava’s, is much faster and dynamical tunes itself so no setting is needed.

arcturus2:
> FYI: Guava defaults to a conservative setting (4) and the concurrencyLevel has to be increased in performance sensitive code (e.g. 64). Caffeine cache, a rewrite of Guava’s, is much faster and dynamical tunes itself so no setting is needed.

Thanks @ben-manes this looks promising. I profiled it and Caffeine isn't even making it on the profile stack. It all seems to work really fast, just like a Map would. I'll do a bit more testing but I think I will go ahead with including Caffeine as the cache.

PR #1006, opened 2021-03-31
Added bin count to Histogram

Delphinus8821:
### Description of the Change
Added a total bin count to the top of the Histogram view. 
Also set the minimum width of the Histogram view so that all elements are always able to be viewed.

Allows for the user to easily tell how many bins are currently in the Histogram.

### Applicable Issues
#996

### DIFF
--- a/CoreHistogramView/src/au/gov/asd/tac/constellation/views/histogram/HistogramDisplay.java
+++ b/CoreHistogramView/src/au/gov/asd/tac/constellation/views/histogram/HistogramDisplay.java
@@ -66,7 +66,7 @@ public class HistogramDisplay extends JPanel implements MouseInputListener, Mous
     static final String NO_VALUE = "<No Value>";
     private static final String PROPERTY_VALUE = "Property Value";
     private static final String COUNT = "Count";
-    private final String totalBinsCount = "Total Bin Count: ";
+    private static final String totalBinsCount = "Total Bin Count: ";
 
     // The color that shows where a bar would be if it was bigger.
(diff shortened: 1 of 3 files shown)


CoreHistogramView/src/au/gov/asd/tac/constellation/views/histogram/HistogramDisplay.java
```diff
@@ -339,9 +341,16 @@ public void paintComponent(Graphics g) {
...
+                final String headerStringTotalBins = getStringToFit(TOTAL_BINS_COUNT + bins.length, barsWidth, g2);
+                
+                int countTextWidth = g2.getFontMetrics().stringWidth(headerStringTotalBins);
+                
+                g2.drawString(headerStringValue, LEFT_MARGIN + iconPadding, 
+                        TOP_MARGIN + (barHeight / HALF) + correction);
+                g2.drawString(headerStringCount, barLeft, TOP_MARGIN + (barHeight / HALF) + correction);
+                g2.drawString(headerStringTotalBins, getParent().getWidth() - countTextWidth, 
+                        TOP_MARGIN + (barHeight /HALF) + correction);
+
```
arcturus2, line 353: @antares1470 I find the rule about being more than 120 chars too strict and as a result its making it harder to read for me when its split into 2 lines like this. If you agree can we omit that SonarCloud rule for now?

@Delphinus8821 don't worry about changing this back.

antares1470 (changes requested):
If you're able to add a whatsnew entry, that would be great. Looks good otherwise

antares1470 (approved):

Issue #347, opened 2020-03-23
Ability to type ahead in the drop down lists


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Abilty to type ahead in the drop down lists

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** Start typing in drop down list and it starts to list options based on what you have typed

**Actual behaviour:** nothing

**Reproduces how often:** n/a

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

sol695510:
@antares1470 which drop down lists are you referring to, or are you saying to create general functionality that can be applied to all uses of drop down lists.

antares1470:
> @antares1470 which drop down lists are you referring to, or are you saying to create general functionality that can be applied to all uses of drop down lists.

@sol695510 The drop down lists in case I believe refers to the single-choice and multi-choice parameters (which use drop down lists). In theory, the drop down lists can actually get quite large and can result in doing a lot of scrolling and searching without the ability to type ahead.

In theory I suppose this could be applied to more general combo boxes but I don't think it's necessary for this. @arcturus2 may be able to clarify this

arcturus2:
@sol695510 and @antares1470 so single and multi choice options would be fine to start with. If this option could be extended to Text Fields that would be pretty cool though.

So for some background, we originally used the ControlsFX library for a lot of the GUI components but as Java started to build these in we switched it using them. JavaFX is open source now so maybe these features exist that we can use (https://openjfx.io/index.html)? Also, these "typeahead" options look like they are build into ControlsFX so perhaps we can check out they do it and switch the drop down lists to ControlxFX again.

Auriga2:
[...]

PR #1007, opened 2021-04-01
Ability to search in `SingleChoiceParameterType`  drop down lists

Auriga2:
### Description of the Change
Users can now type on the SingleChoiceParameterType drop downs, which would filter the items in the drop down list.

### Alternate Designs
Tried a few different options but using `controlsfx.controls.SearchableComboBox` seems to handle it very well.

Better ux, specially with longer drop down lists.

### Benefits
Faster ux, specially with longer drop down lists.

### Possible Drawbacks
n/a

### Verification Process
Start typing in a `SingleChoiceParameterType` drop down. It should filter options based on what you have typed.
E.g.: `Mode` and `Type Category` drop downs in `Select Top N` plugin
Drop downs in `Extract words from Text`, `Merge Nodes`, `Merge Transactions` plugins etc.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/347


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/SingleChoiceParameterType.java
```diff
@@ -128,9 +128,22 @@
...
+
+            //Clear the existing selection
+            parameter.setObjectValue(null);
+
         parameterValue.setOptions(options);
         parameter.setProperty(CHOICES, new Object());
+        }
+    }
+
+    private static boolean optionsChanged(PluginParameter<SingleChoiceParameterValue> parameter, List<String> options) {
```
antares1470, line 143: both parameters can be made final

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/SingleChoiceParameterType.java
```diff
@@ -128,9 +128,22 @@
...
+        final SingleChoiceParameterValue parameterValue = parameter.getParameterValue();
+        boolean bol = !options.equals(parameterValue.getOptions());
+        return bol;
```
antares1470, line 146: these lines can be reduced to `return !options.equals(parameter.getParameterValue().getOptions())`

antares1470 (changes requested):
If you're able to add a whatsnew entry as well that would be great

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/SingleChoiceParameterType.java
Auriga2: Yeah already fixed it but didn't push that file. Added that for testing.

arcturus2 (changes requested):
Thank you @Auriga2 this is really nice to have as a feature and works well. There seems to be one issue with a delayed update to sub lists and I had to click twice to get the values for the list.

I noticed that if I start to type after the first click, items do show up so it looks like the list is populated but its not visible until the second click/GUI update.

Do you think this could be fixed and if its an easy fix?

drop down list delay

Auriga2:
@arcturus2 I can recreate it only when running after a clean build. Could you recreate it without that?

Auriga2:
No need to clean and build to recreate. But still need to close the consty and debug run each time. This is a bit tricky as I could not able to recreate in ControlsFX yet. I tried to fix locally by initializing some values in the drop downs by default. 

If I set to Node (the longer list), selecting the transactions throws the same `NoSuchElementException` exception as before. 
If I set to Transaction ( the shorter list),  selecting the Node doesn't show the entire list or the side scroll bar until clicked twice on it. Looks like this is still initialized to the smaller size still.

Also tried to select an option by default. This populated the selection, but the behavior in the drop down list is still the same.

This issue presents only when the DAV is loaded for the first time , and also only on a dependent drop down, probably only in the `Select Top N` plugin. (Are there any other such drop downs in Consty?)

Auriga2:
Found a workaround. Also removed some redundant code.

arcturus2 (approved):
Your a legend @Auriga2 awesome stuff and it works really well.

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,9 @@
 == 3030-12-31 Getting Started
 <p>If you're new to CONSTELLATION, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.introduction">introduction</a>.</p>
 
+== 2021-04-01 Ability to Search in Single Choice Drop-Down Lists
+<p>Single Choice Drop-Downs are now editable, which would filter the options in the drop down list based on what is typed.</p>
```
antares1470, line 5: change "editable" to "searchable" and "which would" to "which allows you to"

antares1470 (changes requested):
Tested and looks good otherwise

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
Auriga2: Yes they are not editable. Fixed.

antares1470 (approved):

PR #1014, opened 2021-04-15
Bugfix/remember dropdownlist options

mimosa2:
### Description of the Change

Back out the  changes - remember drop down list options 

Core feature

### Benefits

Avoid conflicts in handling global parameters in DA view tabs 

### Possible Drawbacks

Removes the feature - remember drop down list options 

### Verification Process
1. Open a graph & Data Access view.
2. Select a plugin with drop down list --> select a value & run the plugin
3. Close the application
4. Re-open the application
5. The value selected previously is not remembered.


arcturus2 (changes requested):
@mimosa2 would you be able to please delete the commented code given we are looking to take a different approach this with change? Thank you.

arcturus2 (approved):

antares1470 (approved):

PR #1015, opened 2021-04-18
Significant memory saving in Table View

arcturus2:
### Description of the Change

Reuse strings to reduce the amount of String objects in memory

The screenshots below show the before and after with a Sphere Graph build of 30k nodes and transactions and then opening the table view

**Before**
image

**After**
Screenshot from 2021-04-18 09-38-00


antares1470 (approved):

aldebaran30701 (approved):
Seems to work as intended and I couldn't get any bugs to occur which is always good

PR #1018, opened 2021-04-21
Reduce memory used by GlyphManagerBI

arcturus2:
### Description of the Change

* Converted a List to an array given there was only a few elements in the List there was too much overhead using a List
* Cached the Rectangle objects given many duplicates existed in memory
* Created a new FourTuple data structure

For a 30k nodes and transactions graph the below was the improvement observed comparing heap dumps:

**Before**
File size: 348.7 MB

image

**After**
File size: 335.2 MB

image

image

**Conclusion**
* A reduction of close to 3.8% of memory use.
* The amount of ArrayList objects went down from 41k to 10k
* The amount of Rectangle objects went down from 150k to just 4k

### Benefits

Reduce memory footprint of the application


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/FourTuple.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param <C> the type of the third object in the 4-tuple
+     * @param <D> the type of the third object in the 4-tuple
+     * @param first the first object in the tuple
+     * @param second the second object in the tuple
+     * @param third the third object in the tuple
+     * @param fourth the fourth object in the tuple
+     * @return a 4-tuple that is templated with the types of first, second,
+     * third and fourth
+     */
+    public static <A, B, C, D> FourTuple<A, B, C, D> create(A first, B second, C third, D fourth) {
```
antares1470, line 63: all the parameters can be final

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/FourTuple.java
```diff
@@ -0,0 +1,198 @@
...
+        this.fourth = fourth;
+    }
+
+    public Stream<Object> stream() {
+        return Stream.of(first, second, third, fourth);
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (!(obj instanceof FourTuple)) {
```
antares1470, line 105: was there a reason for formulating the if conditional this way as appose to `this.getClass() != obj.getClass()` (which is what I see more commonly in equals functions)

antares1470 (changes requested):

aldebaran30701 (approved):
Love your work on the tests!
All good by me after the comments from @antares1470 are resolved.
No noticeable issues when I tested it either.

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/FourTuple.java
arcturus2: Fixed and updated the other Tuple classes also

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/FourTuple.java
arcturus2: Good pickup and after reading https://stackoverflow.com/a/4989843 I have gone with using `getClass`. Updated for all the Tuple classes.

antares1470 (approved):

Issue #340, opened 2020-03-23
The application documentation needs an overhaul


serpens24:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The current state of Constellation's documentation is rather poor. Some of the issues I've identified include:
 * documentation of the graph itself is somewhat outdated and fails to include some of the more recent functionality additions.
 * schemas are completely undocumented.
 * the plugin framework is completely undocumented.
 * several import/export tools are completely undocumented (google earth, shapefile, image, etc).
 * several arrangements are completely undocumented (hierarchy, proximity, spectral, etc), and some of those that are documented are done so in an extremely poor manner (bubble tree).
 * several tools are completely undocumented (document clustering, hierarchical clustering, complete with schema, external scripting api, composite nodes, etc).
 * several views are completely undocumented (histogram, schema view, scripting view, etc).
 * the attribute editor documentation is lacking any real information, and there still exists a help page on the now non-existent attribute value editor (under graph > editing > edit attributes).
 * named selections documentation is in the wrong category (in graph, should be in views).
 * there is no consistent style or formatting across documents (header styles differ, font sizes differ, many images are poorly aligned, poorly sized, or stretched out, etc).
 * Add missing help pages
 * Help topics for data access view categories
 ** Help topics for data access view categories were requested. These would only need to be very brief.
 ** Consider using tool tips here

Suggested the Analytic view help could appear similar to the DAV or Schema View

- Could include pre-requisite information of how the graph should be structured before running some analytics
- The help documentation could include example graph diagrams to help explain how they work
- Some of the help documentation is still too technical

 I believe this poor presentation will create a bad first impression for anyone learning to use the application and should be completely overhauled.

**Expected behaviour:** Documentation is as complete as possible and make it as easy as possible for new users to use Constellation.

**Actual behaviour:** Documentation is out of date and incomplete.

**Reproduces how often:** 100%

## Additional Information
**Help pages to review/update:**
*Overview*
* [x] About Constellation
* [x] Getting Started
* [x] The Graph Window

[...]

PR #1019, opened 2021-04-26
Help Page Update 2

antares1470:
### Description of the Change

- removed analytic help pages in the Tools section, tidied up the analytic help pages in the Analytic View section
- added help pages for developer preferences and graph preferences
- removed help page for display preferences
- moved the label fonts help page from Visualisation section to Preferences section
- updated the following help pages:
    - all the analytic questions
    - Find View
    - Table View (Experimental)
    - datetime range global parameter
    - application preferences
    - proxy preferences (note: this one still requires further updates to the content, main focus ended up being formatting)
    - label fonts
    - all the map view layers (bar the experimental one which currently doesn't have a page)
    - all the map view overlays
    - compare graph
    - graph model
    - attributes
    - types

Updating help pages in Core

### Benefits

Help pages are more up to date

### Possible Drawbacks

Nil

### Applicable Issues

#340


aldebaran30701 (approved):
Couldn't see anything that stuck out as not being correct to me. 
Every help page that seemed to be changed has been tested by locating it's help button and verifying that it opens the correct page.
Love the work with the analytic view help pages - formatting and content is excellent.

Delphinus8821 (approved):

CorePreferences/src/au/gov/asd/tac/constellation/preferences/docs/developer-preferences.html
```diff
@@ -29,12 +29,31 @@ <h2>Memory Usage</h2>
...
         <h2>Master Reset Constellation</h2>
         <p>
-            
+            Constellation is built on top of Apache Netbeans with a number of module suites and modules within those module suites.
+            Running this option is a distruptive change that will uninstall all Constellation modules, stripping the application
+            down to just the NetBeans runtime libraries.
+        </p>
+        <p>
+            Note that this option does not work for the standalone version of Constellation and it rather applicable to the version
+            of Constellation that is continously updating via the update site mechanism. This version is called the "Constellation Shell".
```
Auriga2: Typo "continously"

Auriga2:
A help button should be added to the Schema View as well.

CorePreferences/src/au/gov/asd/tac/constellation/preferences/docs/developer-preferences.html
antares1470: Good spot. Will fix

antares1470:
> A help button should be added to the Schema View as well.

Agree that one should be added. The reason I haven't added one at this stage is that there wasn't an intuitive place to slot it in (with respect to both the gui and the code) so will probably save it for another time

Auriga2:
Makes sense. I noticed some more views with missing help button. (Scripting view, Scatter plot, Named selection, Table View (Experimental) etc. ) 

Also the "Select All" plugin in DAV, selects only the nodes. Is it broken functionality, or should we reword the help text?

arcturus2 (approved):
Great stuff @antares1470

antares1470:
> Also the "Select All" plugin in DAV, selects only the nodes. Is it broken functionality, or should we reword the help text?

@Auriga2 I don't think it is broken functionality as it appears it was written to only select all the nodes. On the other hand, I feel that select all should probably select all the transactions as well so I will go ahead and create a ticket to make that amendment.

Issue #977, opened 2021-02-18
Create custom attribute dialog pops under main window on Linux v2.1.1


mckeowbc:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

On the Linux version of v2.1.1 when trying to import from a delimited file, the + buttons to add a new custom attribute brings up the dialog behind the main window. This requires moving the window each time to find the dialog.  Additionally in the Ubuntu 20.10 window manager you cannot Alt-Tab to these dialogs.

### Steps to Reproduce

1. Open Constellation
2. Select File->Import->From Delimited File...
3. Add a source file
4. Click on one of the + buttons under attributes to attempt to create a custom attribute.

**Expected behaviour:** Dialog box appears above main window

**Actual behaviour:** Window pops behind main window

**Reproduces how often:** 100%

### Additional Information

OS: Ubuntu 20.10

[SupportPackage-mckeowbc-2021-02-18-16-26-34.zip](https://github.com/constellation-app/constellation/files/6005984/SupportPackage-mckeowbc-2021-02-18-16-26-34.zip)

aldebaran30701:
I'm yet to test if this is an issue with windows, but am working on similar code and it looks like the `NewAttributeDialog.java` class calls `initOwner(...)` with the parent window. 
There may be a call to something that isn't correct for Linux or the call to get the parent window is not correct.
Another option which may be worth pursuing is to set the modality using `initModality()` to `Modality.APPLICATION_MODAL` which  may cause the focus to be on that window even when the parent window is not set correctly.

aldebaran30701:
UPDATE: Tested on Windows using NetBeans debug to launch, the dialog initially pops up on top of the window, but it is not appearing to be modal as you can select Constellation behind it. The .executable version of Constellation also displays the same behaviour.

aldebaran30701:
@mckeowbc Are you able to test out #1020 to verify that it solves the problem you have experienced?

arcturus2:
[...]

PR #1020, opened 2021-04-27
:bug: Launch custom attribute creation dialog as front-most dialog

aldebaran30701:
### Description of the Change
This sets the dialog to always be on top and also calls it to be the front-most window of the dialogs displayed.
I have not been able to test it on Linux as per the ticket - but it does appear to circumvent it on windows to the level it was appearing.
@arcturus2 @mckeowbc  Would you be able to test if this fixes the problem on Linux?

### Alternate Designs
Place window into a Swing or JFX Dialog to make it appear as front-most.

Hopefully it fixes the issue related to the ticket.

### Benefits
No hidden dialogs

### Possible Drawbacks
Annoying window which can't be minimized, only closed.

### Verification Process
Tested original ticket on windows, and fixed the issue of being able to put the dialog behind the main window.

### Applicable Issues
#977


aldebaran30701:
This addition at the current time does not solve the issue of having the dialog appear at the front on Linux. More testing and an alternate solution is needed to solve the issue ticket.

Issue #998, opened 2021-03-22
GitHub Actions Improvements


aldebaran30701:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Following the conversion from TravisCI to GitHub Actions #993 , there are a few improvements that could be made to round this ticket out.
~~1.Cache the downloaded dependencies within the Docker container. (tried using the cache action and mounting it to the correct folder on Docker but it seems to always cache nothing even though there is only one .ivy2/cache directory at /root/)~~ (Fixed with #1021 )
2.Add an action to package up the release on tagged versions.

### Steps to Reproduce

~~1. Trigger a github action to run, and view the logging in the checks tab to ensure that the cache is reloaded between triggers of docker containers. The cache is stored with a hash of the ivy.xml file.~~

**Expected behaviour:** ~~the cached dependencies will be reloaded into the correct docker .xml file and will not dowload any further dependencies if working correctly.~~

**Actual behaviour:** ~~The dependency is compared, found to be correct - and then loaded into docker. The cache is empty at that point. Where it should be full of downloads.~~

**Reproduces how often:** 100%

### Additional Information
~~The cache action is specified in .github -> workflows -> `build-and-scan.yml`~~
~~The cache directory is set at ${HOME}/.ivy2/cache~~
~~This is then mounted when launching the docker container.~~

~~Saving the cache seems to work at correct time, but quite possibly doesn't save the information into the cache for some reason or another.~~

aldebaran30701:
To also add - Sonar scanning does not seem to be picking up test coverage.
I believe this is because it is running one docker container for the tests, and another for scanning with sonar.
Ideally we need the resultant files that are output by jacoco to be within the container of the sonar scan.
They currently run simultaneously - so unless there is some tricky wait call in github actions that would allow us to add the file beforehand - I don't have any great ideas on how to resolve this.

aldebaran30701:
The remaining work on this ticket is to configure a GitHub action to build constellation when a tagged version is made.
[...]

PR #1021, opened 2021-04-27
Add code coverage to sonar report

aldebaran30701:
### Description of the Change
This PR fixes the code coverage rating for the tests ran on Constellation.
The Coverage % was not working because the docker container running the sonar scan did not contain the reports needed. To rectify this, I have merged the two actions of running tests and scanning sonar into the same GitHub Action.
The issue with this is that the jobs cannot be ran at the same time. There was an additional step which performed a clean and build for the different Action. This has now been removed, and has allowed Sonar to pick up the test results.

### Alternate Designs
N/A

Adds code coverage figures for test results

### Benefits
Adds code coverage figures for test results
Easier to measure progress between efforts to increase coverage

### Possible Drawbacks
Slightly longer CI times in the Action.

### Verification Process
Tested via opening the sonar results set online, and viewing logs of the github actions.

### Applicable Issues
#998


arcturus2 (approved):
@aldebaran30701 ❤️ ur work thank you for fixing this.

.githubutilities/sonar.sh
```diff
@@ -2,15 +2,6 @@
...
 
-title "Updating dependencies and building"
-
-ant \
-  -Dnbplatform.active.dir="${NETBEANS_HOME}" \
-  -Dnbplatform.default.netbeans.dest.dir="${NETBEANS_HOME}" \
-  -Dnbplatform.default.harness.dir="${NETBEANS_HOME}"/harness \
-  -Dupdate.dependencies=true \
-  -Dbuild.compiler.debug=true update-dependencies-clean-build
-
```
arcturus2, line 13: good catch. no need to do this again.

antares1470 (approved):

Issue #1023, opened 2021-04-30
Code Test Coverage


aldebaran30701:
### Description

It is a good idea to increase the coverage of testing on the most commonly used features of Constellation.
From this came the idea to aim for 100% test coverage on the following list of most commonly used features. The list of classes that are involved in the execution of that feature must also be tested.

Mark the class as in progress before you take on any test development.

## Plugin Framework

* [ ] DefaultPluginEnvironment (au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.java)
    * [x]  In Progress
* [x] DefaultPluginInteraction (au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginInteraction.java)
    * [x]  In Progress

## Common Plugins

* [x] Select In Box (au.gov.asd.tac.constellation.graph.interaction.plugins.select.BoxSelectionPlugin.java)
    * [x]  In Progress
* [x] PointSelectionPlugin (au.gov.asd.tac.constellation.graph.interaction.plugins.select.PointSelectionPlugin.java)
    * [x]  In Progress
* [x] QualityControlAutoVetter (au.gov.asd.tac.constellation.views.qualitycontrol.daemon.QualityControlAutoVetter.java)
    * [x]  In Progress
* [x] DataAccessViewTopComponent (au.gov.asd.tac.constellation.views.dataaccess.panes.DataAccessViewTopComponent.java)
    * [x]  In Progress
* [x] QualityControlViewPane (au.gov.asd.tac.constellation.views.qualitycontrol.QualityControlViewPane.java)
    * [x]  In Progress
* [x] QualityControlViewTopComponent (au.gov.asd.tac.constellation.views.qualitycontrol.QualityControlViewTopComponent.java)
    * [x]  In Progress
* [x] DefaultQualityControlAutoButton (au.gov.asd.tac.constellation.views.qualitycontrol.widget.DefaultQualityControlAutoButton.java)
    * [x]  In Progress
* [ ] ClusteringManager (au.gov.asd.tac.constellation.views.timeline.clustering.ClusteringManager.java)
    * [ ]  In Progress
* [x] DefaultInteractionEventHandler (au.gov.asd.tac.constellation.graph.interaction.visual.DefaultInteractionEventHandler.java)
    * [x]  In Progress
* [x] AutosaveGraphPlugin (au.gov.asd.tac.constellation.graph.interaction.plugins.io.AutosaveGraphPlugin.java)
    * [x]  In Progress

[...]

PR #1027, opened 2021-05-04
QualityControlViewPane unit tests

antares1470:
### Description of the Change

Added unit tests for `QualityControlViewPane`. Also did a little code tidyup + made some adjustments in order to allow testing to occur

Unit testing a class in Core

### Benefits

Better chance of catching out errors in the code

### Possible Drawbacks

-

### Applicable Issues

#1023


aldebaran30701:
There looks to have been an issue with the CI - running tests. Would you be able to look into if this is an issue with your changes specifically?

CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewPane.java
```diff
@@ -680,46 +677,34 @@ private static void readSerializedRulePriorities() {
...
         }
         return rulePriorities;
     }
 
-    /**
-     * Lazily instantiates the lookup object
-     *
-     * @return the cached lookup object
-     */
-    public static Lookup getLookup() {
```
aldebaran30701, line 695: I believe @arcturus2 mentioned that in the past there has been slowness issues with always fetching the lookup. I think this was added as a way to store the lookup object - Keen to hear if @arcturus2 thoughts are still the same or if it's a non-issue

arcturus2: You've got a good memory @aldebaran30701. So there is definitely a performance issue making `Lookup` calls like `Lookup.getDefault().lookup(Something.class)` but `Lookup.getDefault()` should be fine as its just asking for the singleton version of `Lookup` which is OK.

Just to double check I profiled the 2 methods where `Lookup` was being called in this branch and can confirm its fine.

image

aldebaran30701: Thanks for the detailed response! Continue as you were then :D

arcturus2:
Regarding the issue with the unit test not completing I think the issue might be related to the calls to `new JFXPanel()` or something else that requires graphics to be enabled. I am going to look at introducing Xvfb (https://en.wikipedia.org/wiki/Xvfb) into the build image to see if that will help resolve the issue.

aldebaran30701 (approved):

arcturus2 (approved):

PR #1035, opened 2021-05-11
Delete Selection Plugin Unit Tests

Delphinus8821:
### Description of the Change
Added unit tests for the DeleteSelectionPlugin class.

Created a unit test of a class that is in core.

### Benefits
Better chance of catching errors with the Delete Selection Function

### DIFF
--- a/CoreInteractiveGraph/test/unit/src/au/gov/asd/tac/constellation/graph/interaction/plugins/delete/DeleteSelectionPluginNGTest.java
+++ b/CoreInteractiveGraph/test/unit/src/au/gov/asd/tac/constellation/graph/interaction/plugins/delete/DeleteSelectionPluginNGTest.java
@@ -98,18 +98,6 @@ public class DeleteSelectionPluginNGTest {
     @AfterMethod
     public void tearDownMethod() throws Exception {
     }
-
-    /**
-     * Test of edit method, of class DeleteSelectionPlugin.
-     * @throws Exception
-     */
-    @Test
-    public void testEdit() throws Exception {        
-        testNothingSelected();
-        testSomeVxSelected();
-        testSomeTxSelected();
-        testAllSelected();      
-    }
     
     /**
      * Tests that nothing is deleted as nothing is selected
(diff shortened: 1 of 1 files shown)


CoreInteractiveGraph/test/unit/src/au/gov/asd/tac/constellation/graph/interaction/plugins/delete/DeleteSelectionPluginNGTest.java
```diff
@@ -0,0 +1,215 @@
...
+    @AfterClass
+    public static void tearDownClass() throws Exception {
+    }
+
+    @BeforeMethod
+    public void setUpMethod() throws Exception {
+        final Schema schema = SchemaFactoryUtilities.getSchemaFactory(AnalyticSchemaFactory.ANALYTIC_SCHEMA_ID).createSchema();
+        graph = new DualGraph(schema);
+        
+        WritableGraph wg = graph.getWritableGraph("Delete", true);
```
antares1470, line 64: just looking at the rest of the test code, you could potentially just use a StoreGraph here rather than a DualGraph. Will also save you having to use a try-finally block

CoreInteractiveGraph/test/unit/src/au/gov/asd/tac/constellation/graph/interaction/plugins/delete/DeleteSelectionPluginNGTest.java
```diff
@@ -58,53 +59,47 @@ public static void tearDownClass() throws Exception {
...
-        
-        WritableGraph wg = graph.getWritableGraph("Delete", true);
-        try {     
-            attrX = VisualConcept.VertexAttribute.X.ensure(wg);
-            attrY = VisualConcept.VertexAttribute.Y.ensure(wg);
-            vAttrId = VisualConcept.VertexAttribute.SELECTED.ensure(wg);
-            tAttrId = VisualConcept.TransactionAttribute.SELECTED.ensure(wg);
+        graph = new StoreGraph();
+            
+            attrX = VisualConcept.VertexAttribute.X.ensure(graph);
```
antares1470, line 64: If you're able to fix the formatting here that would be great

antares1470 (changes requested):
Looks good otherwise

arcturus2 (changes requested):
It looks neat but the issue for me is that it looks like only 1 test is run. Actually you have run several tests here. It also means that if 1 of the 4 tests fail we don't know which one.

Also the Test NG framework will run the `setUpMethod` for you so all you need to do is add the `@Test` annotation to each individual test instead and get rid of `testEdit()`.

image

arcturus2 (approved):

Issue #1016, opened 2021-04-19
Default Font Issue


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When Constellation launches the font I think is Arial throughout the application. However, when I go into Setup -> Options -> Constellation to change a preference and apply it, it also picks up the font setting in the Misc tab and changes everything to Monospaced.

### Steps to Reproduce

1. Go to Setup -> Options -> Constellation -> Application
2. Untick the Freeze Graph View
3. Click on OK

**Expected behaviour:** [What you expect to happen]
The fonts should remain as Arial

**Actual behaviour:** [What actually happens]
The font changes to Monospaced

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Additional Information

The Miscellaneous tab is what we use to change the font in Constellation. Refer to `FontUtilities.class` to how that works. The Misc tab is actually meant to be for the Output window so we shouldn't really be listening to this for Application font changes.

I would like to have a Font selection option in the Application preferences tab and perhaps somehow hide the Misc tab altogether.

GammaVel:
On track for PR to be ready by the end of the week

PR #1056, opened 2021-05-21
Create Default Font Settings

Delphinus8821:
### Description of the Change
Created a new tab in the CONSTELLATION settings for users to set their preference for the font family and size they would like the views of Constellation to be displayed as. The majority of the views have been updated to use this preference for all of the text in the views.

**Views included:**
Analytic View
Attribute Editor
Conversation View 
Data Access View
Delimited File Importer
JDBC Importer
Layers View
Notes View
Plugin Reporter 
Quality Control View
Scatter Plot
Table View
Welcome View
What's New View 

### Alternate Designs
None realised

This preference setting allows for users to view Constellation in the font and font size they prefer.

### Benefits
Users can set the default font and size to what they find easier to read.

### Possible Drawbacks
N/A

### Verification Process
Tested opening all views with a variety of different fonts and sizes. 
Also tested opening Constellation without closing the views prior to ensure they updated correctly at program launch.

### Applicable Issues
#1016


CorePreferences/src/au/gov/asd/tac/constellation/preferences/docs/font-preferences.html
```diff
@@ -0,0 +1,19 @@
...
+<html lang="en">
+    <head>
+        <title>Font Preferences</title>
+        <link rel="stylesheet" href="nbdocs://au.gov.asd.tac.constellation.preferences/au/gov/asd/tac/constellation/preferences/constellation.css" type="text/css">
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    </head>
+    <body>
+        <h1>Font Preferences</h1>       
+        <p>
+            Font preferences can be accessed via Setup -&gt; Options -&gt; CONSTELLATION -&gt; Font.
```
antares1470, line 11: The name of the tab from what I could see was "Application Font" rather than "Font" so this should probably be updated here and in other places in the help (such as table of contents, title of page, etc.) just to avoid confusion.

antares1470 (changes requested):
One minor thing but otherwise it looks really good.

antares1470 (approved):
Once the merge conflict is resolved it should be good to go

arcturus2 (changes requested):
@Delphinus8821 I had a go to see if I could fix some of the bugs like the font not updating when you hit apply and then got stuck into things and was able to find a way to set the font without having to `setStyle` to all buttons and labels. See my suggestions https://github.com/constellation-app/constellation/pull/1061.

The advantage is that it will allow views to be created without having to explicitly set the font and family making it easier to create views and be compliant.

Let me know what you think.

arcturus2 (approved):
I'd like to get this into the RC so happy to approve and I continue my PR when we can.

Issue #987, opened 2021-03-10
JDBC Importer Fixes


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The following issues we're identified by @cygnus-x-1 when testing the JDBC importer:
- You currently cannot add a connection without entering a username / password, even if it is not required.
- I'm unclear why there is an option for a username and password as that information was requested as part of the connection setup. Perhaps it was only needed for testing the connection there? If that's the case, then the UI should indicate that.
- The instructions in the 'Configuration' pane don't actually mention that you need to perform an SQL query to retrieve sample data before dragging and dropping attributes.
- I am unable to perform metadata queries of a database (.table, .schema , etc) as they fail validation.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Auriga2:
@cygnus-x-1 
> - You currently cannot add a connection without entering a username / password, even if it is not required.

What do you mean by ` it is not required.` here? Are there databases which don't require authentication to connect to? 
Currently a connection is added only after verified by `test`ing, which requires username and password. 

>- I'm unclear why there is an option for a username and password as that information was requested as part of the connection setup. Perhaps it was only needed for testing the connection there? If that's the case, then the UI should indicate that.

Username and password are used for testing a connection.  Hence we need them when adding a connection. They can be stored with the connection and used when querying in the plugin if it's safe enough to store them as strings in the SQLite database.  Or should we try to encrypt and decrypt the password?

Auriga2:
[...]

Issue #141, opened 2019-10-10
Support for a generic JDBC connector


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The ability to connect Constellation to a database using JDBC is required for end users to be able to import data into Constellation without needing to be a developer and writing their own connector.

The JDBC Import GUI should allow a user to connect to a database by providing the host, port, database, username and password.

You should be able type your SQL query and then map the results to node or transaction attributes.

Consider using the Import From Delimited to implement the GUI.

The intention would be that a generic connector should work with MySQL, Postgres, Impala, Phoenix and for Hive's using their JDBC driver.

Ideally Kerberos support would be good to have but for the initial version maybe keep it out of scope.

Suggestions welcomed.

arcturus2:
A basic experimental version is available but I'd like to look at doing a new one in JavaFX and probably link it in with the delimited file importer. Following on from the developer catchup I said I would close this and create a new one...going to back track on that and instead put this ticket back in todo which should cover the final solution.

cygnus-x-1:
Since the Delimited File Importer is now a sidebar thing - could we also make this a sidebar thing for consistency?

aldebaran30701:
@cygnus-x-1 I've made a ticket just now upon seeing your comment - #956

cygnus-x-1:
I ran into some major issues, but first here are some usability comments:

- The workflow is add a driver > add a connection using that driver > query a database using that connection, however the UI asks you to set this up in the reverse order.
- When adding a driver, the file dialog should limit to only relevant file types (JAR presumably).
- You currently cannot add a connection without entering a username / password, even if it is not required.
- I'm unclear why there is an option for a username and password as that information was requested as part of the connection setup. Perhaps it was only needed for testing the connection there? If that's the case, then the UI should indicate that.
- There is no way to modify a driver or connection once it has been created short of removing and re-adding it.
[...]

PR #1060, opened 2021-05-24
JDBC importer fixes

Auriga2:
### Description of the Change

- Connections  can now be added without entering a username / password, if they are not required. Leave the username / password blank when querying such databases.
- Improved the instructions in the 'Configuration' pane 
- Imported data now displays all the rows, where it was skipping the first row before
- The IllegalStateException when quering a SQLite database is fixed now.
- Exceptions are no longer thrown with error pop ups, E.g.: When username/password are missing when they are required.
- In the `Add Driver` Window, the `Name` field now lists under the `Driver` field and selecting the `Driver` now auto populates the `Name` field. 

### Alternate Designs
n/a

Improved UX

### Benefits
Improved UX

### Possible Drawbacks
n/a

### Verification Process
As mentioned in the description.
More details are in the https://github.com/constellation-app/constellation/issues/987

### Applicable Issues
https://github.com/constellation-app/constellation/issues/987
Details of the SQLite exception listed here: https://github.com/constellation-app/constellation/issues/141


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
```diff
@@ -145,31 +145,36 @@ protected void updateSampleData() {
...
-                currentColumns = new String[ps.getMetaData().getColumnCount() + 1];
-                currentColumns[0] = "Row";
-                final ResultSetMetaData rsmd = ps.getMetaData();
-                for (int i = 0; i < rsmd.getColumnCount(); i++) {
-                    currentColumns[i + 1] = rsmd.getColumnName(i + 1);
-                }
+
             } catch (final MalformedURLException | ClassNotFoundException | SQLException | NoSuchMethodException
-                    | InstantiationException | IllegalAccessException | IllegalArgumentException
+                    | InstantiationException | IllegalAccessException | IllegalArgumentException | NullPointerException
```
antares1470, line 174: Rather than catching a null pointer here, a null check should be done where the exception could be thrown

antares1470 (changes requested):

antares1470 (changes requested):
Just gave it a test. I note that the instructions in the config pane don't wrap around when they don't fit so you have to expand the component in order to read the instructions. If there is a way to wrap that text, that would be good

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
arcturus2: I was unable to see when the NPE was occurring. @Auriga2 do you have some examples?
@antares1470 I have removed the check for now and will see if we can add the null check when we get some examples.

arcturus2:
@Auriga2 @antares1470 I'm going to merge this in as I want to include these fixes into the v2.3 RC

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
Auriga2: I added it while troubleshooting the SQLite exception, so it makes sense to remove now.

Issue #57, opened 2019-07-26
Allow pinning of nodes to the graph


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a user
i want to be able to pin a node to its location
so that i can arrange other nodes around them

Pinning nodes to the graph would be great.

Being able to fix particular nodes in place and then rearrange any unpinned nodes (especially for new nodes) around them would be great.

serpens24:
@arcturus2 - how would you propose this would work if say 2 nodes were pinned, and both selected, and the user selected to arrange in a grid (or for instance a new horizontal or vertical line) ...... how would you propose it would work .. would the pinned nodes just stay where they are (as if not selected) ?
The reason I ask, is this sort of 'arrangement' operation could potentially conflict with the pinned postion of nodes.

arcturus2:
Hi @serpens24 the intended behaviour would be as if they are not selected. By being pinned, they really should not move from that x,y,z position. Noted that other arrangements could sit on top of them and that will probably be something we have to deal with down the track like arranging nodes in a place that doesn't overlap with existing nodes. That isn't something our arrangement framework currently supports and would be out of scope for this implementation.

I made a poor/failed attempt at keeping nodes positioned at the same place which is the "Freeze graph view" feature you will see on the toolbar (blue snowflake). This implementation was mainly around the Data Access View and the scenario of running a data access plugin which finishes off by doing some arrangement. If a user has their graph laid out in a particular way, running plugins shouldn't re-arrange them if they don't wont them to.

I say that my implementation is broken because it only works for plugins that implement `RecordStoreQueryPlugin` and each plugin has to have an if check to see if the Freeze graph view is enabled or not. 

**Base on feedback this feature was proposed and the implementation should ideally be down at the OpenGL level for maximum efficiency and to also not require any plugin to know/care if nodes are pinned.**

Does that make sense?

Guilty-Spark-343:
[...]

PR #1070, opened 2021-05-28
Visual Schema Update for pinned decorator

antares1470:
### Description of the Change

This schema update adds the pinned attribute as a decorator to old graphs so that there is visual indicator to users when a node is pinned. This overrides any existing decorator in the top right.

### Alternate Designs

Also considered only applying the decorator if there already wasn't a decorator in the top right

Update to reflect a change in core

### Benefits

Users can see the pinned attribute as a decorator on old graphs without having to manually change the decorators from the attribute editor

### Possible Drawbacks

If a user had customised the decorators in their graph, they will the NE decorator they had set as it will be overwritten

### Applicable Issues

#57


arcturus2 (approved):

arcturus2:
I created https://github.com/constellation-app/constellation/pull/1075 if you want to include that before you merge this?

PR #1074, opened 2021-05-28
Fix a bug with Import from Delimited

arcturus2:
### Description of the Change

Fix a bug with Import from Delimited
* Fixed an exception being thrown when importing via import from
  delimited

**NOTE:** I think this PR should go in before we release v2.3


aldebaran30701 (approved):

Auriga2:
The bug there seems fixed now. But there is another exception thrown if the `Datetime` attribute is dropped onto a non date time column and continue to import. The column header is highlighted in red, but not so obvious in the black background:
image

image
It'd be nice if a proper message is displayed when the  `Datetime` attribute is dropped onto a non date time column.

antares1470 (approved):

Auriga2:
I'll fix the above exception and create a new PR.  The ideal solution would be to prevent it from dropping any incorrect attributes onto columns, rather than highlighting in red, because currently it throws the exceptions if the user continues to import with that.

Auriga2 (approved):

PR #1034, opened 2021-05-11
✅ Quality Control Auto Vetter Unit Tests

aldebaran30701:
Created `QualityControlAutoVetterNGTest.java`
Created `QualityControlStateUpdaterNGTest.java`
Created `QualityControlStateUpdater.java` to refactor out the state update plugin

`QualityControlAutoVetterNGTest.java` includes tests to do the following:
- testUpdateQualityControlStateWithNoGraph
- testGetInstance
- testInit
- testInitWithRefresh
- testAddRemoveObserver
- testAddRemoveListener
- testUpdateQualityControlState
- testGraphChangedNoGraph
- testGraphChangedWithGraph
- testObserver (Implementation of an interface for testing)
- testListener (Implementation of an interface for testing)

`QualityControlStateUpdaterNGTest.java` includes tests to do the following:
- testReadNoNodesNoAttributes
- testReadNoNodes
- testReadSelectedNodesWithRules
- testGetName

There are a few calls to `Thread.sleep()` which seemed to be necessary to get the GraphManager to return the current active graph correctly, every time. Sleeping made a delay that seemed to give 100% replicability.

Related work: #1023

PR #1076, opened 2021-06-01
QCV auto vetter unit tests fix

antares1470:
### Description of the Change

One of the QualityControlViewPane tests failed when run with the auto vetter tests due to its use of the lookup for grabbing QC Rules. This change executes the rules within the unit test to enable the test to pass both in isolation and when run with all the other tests.

Once merged, this should enable #1034 to be merged.

Fixes problem with unit test

### Benefits

Passes CI


aldebaran30701 (approved):

aldebaran30701:
Tested and seems to work without issue

PR #1093, opened 2021-06-07
Unit test guidelines

antares1470:
### Description of the Change

Added some unit testing guidelines. I don't by any means claim to be an expert on writing unit tests but this what I've learned to be helpful when writing unit tests for Constellation. 

It's helpful to have a standard to follow for unit tests which in turn helps improve consistency across unit tests.

### Benefits

It's helpful to have a standard to follow for unit tests which in turn helps improve consistency across unit tests.

### Possible Drawbacks

If not maintained, these could get out-of-date

### DIFF
--- a/UNIT_TEST_GUIDELINES.md
+++ b/UNIT_TEST_GUIDELINES.md
@@ -20,7 +20,7 @@ code in the following:
     }
 ```
     
-    - Doing this allows the code to pass in our CI which will otherwise hold up 
+Doing this allows the code to pass in our CI which will otherwise hold up 
 until it times out. The catch though is that the CI is passing the test 
 essentially by skipping it which is not ideal. As a result, this advice will 
 likely change once a better solution is found.
--- a/UNIT_TEST_GUIDELINES.md
+++ b/UNIT_TEST_GUIDELINES.md
@@ -19,6 +19,7 @@ code in the following:
         //insert test code here
     }
 ```
+    
     - Doing this allows the code to pass in our CI which will otherwise hold up 
 until it times out. The catch though is that the CI is passing the test 
 essentially by skipping it which is not ideal. As a result, this advice will 
--- a/UNIT_TEST_GUIDELINES.md
+++ b/UNIT_TEST_GUIDELINES.md
@@ -1,0 +1,47 @@
+# Unit Test Guidelines
+
+Here are a few helpful guidelines for writing unit tests for Constellation.
+
+## What To Include
+
+- Ideally, the unit tests you write for a function should cover any edge cases 
+that could be encountered when running it.
+
+- For functions that take a graph as input, you should include a test with a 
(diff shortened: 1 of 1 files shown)


Delphinus8821 (approved):

aldebaran30701:
Is it worth mentioning to use NGTest, and Mockito as the standard for testing?
Other than the above suggestion, it reads well and gets the point across

antares1470:
> 
> 
> Is it worth mentioning to use NGTest, and Mockito as the standard for testing?
> Other than the above suggestion, it reads well and gets the point across

Good pickup. I'll add that now

aldebaran30701 (approved):

Delphinus8821 (approved):

Auriga2 (approved):

arcturus2:
Awesome @antares1470.

I'm wondering if we can add something about using a StoreGraph instead of DualGraph's. Something like:

When you need to test with a graph, create a `StoreGraph` so that you can focus on the test and not have to deal with read/write locks. Testing of read/write locks is already covered elsewhere.

Also can we have one that talks about using the minimum amount of attributes to cover the test? Something like:

Create the minimum amount of attributes to satisfy your unit test. This helps explain exactly the minimum requirements the plugin needs to function. Also avoid creating unnecessary amount vertices and transactions. In some cases 3 vertices is sufficient to create the context for your test.

@antares1470 I'd also like to tap into unit vs integration but until we have a proper naming convention for it we can omit this for now.

antares1470:
> 
> 
> Awesome @antares1470.
> 
> I'm wondering if we can add something about using a StoreGraph instead of DualGraph's. Something like:
> 
> When you need to test with a graph, create a `StoreGraph` so that you can focus on the test and not have to deal with read/write locks. Testing of read/write locks is already covered elsewhere.
> 
> Also can we have one that talks about using the minimum amount of attributes to cover the test? Something like:
> 
> Create the minimum amount of attributes to satisfy your unit test. This helps explain exactly the minimum requirements the plugin needs to function. Also avoid creating unnecessary amount vertices and transactions. In some cases 3 vertices is sufficient to create the context for your test.
> 
> @antares1470 I'd also like to tap into unit vs integration but until we have a proper naming convention for it we can omit this for now.

Thanks @arcturus2. I'll add something about using StoreGraph.

Regarding your other suggestion, I already kind of cover this when explaining what a non-empty graph should compose of but maybe I can make some of your points more explicit

GammaVel (approved):
Looks great to me!  Thanks for putting these together. A couple of things - just wondering if you wanted to mention that it is good practice to make the unit test fail to see the test is working correctly. And each new part of code it is expected to have unit tests associated with it as part of Test Driven Development?

sol695510 (approved):

arcturus2 (approved):
amazing stuff thank you @antares1470

PR #1095, opened 2021-06-07
Fix exceptions in Delimited File importer

Auriga2:
### Description of the Change

1. Fixed the exception when removing files in the Delimited File Importer.
image

2. Fixed the exception when the `Datetime` attribute is dragged and dropped onto a non Datetime type column and continue to import. Now it'll prevent the user from dropping an incorrect attribute, with an error message.
image

### Alternate Designs
For 2, I initially added a catch block in SimpleEditPlugin, but felt like it could affect many other plugins and the error message thrown there was a bit confusing. Also it's much better to prevent it rather than handling.

Better UX

### Benefits
Better UX

### Possible Drawbacks
N/A

### Applicable Issues
https://github.com/constellation-app/constellation/pull/1074


antares1470 (approved):

antares1470 (changes requested):
Actually, if you're able to add a changelog entry for the validate function return type changing, that would be good

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Constellation Changes
 
+## Changes in June 2021
+
+-    `ImportTableColumn.validate` and `importexport/RunPane.validate` 
+	functions return type changed from `void` to `boolean`.
```
antares1470, line 6: If you can slightly rephrase this to start with "Changed" or "Updated" (just to align it with the way other entries are written), that would be great

antares1470 (changes requested):
Other than the suggestion and resolving the merge conflict, this looks good to go

antares1470 (approved):

PR #1104, opened 2021-06-08
updated-ISSUE_TEMPLATE.md

danger-ahead:

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -8,7 +8,7 @@ maintainers' discretion.
 * Have you read Constellation's Code of Conduct? By filing an issue, you are
 expected to comply with it, including treating everyone with respect:
 https://github.com/constellation-app/constellation/blob/master/CODE_OF_CONDUCT.md
-
+* The "Steps to Reproduce" should outline the steps required to reproduce the issue. The rest should filled out accordingly.
 -->
 
 ### Prerequisites


aldebaran30701:
I believe the issue template should include a message hint in the sections of expected and actual behaviour. The hint should specify that if applicable, include the time something should execute within, and specify the actual time something took to execute. 
@danger-ahead does that sound like an improvement you could add to the PR ?

danger-ahead:
> I believe the issue template should include a message hint in the sections of expected and actual behaviour. The hint should specify that if applicable, include the time something should execute within, and specify the actual time something took to execute.
> @danger-ahead does that sound like an improvement you could add to the PR ?

@aldebaran30701 will do that

ISSUE_TEMPLATE.md
```diff
@@ -8,7 +8,7 @@ maintainers' discretion.
 * Have you read Constellation's Code of Conduct? By filing an issue, you are
 expected to comply with it, including treating everyone with respect:
 https://github.com/constellation-app/constellation/blob/master/CODE_OF_CONDUCT.md
-
+* The "Steps to Reproduce" should outline the steps required to reproduce the issue. The rest should filled out accordingly.
```
antares1470, line 11: I would possibly move this point to immediately after the first point in the Requirements section as I think this follows on from that point

antares1470 (changes requested):
To extend @aldebaran30701's comment, I'd suggest a comment including that the Expected and Actual behaviour should be measureable (which is more applicable for more issues than just time-related ones)

antares1470:
Closing as this has been done in #1139

Issue #756, opened 2020-08-03
Banners not rendering - Remove Banners Implementation


formalhaut69:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Banners should render on the top and bottom of the graph when the user requests it through the dialog window. 

### Steps to Reproduce
![image](https://user-images.githubusercontent.com/58677762/89153478-16acd600-d5a4-11ea-8820-2d9a2edb60a3.png)

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information
According to @arcturus2 The renderer framework was updated 2 years ago and this is likely to be the cause of this issue. This issue will likely require some level of experience with OpenGL

Nova-2119:
Had a look,. it looks to me like all of the code to render the banners is just plain missing.
Checked the original public release of Constellation and the Banners don't work there, so this problem has been around for a long time and there is no working version to compare to.

Given it was only ever experimental, I feel like it is alot of effort to fix for something experimental that has never worked in modern Constellation. I think pretty much every other bug ticket is going to provide better value/hour of work, so if its okay @antares1470 @arcturus2 i think we should move this to the bottom of the priorities for this bug sprint and I'll move onto something else?

arcturus2:
yeah OK @Nova-2119 lets work on the other bugs first then and come back to this later. I'd still be keen to get this done but I'll take it off the list then.

aldebaran30701:
This is getting added to the [Brainstorming Ideas](https://github.com/constellation-app/constellation/wiki/Brainstorm-Ideas) list as it has been broken for some time. 

**To close this ticket, the removal of Banners implementation has to be done in one commit so it is easy to re-add or review in the future. Once merged, this ticket can be closed.**

PR #1110, opened 2021-06-11
Remove Banner Implementation

Delphinus8821:
### Description of the Change
The banner implementation has been removed as it currently does not work and has not worked in any modern version of Constellation. Removing it through this PR allows for this code to be found easily if it is wanted to be reimplemented in the future.

### Alternate Designs
N/A

The banners were decided to be removed as they have not worked previously and the idea of the banners will be added to the 
Brainstorming Ideas list.

### Possible Drawbacks
None realised

### Applicable Issues
#756


antares1470 (approved):

arcturus2 (approved):
Perfect thank you @Delphinus8821

Issue #1002, opened 2021-03-26
Layer queries not working on Type attribute


aldebaran30701:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Querying a layer for the type throws an error. the error seems to stem from the querying section as opposed to the layers view module. 

### Steps to Reproduce
1. Open layers view
2. Query a layer using Type == 'Communication'

**Expected behaviour:** set visible the nodes which meet that criteria

**Actual behaviour:** Error cannot find EQUALS operator

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

aldebaran30701:
To add - Make a few new layers using hotkeys without a graph. Then go and set one of those layers to on with the hotkey. Null pointer error. This may be a simple null check.
Also to recheck C-A-L for new layer instead of CA-L. This may be the source of an issue where the right alt/ctrl button does not get triggered in a hotkey action whereas the left ctrl alt does.

aldebaran30701:
I'd like to have included unit tests with this fix - but the querying code is very complex and would turn into more of an integration test for the bi-functions. Having a think about the way it could be done currently.

PR #1113, opened 2021-06-17
🐛 Layers view null pointer fix

aldebaran30701:
### Description of the Change
A few noted null pointers to finish off #1002 
Returning a boolean as false instead of null made sense.

Tested by having no graph open.
Use the hotkeys ctrl + alt + L
ctrl + alt +1

The error should occur at that point.

With the fix, there is no null pointer and no drawbacks because nothing should  be occurring without a graph or the Layers View opened.

### DIFF
--- a/CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
+++ b/CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
@@ -188,7 +188,7 @@ public class LayersViewController {
     }
 
     public boolean getParentVisibility() {
-        return parent != null ? parent.getVisibility() : false;
+        return parent != null && parent.getVisibility();
     }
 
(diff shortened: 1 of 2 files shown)


CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewController.java
```diff
@@ -182,10 +188,13 @@ public void readState() {
     }
 
     public boolean getParentVisibility() {
-        return parent != null ? parent.getVisibility() : null;
+        return parent != null ? parent.getVisibility() : false;
```
antares1470, line 191: this can be changed to `return parent != null && parent.getVisibility()`

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):

PR #957, opened 2021-02-02
Conversation search option

sol695510:
### Description of the Change
All instances of the text typed in the search bar now appear in the entire conversation. 'Select on Graph' and 'Copy' context menu functionality has been restored.

### Alternate Designs
N/A

### Why Should This Be In Core?
N/A

### Benefits
Mentioned context menu functionality has been restored and the results generated by the search are more meaningful.

### Possible Drawbacks
N/A

### Verification Process
Manually verified that the context menu options work as expected and tried various combinations of text in the search bar with the same conversation in an attempt to find any faults.

### Applicable Issues
#887 
#939 - resolved in this PR on this branch.

Issue #887, opened 2020-11-05
Search conversations


antares1470:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Would it be possible to add a search function to the conversation view? Similar to excel in that it would jump to a bubble matching keywords/characters chosen.

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

mimosa2:
A a branch called "conversation_search_option" is created with the work @mimosa2 has already done on this ticket. There is one more task on this which @mimosa2 couldn't complete.  If anyone can continue on this to complete that task would be great.
The issue is when search found a string in a bubble which is hidden in the conversation view (i.e need to scroll down to view the bubble). It would be nice if the bubble set focus (scroll up) itself without user to scroll down to view the searched string.

sol695510:
![convo1](https://user-images.githubusercontent.com/58677774/106085439-1c106500-6170-11eb-892d-e118271e46fe.gif)
^Current context menu functionality. Previously available "Select All" has been omitted since the current type of JavaFX region (TextFlow) being used is not selectable/highlightable, so now "Copy" simply copies all text on a selected conversation bubble to the system clipboard.

![convo2](https://user-images.githubusercontent.com/58677774/106085603-6eea1c80-6170-11eb-8627-cd8d81d212a4.gif)
^Current conversation search functionality. All instances of searched text are shown via styling the text with the JavaFX region TextFlow.

sol695510:
@antares1470 @mimosa2 Can this ticket be closed in favour of moving ticket #958 onto the board to replace it? Although they ask for the same feature, this ticket refers to the branch for this PR #957, which should be closed soon.

mimosa2:
#957 Can be closed and #958 can be moved into the board.

antares1470:
> @antares1470 @mimosa2 Can this ticket be closed in favour of moving ticket #958 onto the board to replace it? Although they ask for the same feature, this ticket refers to the branch for this PR #957, which should be closed soon.

[...]

PR #1114, opened 2021-06-17
fix refresh of conversation counter

aldebaran30701:
### Description of the Change
This adds a static method to refresh the `ConversationBox` just before an update.
Prior to this, the counter would continue to jump up higher and higher whenever an update would occur.

The `updateComponent` currently triggers on a list change event. Then the view responds by changing only the singular conversations within that list. This is troublesome to reset a count as we don't have access to the iterator which is looping and updating the results. eg. the below is an ideal situation to reset the `foundCount` before the loop
```
foundCount = 0
for conversation in conversations :
    update(conversation)
```

What we have is as below. IE. Some external data structure looping all changes and just providing a means to update.
```
@Override
update(conversation){
    setText(Conversation.getText())
    foundCount+=1
}
```

### Possible Drawbacks
Potential downsides to this is that it couples the classes more. A fix could be to add another listener that would be responsible for only refreshing the view just before the` update(conversation) `call was made. `(Conversation.java -> update() -> Line 558)`

### Applicable Issues
#887 
PR #957

### DIFF
--- a/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/Conversation.java
+++ b/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/Conversation.java
@@ -555,6 +555,7 @@ public class Conversation {
 
         @Override
         protected boolean update(GraphReadMethods graph) {
+            ConversationBox.reset();
             if (resultMessages != null) {
                 resultMessages.setAll(visibleMessages);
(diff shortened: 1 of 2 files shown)


antares1470 (approved):

sol695510:
The changes in this PR have been added in #957.

Issue #970, opened 2021-02-14
Layers View Usability Improvements


aldebaran30701:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

From review of the layers view, it appears that there are some refinements that need to be made so it is user friendly.

1. The ability to delete a layer. (Reduces the clutter of the view when many layers are not needed)
2. Make keyboard shortcuts more obvious. (deselecting all layers (ctrl+alt+d) and toggling layers (ctrl+alt+1-9) Creating a new layer (Ctrl+alt+L) Could possibly be tooltips, hint panel ?)
3. Increase Query Error Visibility. (Direct quote from https://github.com/constellation-app/constellation/issues/223#issuecomment-762569337 - The query language is a little obtuse. My first query (transactions - Identifier startswith 2) caused a hash error popup as i'd not surrounded a number in quotation marks. There were also queries i just couldn't get to work (dim == 'true' gave me "ExpressionException: Unable to find implementation of operator: EQUALS"). There should be validation applied before attempting to execute the query, and the result of that **should be visible to the UI (a red box to say my query is broken)**.)

### Steps to Reproduce

N/A

### Additional Information

#223

GammaVel:
Thanks for making all the improvements @aldebaran30701, I am enjoying how the new layers are working and it is great I can delete layers. There was an issue with deleting layers when no graph was open, I have submitted a ticket for this #1334.

PR #1111, opened 2021-06-16
🐛 Fix layers view querying by type

aldebaran30701:
### Description of the Change
This PR adds bi-functions to the registry which will allow comparison between type objects and the strings it can be compared with.

Also included in this PR is the creation of a dialog box when the user tries to enable a layer with an invalid query. This is in place of throwing an exception to the user. Hopefully this will be less scary and viewed as user error to prompt review of the query passed in.

### Alternate Designs
Adding red text to the UI of Layers View. This may be a future task as it seemed to want to not refresh the UI at the correct time.

### Why Should This Be In Core?
Adds extra query ability to layers view

### Benefits
This will increase the usability of the query interface in layers view.

### Possible Drawbacks
Possibility of popup appearing on a different display to the main user window.

### Verification Process
Ran queries of type attributes and tried to query every attribute type.

### Applicable Issues
#1002

PR #1115, opened 2021-06-22
✅ Expressions tests and Layers View Improvements

aldebaran30701:
### Description of the Change
This should be merged in AFTER #1002 
This PR adds:
- ExpressionParserNGTest.java which is responsible for testing many of the core attributes within various expression strings. It purely tests that the parser returns the correct order and expression types to the caller. I've added checks for invalid queries also. So far, every operator has been tested in the cases, It was required that a flag be set within the test cases to ensure no pop-up was activated during test execution. It should run as intended normally.
- I've fixed a bug with the comparison of expressions which is used within the test cases (AssertEquals(...) and so on). It was to do with equals and hashcode implementations not existing.
- Updated the layers-view.html documentation page with extra information and better formatting.
- Relocated the Expressions specific documentation to the Graph framework module - adding a link from Layers documentation to help assist the user to find the correct help documentation between modules.
- Added tooltips to the buttons of layers view to show the hotkeys for that action.
- Added word operators for most common operators. (equals, and, xor, subtract etc...)
- Implemented an initial parse check on the entry of text into the layers view. This will check that there are no illlegal characters in the query, and that it contains pairs of quotes. Validation will be expensive if the check was to run on the whole graph to see if any elements would be returned.

A red box appears around the text field when it is not valid.
I wanted to include some sort of message that wouldn't interrupt the flow of the Layers View, so I included a context menu to appear. This goes away with a click away.
In normal use the user should really only be concerned about one box, and having multiple context menus appear on different inputs shouldn't be an issue. 
I'm okay to get rid of the context menu if it seems intrusive - keen to hear peoples' thoughts on this.

image

### Alternate Designs
N/A

Reduces bugs, increases documentation and testing.

### Benefits
Reduces bugs, increases documentation and testing.

### Possible Drawbacks
None realized

### Verification Process
Opening graphs, toggling layers with and without manual entries and query entries. 
Purposefully adding incorrect terms to the UI. Such as adding an extra quote, missing an end quote.

### Applicable Issues
Once approved, should solve #970 apart from deletion of a layer.


arcturus2:
@aldebaran30701 I'm yet to test this but loving the changes so far. Something I wanted to ask now that you are adding UX things, can the boxes in the expressions section be increased in height by say x5 times? I find that as soon as the scroll bar appears it is really hard to see or make changes.

aldebaran30701:
@arcturus2 I agree - I haven't thought of a good way to counteract this yet.
Here is with 100px height. I tried without wrapping text and it was just a line of text with a scroll bar still. This is with text wrap on.
image

I'd say it's harder to read - I tried using the panel which is similar to the timeline view - and it appears to display better.
image

Here is multiple lines of text - vertical scroll bar appears when more area is needed.
image

An issue I see with the timeline-esque view is that the lower buttons are hidden and many of the other layers are not visible. I can 'float' the other buttons but the layers not being visible is another story

aldebaran30701:
I did a quick mock-up of a UI if we were looking to keep it on the vertical bar - Needs some improvement but it gives more space.
image

arcturus2:
Nice work @aldebaran30701 with the designs.

With regards to putting the layers on the bottom section like the timeline/table view, if you want the buttons to stay they you could look at how the Data Access View does it to "anchor" the buttons to the top (or you could do bottom). It's using an `AnchorPane`.

Regarding the second option with the Mock Up UI I like that too and I think that could work. One suggestion there is that you could introduce a scroll bar for the list of 15 layers visible there to maybe 5 and then use the rest of the screen real-estate to have a section for the "Vertex Query" going horizontally using the full width and have the "Transaction Query" block underneath?

Keep the design you have but if you can put this as another variation I can try to find some people to do a UX study on showing them different mock-ups and see what they think?

Also this can be done in a separate branch so it does not block this stuff getting out. That said I found an issue which I will post in a separate comment.

arcturus2:
@aldebaran30701 so I actually merged both PRs into my own local branch to test these 2 together and I got an error. Can you confirm>

Reproducer is I created 4 nodes in edit mode and in the layer 1 Vertex query i put in `Type == 'Unknown'`.

I got the following exception:

```
au.gov.asd.tac.constellation.graph.value.expression.ExpressionException: Unable to find implementation of operator: EQUALS (au.gov.asd.tac.constellation.graph.attribute.AbstractObjectAttributeDescription$$Lambda$1235/0x0000000800ad4840, au.gov.asd.tac.constellation.graph.value.expression.ExpressionCompiler$$Lambda$1244/0x0000000800add840)
	at au.gov.asd.tac.constellation.graph.value.expression.ExpressionCompiler.compileSequenceExpression(ExpressionCompiler.java:113)
	at au.gov.asd.tac.constellation.views.layers.query.Query.compile(Query.java:89)
	at au.gov.asd.tac.constellation.views.layers.query.BitMaskQuery.update(BitMaskQuery.java:91)
	at au.gov.asd.tac.constellation.views.layers.query.BitMaskQueryCollection.update(BitMaskQueryCollection.java:103)
	at au.gov.asd.tac.constellation.views.layers.query.BitMaskQueryCollection.updateBitMasks(BitMaskQueryCollection.java:121)
	at au.gov.asd.tac.constellation.views.layers.LayersViewController$UpdateQueryPlugin.edit(LayersViewController.java:376)
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:207)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:115)
```

arcturus2 (changes requested):
Please see my last comment of an issue I got.

arcturus2:
I think we can have a ticket to discuss the UI and come up with proposals if you like but here is a taste of what I was thinking:

image

I'm just using the free mockplus.com site and here is a share link https://app.mockplus.com/app/share-b3f2b17418d4af22c5daf78e28663d18share-EyQFB_dpBbD/rp

aldebaran30701:
I had the same thoughts with the way you did that mockup.
Regarding the error you found - I actually based this branch off the branch that fixed the type issue. Maybe merging it back into this branch caused an issue?
Come to think of it - this PR could be the only one merged in, and the types PR #1111 just closed. I've got all changes in this current PR.

I managed to successfully query all default types in Constellation.
I've also removed the context menu, and added a label down the bottom which toggles it's visibility with a static query error message.
(don't mind the small text - I needed more real estate)
image

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/LayersViewPane.java
```diff
@@ -283,12 +332,26 @@ private synchronized void syncLayers() {
                 case 1:
                     visible = ((CheckBox) item).isSelected();
                     break;
-                case 2:
+                case 2: {
```
antares1470, line 335: omit the braces from the case blocks as they're not necessary

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/docs/layers-view.html
```diff
@@ -7,6 +7,16 @@
...
+        <p>
+            The Layers View holds a collection of Layers. Each Layer can represent a static set of elements, or a dynamically 
+            calculated set of elements which match a query criteria. The query structure is based on the expressions framework.
+            <br/>
+            Each layer has a visibility checkbox which determines if that layer is toggled. All elements are on the base layer 0.
+            When one or more check boxes are selected, only the elements that are members of those layers are visibly displayed. 
+            All other elements will have their <strong>visibility</strong> attribute value toggled to 0.0 (invisible).
+            <br/>
+            Arrangements on graphs with a visual schema will be ran on the base layer 0. 
+        </p>
```
antares1470, line 19: Are you able to move this paragraph below the table?

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/docs/layers-view.html
```diff
@@ -7,6 +7,16 @@
     </head>
     <body>
         <h1>Layers View</h1>
+        <p>
+            The Layers View holds a collection of Layers. Each Layer can represent a static set of elements, or a dynamically 
+            calculated set of elements which match a query criteria. The query structure is based on the expressions framework.
+            <br/>
+            Each layer has a visibility checkbox which determines if that layer is toggled. All elements are on the base layer 0.
+            When one or more check boxes are selected, only the elements that are members of those layers are visibly displayed. 
+            All other elements will have their <strong>visibility</strong> attribute value toggled to 0.0 (invisible).
```
antares1470, line 16: With the updates I've made to the help pages, I've actually moved away from using the strong tag just to avoid clashes with title names (which are also bold). I've moved to using double-quotes for now so if you're able to do that as well, that would be great

antares1470 (changes requested):

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/docs/layers-view.html
aldebaran30701: How does this look? Should it be below the image also?
image

antares1470: I think I'm ok with the paragraph being above the image as it aligns with what has been done in some of the other pages

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/docs/layers-view.html
```diff
@@ -11,16 +11,16 @@ <h1>Layers View</h1>
             <caption>Layers View Actions</caption>
             <tr>
                 <th scope="col">
-                    <strong>Constellation Action</strong>
+                    "Constellation Action"
```
antares1470, line 14: I think the double quotes can be removed in the table headers (my bad for not explaining for not explaining what I did with strong tags well enough)

CoreLayersView/src/au/gov/asd/tac/constellation/views/layers/docs/layers-view.html
```diff
@@ -61,180 +61,73 @@ <h1>Layers View</h1>
...
+            A query layer is a dynamic layer that shows the elements represented by the specified 
+            Vertex and Transaction queries. Query layers are dynamically recalculated upon changes to those attributes being queried. 
+            This is to ensure the accuracy of the shown layer. e.g. Layer 1 has a vertex query described as Label == 'Vertex #0'. This will show the 
+            vertex with that label. If you then change Vertex #0's label to something like 'V#0', the query will note the 
+            change and recalculate for the whole graph. 
         </p>
-        <p>e.g. Label == 'Vertex #0&lt;Unknown&gt;' || Label == 'Vertex #1&lt;Unknown&gt;' shows elements that satisfy either constraint.</p>
-        <p>e.g. Label == 'Vertex #0&lt;Unknown&gt;' &AMP; nradius == '1.5' shows elements that satisfy both constraints.</p>
-    </body>
+        <p>"NOTE: Layers View does not allow for the use of the assignment operator = or keyword assign."</p>
```
antares1470, line 130: Can also remove the double quotes around the Note here (once again my bad)

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/docs/expressions.html
```diff
@@ -0,0 +1,225 @@
...
+        <title>The Constellation Expressions Framework</title>
+        <link rel="stylesheet" href="nbdocs://au.gov.asd.tac.constellation.preferences/au/gov/asd/tac/constellation/preferences/constellation.css" type="text/css">
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    </head>
+    <body>
+        <h1>The Constellation Expressions Framework</h1>
+        <p>
+            The expressions framework is a structure which can be used to query or assign the Vertex and Transaction attributes within a graph.
+            The framework will work on all attributes which <strong>do not</strong> include a restricted character within it's AttributeName. An expression can 
+            return <strong>0..*</strong> (none or many) matches.
```
antares1470, line 13: - remove strong tags
- replace "return 0..* (none or many)" with simply "returns zero or more"
- surround AttributeName in double-quotes

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/docs/expressions.html
```diff
@@ -0,0 +1,225 @@
...
+            used to gain the flexibility necessary. Users will be responsible for forming queries which adhere to the below rules.
+        </p>
+        
+        <h2>Forming a Query Expression</h2>
+        <p> A valid query for a vertex or transaction takes the form of <strong>[AttributeName] [Operator] [value]</strong>. </p>
+        <p>The <strong>AttributeName</strong> used will not include any restricted characters, and will be case sensitive.</p>
+        <p>The <strong>AttributeName</strong> will be represented as plain, unquoted text.</p>
+        <p>The <strong>Operator</strong> must be separated by atleast one whitespace character.</p>
+        <p>The <strong>Operator</strong> must be valid and from the list specified below.</p>
+        <p>The <strong>Value</strong> must be quoted using single('value') or double quotes("value").</p>
```
antares1470, line 28: - replace all the strong tags here with double-quotes
- for all bar the first paragraph, I reckon convert the paragraphs to an unordered list

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/docs/expressions.html
```diff
@@ -0,0 +1,225 @@
...
+        <br/>
+        <p>
+            How to form an expression to find a vertex with the <strong>Label</strong> attribute matching <strong>"Vertex #1&lt;Unknown&gt;"</strong>.
```
antares1470, line 31: bar the table below, I think you can swap all the remaining strong tags with double quotes

antares1470 (changes requested):

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/docs/expressions.html
aldebaran30701: So leave the strong tags within the table headers?

antares1470: I've discovered that table headers are also bold so I don't think they can be removed (as they're not actually adding anything extra)

CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/docs/expressions.html
```diff
@@ -9,8 +9,8 @@
         <h1>The Constellation Expressions Framework</h1>
         <p>
             The expressions framework is a structure which can be used to query or assign the Vertex and Transaction attributes within a graph.
-            The framework will work on all attributes which <strong>do not</strong> include a restricted character within it's AttributeName. An expression can 
-            return <strong>0..*</strong> (none or many) matches.
+            The framework will work on all attributes which do not include a restricted character within it's "AttributeName". An expression can
```
antares1470, line 12: its*

antares1470 (changes requested):
Missed that typo in my last review. The only thing I can add is I think it would be good to be able to specify the type without having to specify the full hierarchy. I don't think it's intuitive to have to specify "Type == 'Online Identifier.Email'" rather than "Type == 'Email'" (especially since you don't need to specify the hierarchy in other places in the application)

aldebaran30701:
>The only thing I can add is I think it would be good to be able to specify the type without having to specify the full hierarchy. I don't think it's intuitive to have to specify "Type == 'Online Identifier.Email'" rather than "Type == 'Email'" (especially since you don't need to specify the hierarchy in other places in the application)


@antares1470 The way the expressions framework is extended, it only allows behaviour changes on the operator parsed in, and the type of the attribute (integer, long, double, object etc.) So for the Type attribute, it is an AbstractObjectAttributeDescription. This just uses the implemented toString() method to compare the values of the Type. 

One way forward is to override the toString() method within SchemaElementType, or more specifically the SchemaVertexType & SchemaTransactionType. This, depending on it's usage will cause behavioral changes throughout the whole application.
The file output doesn't seem to use the hierarchy toString(). Which means maybe a change won't break old files.
The other thing I noticed was that the iconmanager uses the toString() of the type to find it's icon. Currently only the unknown Icon seems to have an icon registered with it's type name.

aldebaran30701:
@antares1470 I've found a way to fix this, there is a method which creates the readable object for the expressions framework. This returns the object which usually just calls the default toString. Implementing it within the VertexTypeAttributeDescription allowed me to cast it to a schemaVertextType and then call getname(). commit coming soon

arcturus2:
@aldebaran30701 besides the feedback on the Type string this is working really well. The issues I found earlier are sorted. This PR has come a long way and a lot of effort has gone into this. I can't wait to take this off experimental probably after a GUI update we've been chatting about here.

Also top work on the unit tests here - a stand out PR. Love your work!

antares1470 (approved):

antares1470:
Other than resolving merge conflicts, this looks good to go

arcturus2 (approved):

aldebaran30701:
@antares1470 Should be no conflicts now - Looks good to merge

Issue #573, opened 2020-06-11
Allow import of delimited files that don't contain column headings


serpens24:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

Consider the following csv data:
![image](https://user-images.githubusercontent.com/58677108/84330244-0a208880-abc6-11ea-8bff-3519113a9f5d.png)
Its clear the first row represents headings, while subsequent rows represent data. When parsed via the importer the follwoing is seen:
![image](https://user-images.githubusercontent.com/58677108/84330386-684d6b80-abc6-11ea-8a49-6662d75f3a67.png)

Consider now the same csv but without the header row:
![image](https://user-images.githubusercontent.com/58677108/84330435-8a46ee00-abc6-11ea-926a-73d6349e8fef.png)

Attempts to import this takes the first row and uses it as headings:
![image](https://user-images.githubusercontent.com/58677108/84330473-a2b70880-abc6-11ea-8e9e-56d431464951.png)

It would be useful to be able to flick a switch to indicate there is no header row. This could sit alongside the "Initialize With Schema" checkbox.

### Steps to Reproduce

1. Import CSV file that doesn't contain headers, ie:
![image](https://user-images.githubusercontent.com/58677108/84330435-8a46ee00-abc6-11ea-926a-73d6349e8fef.png)
2. Confirm the first row of real data is treated as headers

**Expected behaviour:** 
User has ability to indicate there is no header row, in which case, column headers are constructed, ie:
![image](https://user-images.githubusercontent.com/58677108/84330671-2cff6c80-abc7-11ea-9821-5a53c649f510.png)

**Actual behaviour:** 
First row of data is treated as column headers, ie:
![image](https://user-images.githubusercontent.com/58677108/84330473-a2b70880-abc6-11ea-8e9e-56d431464951.png)

**Reproduces how often:**
100%

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

GammaVel:
Thanks for making all the improvements here :-)

PR #1132, opened 2021-07-02
Delimited File Importer - Option to insert column headers

Auriga2:
### Description of the Change
An option `Insert Column Headers` is added in the Delimited File Importer source pane. When importing from delimited files that don't contain column headings, the option can be enabled to insert column headers as `Col1`, `Col2`, `Col3` etc.

### Alternate Designs
n/a

Better UX, as without this, the first row will be skipped as the column headers.

### Benefits
It can now import from files with no column headers.

### Possible Drawbacks
n/a

### Verification Process

1. Create a CSV file that doesn't contain headers, e.g.:
5,6,7,2009-12-05 23:59:59,2009-12-1
1,1,1,2009-12-15 23:59:59,2009-12-12
a,a,a,2009-12-25 23:59:59,2009-12-31

2. Enable the check box and import it.
3. Confirm the columns are inserted and the first row of real data is not lost.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/573


antares1470 (changes requested):
Did some testing of this and made some observations:
- I noted that you had to check the column headers checkbox before you added the file. I think it would be good if possible to be able to do this in either order (and the UI reflect accordingly)
- Related to above point but it appears the tickbox is applied when I add a file and that is then applied to all the files that are currently added. In other words, if I add a file with column headers checked and then add another file with column headers unchecked, the result shows both files without column headers added (including the one I added with column headers checked). It's possible though this issue is related to the greater problem of how importing multiple files is currently handled.

Auriga2:
This sounds good as it will reduce one click in the workflow. But it will also double the options in the File Parser drop down (to 12), and this option would be used only in rare cases I presume. (In terms of coding, it also requires a new set of ImportFileParser classes (6 new) added.)
image


So I still prefer to move the section up, or at least move the new option into the Options list at top left corner? The former will also fulfill the UX Design Review suggestion below, and is a simple fix:
> 3.Based on previous point, consider swapping ‘Destination’ and ‘Files’ sections to swap position under ‘Source’ categories

image

antares1470:
> We can also rename the checkbox to something like "First row of the file is header row" and keep it checked by default?

@Auriga2 that should be fine. To avoid the text being too large though, instead of "First row of the file is header row" maybe something like "File(s) has headers".

Side note: With your proposed solution, the Destination drop-down sits right below the Source pane title which looks a little funny. In terms of flow, I think it makes more sense to pick your destination after you load the file rather than before. Maybe you can keep the other options above the Files part and move the Destination drop-down to below the files part(?)

cygnus-x-1:
This is looking good from the screenshots. I'd suggest renaming "Import File Parser" to simply "File Parser", and "Insert Column Headers" to "Include Headers" (on by default, which would assume the file has headers). I'm going to check out the branch and do a bit of testing... I'll let you know if I find anything.

cygnus-x-1:
Some feedback from testing:

- You should be able to toggle headers on and off after the file has been loaded into the preview pane. As it is now, I needed to remove all files, then toggle the headers option, then add the files back to see the change.
- I think I like your suggestion about moving the "Source" options bar to the top as it better matches the workflow; you're forced to focus on what graph you're populating up front and what file type you're importing... particularly important as you can't change that after adding files.

And some broader feedback about the Delimited File Importer...
- The "Tab Separated" file parser should probably be called "TSV".
- The "SQLite" file parser is somewhat redundant now that the JDBC Importer exists.
- I'd suggest renaming "Import from Delimited File" to simply "Import from File" as the less technical terminology is more approachable. Similarly, rename "Import from JDBC" to "Import from Database".

arcturus2:
@Auriga2 I gave the PR a test and the newer UI in the screenshots look better and the feedback from others is also positive. So keen to test the PR again based on the changes you decide to make.

The additional suggestions from @cygnus-x-1 are great but if you are not able to do them in this PR then happy for that to be done separately later on if it could be documented by either @cygnus-x-1 or yourself. If you have time to do it all in this PR then also great and fine - will leave that up to you.

Auriga2:
Yeah good feedback. Thanks everyone.  I'll try to fix everything here.

Auriga2:
"File(s) Has Headers" or "Include Headers"? A combined "Files Include Headers" sounds better IMO. 

With regards to moving the Destination graph selection, the comments above are contradictory. In fact, in the current version, "Destination" sits in the Source pane. Moving it beloa the Files section wll do the same. Ideally we should move it to a new Pane. Or as a simple fix, can we keep it at the top and rename "Source" to "Source and Destination" or "Import Details" or ""Details" or something similar?

Shall we remove "SQLite" file parser?

arcturus2:
@Auriga2 I don't mind the quick fix of renaming the section to "Source and Destination". Ideally it would be its own section but not essential.

>Shall we remove "SQLite" file parser?
OK.

antares1470 (changes requested):
Minor thing but the panes themselves are still titled "From Delimited File..." and "From JDBC..." despite the name changes occurring elsewhere so it would be good to fix that.

Auriga2:
@antares1470 I cannot find where to fix this, it doesn't pop up with a string search.

antares1470:
> 
> @antares1470 I cannot find where to fix this, it doesn't pop up with a string search.

So it turns out I missed something when I built the branch. After doing so, the tab names presented as expected

antares1470 (approved):

arcturus2 (approved):

GammaVel:
hi @Auriga2, is this the issue which you are using to remove the colon after the E.g  ? If so, I have noticed this Text is also in this hints section too. It would be good to update both at the same time. 


image

Auriga2:
Yes, I have fixed all occurrences, looks like you don't have these changes yet.

Issue #1118, opened 2021-06-23
Make Search in Delimited File + JDBC Importers more obvious


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

There is a search functionality in the Delimited File and JDBC Importers in the preview section. This search bar is hard to see though as it looks more like a section separator than a search bar. The placeholder text doesn't help either (simply 'Filter').

### Steps to Reproduce

1. Open Delimited File Importer
2. Look for the search bar

**Expected behaviour:** Search bar location should be clear and obvious

**Actual behaviour:** Can't see where the search bar is 

**Reproduces how often:** 99% (I'll give the benefit of the doubt to the 1% with really good eye sight)

### Additional Information

This is where the search box is for reference:
![DelimitedSearchBox](https://user-images.githubusercontent.com/62311243/123178404-7668e400-d4ca-11eb-99b6-b840b9357503.png)

Could be useful to refer to the Data Access View for useful placeholder text

Auriga2:
Handles in https://github.com/constellation-app/constellation/pull/1124 So this can be closed.

antares1470:
I'd probably wait until that PR has been merged in before closing this so I'll re-open it for the time being.

antares1470:
![delimitedImportSearchPopup](https://user-images.githubusercontent.com/62311243/123564153-8f8bd080-d7fb-11eb-8c50-50f22d777e14.gif)

This is another not so obvious search bar which I think should be addressed as a part of this ticket (not even visible until you start typing)

Auriga2:
So this is the filter mentioned in the Hints section, which I have to add back.

GammaVel:
Hi @Auriga2. Thank you for making the filter more obvious now. Just a slight adjustment please. 

Can you please remove the ":" from the search filter under run - 
![image](https://user-images.githubusercontent.com/83679004/126254273-4e0ce4bd-06e6-4a5a-97fd-dfed30f23bb9.png)

[...]

PR #1133, opened 2021-07-07
Set the Attributes filter visible in JDBC and delimited file importers

Auriga2:
### Description of the Change
Attributes filter in the in JDBC importer and the delimited file importer is now always visible and is also consistent with the other filter in the Configuration pane.

### Alternate Designs
n/a

Better UX

### Benefits
Better UX

### Possible Drawbacks
n/a

### Verification Process

Check the `Attributes filter` in the in `JDBC importer` and the `Delimited file importer`. It should now always be visible and should also be consistent with the other filter in the Configuration pane.
### Applicable Issues
https://github.com/constellation-app/constellation/issues/1118


Auriga2:
I added my changes on top of `feature/JDBC_UX_Improvements` branch which is not merged into master yet, hence the long list above. Only the last commit is relevant here. Let me know if it should be changed.

antares1470:
> 
> 
> I added my changes on top of `feature/JDBC_UX_Improvements` branch which is not merged into master yet, hence the long list above. Only the last commit is relevant here. Let me know if it should be changed.

As a general rule of thumb, I would avoid creating branches off anything other than master unless your intention is to merge those changes into said branch (especially if a PR for that branch hasn't been merged yet). That said, I recently merged the PR related to your branch so it should be alright now (though it appears there are some merge conflicts as a result)

antares1470 (approved):

arcturus2 (approved):

Issue #50, opened 2019-07-26
Increase Error visibility


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a user
I want error information to be more visible
so that constellation doesn't seem to quietly fail

Sometimes when errors occur during plugin execution, there is no pop up; rather a small icon on the bottom right so the user does not know there was an error.

antares1470:
`ConstellationDialog` might be something to look at if we go down the dialog box path

aldebaran30701:
I've had a decent look into the way Constellation handles its exceptions. Currently, I've got a PR which needs a little more work to get it over the line, but this PR increases some consistency of the error handling capabilities.

The fundamental issue with error logging to the user is that if the unexpected exception is not thrown within code running on the default plugin environment, we cannot catch it and handle it (display to user).

**Run Constellation in debug through Netbeans**
- Raise an unexpected exception from code not running in plugin environment = Netbeans handles it as an exception and displays a pop up with the error.
- Raise an expected exception from code not running in plugin environment = ConstellationDialog handles it as an exception and displays a pop up with the error.

**Run Constellation in release**
- Raise an unexpected exception from code not running in plugin environment = Netbeans does not display anything to the user. **Constellation code cannot capture this case.**
- Raise an expected exception from code not running in plugin environment = ConstellationDialog handles it as an exception and displays a pop up with the error.

arcturus2:
@aldebaran30701 I found the answers in https://netbeans.apache.org/wiki/DevFaqUnexpectedExceptionDialog.asciidoc.

**show every error**
```
./constellation -J-Dnetbeans.exception.report.min.level=1 -J-Dnetbeans.exception.alert.min.level=1
```

**hide all errors** (more for FYI but not what we want)
```
./constellation -J-Dnetbeans.exception.report.min.level=99999 -J-Dnetbeans.exception.alert.min.level=99999
```

aldebaran30701:
Awesome find @arcturus2 
Do you feel that anything further is necessary for this ticket other than adding those parameters?

[...]

PR #1134, opened 2021-07-07
Show exception to user in executable version

aldebaran30701:
### Description of the Change
Add final modifiers to exceptions
Add below line of code where exceptions are commonly thrown as a result of something broken.
`NotifyDescriptor.Exception exceptionDialog = new NotifyDescriptor.Exception(ex);`
this line of code allows the error to be shown to the user in the release version meaning that they will know something is failing rather than failing silently. 

Additions to changelog and whatsnew also added. 

Exception when running from NetBeans:
image

Exception when running from executable:
build exception

Notification bubble is still available to be clicked and therefore possibly reporting the stacktrace to GitHub:
build exception click bubble

The link within the bubble, when clicked will load the very top window of the first screenshot - this allows the user to see and possibly copy the exception and report it on GitHub.

### Alternate Designs
Notification bubble was tested, but a dialog was seemingly the best option as it stops the user continuing without closing the dialog.

Allows users to see bugs and report them to us on GitHub

### Benefits
Allows users to see bugs and report them to us on GitHub

### Possible Drawbacks
None realised

### Verification Process
Built the executable version in order to test.
Tested using netbeans debug launch.

Both situations I replicated a known bug that threw an exception (or in the case of testing, a bug I planted to replicate the exception).

I ensured that during the executable build, that the exception was shown.
I ensured that during the NetBeans launch, that the exceptions were not extraneous - this doesn't seem to be an issue because one dialog is mentioning the issue and the other one is giving more details. 

### Applicable Issues
#50


aldebaran30701:
Wondering people's thoughts on the sonar issues such as removing the catch block when nothing is done within it. 
Also the unused variable which launches the dialog. If I don't assign it to a variable then NB complains also

antares1470:
> 
> Wondering people's thoughts on the sonar issues such as removing the catch block when nothing is done within it.
> Also the unused variable which launches the dialog. If I don't assign it to a variable then NB complains also

@aldebaran30701 Regarding the catch blocks, my preference is generally to avoid having nothing in a catch block. Looking at the specific instances of this, I think logging the exception should be enough as that alone doesn't halt execution (you need to throw the exception to do that) so the code would otherwise run as it did.

I've encountered the unused variable situation before so understand the dilemma. My thoughts have probably changed over time but currently I would say that the better option of the two is not declare a variable if you don't intend to use it later. We can then mark the subsequent sonar issue as won't do

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Constellation Changes
 
+## Changes in July 2021
+
+-   Exceptions thrown in the `DefaultPluginEnvironment` are now presented to
```
antares1470, line 5: Are you able to re-words this entry to start with the word "Updated"? Just to align it with other changelog entries

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginEnvironment.java
```diff
@@ -97,7 +100,8 @@
 
             try {
                 ConstellationLogger.getDefault().pluginStarted(plugin, parameters, graph);
-            } catch (Exception ex) {
+            } catch (final Exception ex) {
+                LOGGER.log(Level.WARNING, EXCEPTION_MESSAGE, ex);
```
antares1470, line 104: Should've mentioned it in my earlier comment but passing the exception as an argument will actually halt the execution (from previous experience) so just omit that so that execution runs as it did. I would also change the warning level to Severe since technically getting here means something went wrong:
`LOGGER.log(Level.SEVERE, EXCEPTION_MESSAGE)`

Similarly for the other log entries you added

antares1470 (changes requested):

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginEnvironment.java
aldebaran30701: So just log it with a default message like "An unexpected exception was caught."
 ?

antares1470: For now I would maybe log `ex.getLocalisedMessage()`

aldebaran30701: Done, let me know if that's how you meant

antares1470: Yep that's what I meant. The only thing I would add is change the log level to Severe rather than Warning. While we're not throwing the exception, getting to those catches does technically mean something went wrong so best to log it as much

Auriga2:
I tested this and looks good to me, but the only issue is `org.openide NotifyDescriptor.Exception ` is depricated. Can we use any  alternative suggested below?
http://bits.netbeans.org/dev/javadoc/org-openide-dialogs/org/openide/NotifyDescriptor.html
image

aldebaran30701:
Good point @Auriga2 
It looks like this post highlights some alternatives.
https://dev.platform.netbeans.narkive.com/VRm1x81G/why-notifydescriptor-exception-is-deprecated#post4

I was working on an alternate design using ErrorManager before trying the current PR. I can switch it over to this if all are in agreeance?

antares1470:
> 
> Good point @Auriga2
> It looks like this post highlights some alternatives.
> https://dev.platform.netbeans.narkive.com/VRm1x81G/why-notifydescriptor-exception-is-deprecated#post4
> 
> I was working on an alternate design using ErrorManager before trying the current PR. I can switch it over to this if all are in agreeance?

If `NotifyDescriptor.Exception` is deprecated then I'm keen to explore other options so happy for you to go with whatever option you think is best @aldebaran30701

arcturus2:
a few things @aldebaran30701:
* agreed also to not use the deprecated method. I had a look at current Apache NetBeans master branch and they still have it deprecated (so in theory we could use it as its not removed) but reading about some of this it looks like there was some deadlock issues
* in case you didn't know a quick way to build the stand alone version is to run the `build-zip` target, then unzip the `constellation.zip` file in the `dist` folder and run it including the path to your jre/jdk like `constellation/bin/constellation --jdkhome <path here>`
* i used the Experimental -> Developer -> Test Parameter Building for testing and introduced an error by changing the value in "Some Integer" from 5 to 5.5 and clicked OK. When run the exe version I would get the error icon appear but no popup. When I tested it through NetBeans I would get a popup every second time. This is so strange but may be clues to a deeper issue. Keen to know if you or @antares1470 or anyone else testing this PR experienced this?

aldebaran30701:
>a few things @aldebaran30701:
agreed also to not use the deprecated method. I had a look at current Apache NetBeans master branch and they still have it deprecated (so in theory we could use it as its not removed) but reading about some of this it looks like there was some deadlock issues

Agreed, plus there appears to be a direct replacement for it that should work fine.

>there is the ConstellationDialog class that you could extend with a ConstellationErrorDialogwith error specific things and use that? also keen to know what you were thinking of using.

I was thinking of just leaving it as is, but I can understand that if we extend the class we only have to change it in one spot in the future. I can add that as a fix to this.

>in case you didn't know a quick way to build the stand alone version is to run the build-zip target, then unzip the constellation.zip file in the dist folder and run it including the path to your jre/jdk like constellation/bin/constellation --jdkhome <path here>
I didn't know that, but will surely be handy for testing futher!


>i used the Experimental -> Developer -> Test Parameter Building for testing and introduced an error by changing the value in "Some Integer" from 5 to 5.5 and clicked OK. When run the exe version I would get the error icon appear but no popup. When I tested it through NetBeans I would get a popup every second time. This is so strange but may be clues to a deeper issue. Keen to know if you or @antares1470 or anyone else testing this PR experienced this?

The problem here is my assumption of things. Because everything in Constellation is a plugin, I added the error handling code to the only place I think I could have caught the application - wide exceptions; the plugin framework. 
It will only report the exception correctly when one occurs within a plugin.

arcturus2:
> It will only report the exception correctly when one occurs within a plugin.
Noted and your approach is fine. The "Test Parameter Building" is a plugin so I would expect it to be caught.

aldebaran30701:
>Noted and your approach is fine. The "Test Parameter Building" is a plugin so I would expect it to be caught.

Just checked the stack trace and I'm wondering if it isn't catching it because it's thrown and handled within JavaFX code. I'm getting only JavaFX libraries within the call stack (NumberFormatException).

aldebaran30701:
Note with this most recent commit, that a heap more has changed. 
I noted that there was some inconsistent behaviour between different plugin types and even the same exceptions across different classes. 

The convention I've went with in my mind is as follows:
- User implements code that may throw unexpected exceptions. These will get caught just inside of the Constellation code. (within methods like run or read of SimpleEditPlugin.java)
- The code will either report to the SimpleEditPlugin.java that it has had an exception occur and therefore cancel, and rethrow the exception for the more generic pluginframework code to handle.
- DefaultPluginEnvironment.java will then catch those exceptions and report them using the PluginInteraction class to display to the user (and also log to the console.)

The thing I was not sure about is the interrupted plugins which occur at the lower levels like in SimpleEditPlugin.java. My implementation takes note that it was interrupted, cancels the plugin via the already used boolean, and reports it up the chain to DefaultPluginEnvironment.java. That class then calls the thread.currentthread.interrupt. 

It seems that during the execution of the application on the build versions, that the LOGGER and also the std.err stream do not get picked up as displaying a dialog box (where it does so in the netbeans launch). I feel this is the case because of using the LOGGER.log(level,message,exception) to print out the stacktrace. The RuntimeExceptionAction simply throws an exception from the toolbar. No caller within Constellation-derived code. This makes it hard to capture in the same way the plugin framework currently uses. I have not found a solution for this.

I've added a class in the testing module which simply adds a toolbar option which throws an exception from within a plugin.

tl;dr - I made the exceptions more consistent within the pluginframework, but still have not got a solution for exceptions thrown outside of it.

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginInteraction.java
```diff
@@ -222,6 +223,30 @@ public void notify(final PluginNotificationLevel level, final String message) {
         }
     }
 
+    @Override
+    public void notifyException(PluginNotificationLevel level, Exception exception) {
```
antares1470, line 227: final

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/text/TextPluginInteraction.java
```diff
@@ -72,4 +72,9 @@ public boolean prompt(String promptName, PluginParameters parameters) {
         throw new UnsupportedOperationException(NOT_SUPPORTED); //To change body of generated methods, choose Tools | Templates.
     }
 
+    @Override
+    public void notifyException(PluginNotificationLevel level, Exception exception) {
```
antares1470, line 76: final

CoreTesting/src/au/gov/asd/tac/constellation/testing/RuntimeExceptionPlugin.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author aldebaran30701
+ */
+@ServiceProvider(service = Plugin.class)
+@PluginInfo(pluginType = PluginType.NONE, tags = {"LOW LEVEL"})
+@Messages("RuntimeExceptionPlugin=Throw a RuntimeException")
+public class RuntimeExceptionPlugin extends SimpleReadPlugin {
+
+    @Override
+    protected void read(GraphReadMethods graph, PluginInteraction interaction, PluginParameters parameters) throws InterruptedException, PluginException, RuntimeException {
```
antares1470, line 40: final

CoreTesting/src/au/gov/asd/tac/constellation/testing/RuntimeExceptionPluginAction.java
```diff
@@ -0,0 +1,54 @@
...
+@ActionRegistration(displayName = "#CTL_CreateRuntimeExceptionAction", surviveFocusChange = true)
+@ActionReferences({
+    @ActionReference(path = "Menu/Experimental/Developer", position = 0)
+})
+@Messages("CTL_CreateRuntimeExceptionAction=Throw a RuntimeException from a plugin")
+public final class RuntimeExceptionPluginAction implements ActionListener {
+
+    private final GraphNode context;
+
+    public RuntimeExceptionPluginAction(GraphNode context) {
```
antares1470, line 46: final

antares1470 (changes requested):

arcturus2:
@aldebaran30701 so gave this a test. Just to summarise I am seeing a popup from "Experimental -> Developer -> Throw a RuntimeException from a plugin" but for everything else it is the little icon and no popup shows. Is that consistent to what you are seeing or not? I am testing on Linux.

aldebaran30701:
Converting to draft until I get a few moments to review the comments and make some changes.

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/templates/SimpleEditPlugin.java
```diff
@@ -116,34 +107,17 @@ public final void run(final PluginGraphs graphs, final PluginInteraction interac
...
-                        LOGGER.log(Level.WARNING, msg, e);
-                        throw e;
-                    } catch (Exception ex) {
-                        final String msg0 = String.format("Unexpected non-plugin exception caught in %s.run()", SimpleEditPlugin.class.getName());
-                        final String msg = Bundle.MSG_Edit_Failed(graph, getName());
-                        interaction.notify(PluginNotificationLevel.ERROR, msg0 + SeparatorConstants.SEMICOLON + SeparatorConstants.NEWLINE + msg + SeparatorConstants.NEWLINE + ex.getMessage());
+                    } catch (final Exception ex) {
                         cancelled = true;
-                        LOGGER.log(Level.WARNING, ex, () -> msg0 + SeparatorConstants.SEMICOLON + " " + msg);
-                        throw new RuntimeException(ex);
```
arcturus2, line 135: Deciding not to wrap the extension in a RuntimeException() - https://softwareengineering.stackexchange.com/questions/121328/is-it-good-practice-to-catch-a-checked-exception-and-throw-a-runtimeexception

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginEnvironment.java
arcturus2: Updated

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/plugins/DefaultPluginInteraction.java
arcturus2: Refactored and updated.

CoreTesting/src/au/gov/asd/tac/constellation/testing/RuntimeExceptionPlugin.java
arcturus2: Updated

CoreTesting/src/au/gov/asd/tac/constellation/testing/RuntimeExceptionPluginAction.java
arcturus2: Updated

arcturus2:
@aldebaran30701 I closed your PR and started a new one consolidating everything to kick off a fresh review https://github.com/constellation-app/constellation/pull/1459. I hope that is OK with you. I wanted to add some unit tests and then I will open the PR for review. Thanks for starting this work and also for allowing me to continue it.

I really like what its turned into. Check it out when you can.

Issue #1103, opened 2021-06-08
Update Issue Template


sol695510:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Issues lack enough concise information in the case of expected performance. For example tickets just state in the issue description:

**Description:** Plugin "X" is too slow and should be optimised to execute faster.

When they should use more measurable metrics that we can work towards in the expected/actual behaviour such as:

**Expected Behaviour:** Plugin "X" takes 5 seconds to complete execution.

**Actual Behaviour:** Plugin "X" takes 10 seconds to complete execution.

The "Steps to Reproduce" should outline the steps required to reproduce the 10 second example metric above. The rest should filled out accordingly.

The new issue ticket template needs to be updated to include instructions to do the above in the requirements and/or the sub-heading descriptions.

danger-ahead:
I can do this. Please assign.

sol695510:
> I can do this. Please assign.

Make the changes on the file ISSUE_TEMPLATE.md and create a PR.

danger-ahead:
@sol695510 possibly you were referring to this change. Please have a look.
Do let me know if there are other changes necessary.

PR #1139, opened 2021-07-09
Update ISSUE_TEMPLATE.md

sol695510:
### Description of the Change

Update the descriptions of fields to be filled in the Issue Template.

### Alternate Designs

N/A

N/A

### Benefits

Provides clearer goals for fixing the issue.

### Possible Drawbacks

N/A

### Verification Process

N/A

### Applicable Issues

#1103

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -45,17 +45,13 @@ https://github.com/constellation-app/constellation/blob/master/CODE_OF_CONDUCT.m
 
 ### Behaviour
 
-Strive to include measurable metrics that can be aimed for (if applicable), see examples below.
+Try to include measurable metrics that can be aimed for (if applicable to the issue),
+**e.g.** If its a time issue then "should take 5 seconds or less to finish"
+or if its a workflow issue "should be accomplished in 2 or less clicks".
 
 **Expected behaviour:** \[What you expect to happen\]
 
--   e.g. should take 5 seconds or less to finish **OR** should be accomplished in 2 or less clicks,
-    **NOT** should finish faster **OR** should take less clicks to be accomplished.
-
 **Actual behaviour:** \[What actually happens\]
-
--   e.g. takes 10 or more seconds to finish **OR** takes 2 or more clicks to accomplish the task,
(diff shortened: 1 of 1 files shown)


aldebaran30701 (changes requested):
I'd maybe suggest writing it only once - like a section just above Expected Behaviour. This section could say
```
Strive to include measurable metrics that can be aimed for when describing behaviour.

e.g. action 'x' should take 5 seconds or less to finish, NOT action 'x' should finish faster.

e.g. action 'x' takes 4 clicks, NOT action 'x' multiple clicks to be accomplished.
```

Let me know your thoughts

sol695510:
> I'd maybe suggest writing it only once - like a section just above Expected Behaviour. This section could say
> 
> ```
> Strive to include measurable metrics that can be aimed for when describing behaviour.
> 
> e.g. action 'x' should take 5 seconds or less to finish, NOT action 'x' should finish faster.
> 
> e.g. action 'x' takes 4 clicks, NOT action 'x' multiple clicks to be accomplished.
> ```
> 
> Let me know your thoughts

The "strive" part can be just written once, but the "e.g." parts are not all the same.

aldebaran30701:
>The "strive" part can be just written once, but the "e.g." parts are not all the same.

Understood, that's why I worked in an example of each.

I'm thinking with the new changes made, it should only include one example for each heading. We don't want to make it a very cluttered PR template, and as such only one line should be suffice to get the point across.

ISSUE_TEMPLATE.md
```diff
@@ -43,12 +43,17 @@ https://github.com/constellation-app/constellation/blob/master/CODE_OF_CONDUCT.m
 
 3.  \[and so on…\]
 
+### Behaviour
+
+Try to include measurable metrics that can be aimed for (if applicable to the issue),
+**e.g.** If its a time issue then "should take 5 seconds or less to finish"
+or if its a workflow issue "should be accomplished in 2 or less clicks".
```
antares1470, line 50: I would comment this out (i.e. surround it in \).

Also "it's a time" and "it's a workflow"

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):

Issue #1136, opened 2021-07-08
Change Quality Category names and colours


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The current quality categories are Default, Info, Warning, Severe, and Fatal. Some of these make sense as categories but some really don't make sense as to that they mean (e.g. what does it mean for something to flag as 'Default'). The colours assigned to them are also a bit random (White, Blue, Purple, Red, Black respectively).

I'm proposing changing these categories to something that makes a little more sense to read and understand. Something like Ok (if no rules flag), Minor, Medium, Major, Severe, Critical. Colours could then follow something like Green, Blue, Yellow, Orange, Red, Dark Red/Black respectively.

**Expected behaviour:** The meaning of a Quality category is easy to understnad

**Actual behaviour:** Some quality categories (e.g. Default) are confusing to understand

**Reproduces how often:** 40-60% (The number of categories I think that are currently a tad confusing)

### Additional Information

Nil

aldebaran30701:
Just some background, the current categories are based off the netbeans logging system for consistency. The colours were meant to be linear but I agree having orange in the mix is a better representation of getting worse. We'd have to ensure that the colors fit with colorblind users (Yellow background, white text | Yellow category, Orange category), but also worth noting that the categories are also listed in text on the row of the table.

- Default - now OK - Green
- Info - now Minor - Blue
- Warning - now Medium - Yellow/Orange
- Severe - now Major - Light Red
- Fatal - now Critical - Dark Red

When you have many rules which are in the black coloured category, it is hard to tell there is any issue in dark mode - hence the change to dark red.
![image](https://user-images.githubusercontent.com/58677759/125368087-a401c880-e3b8-11eb-994f-de463bfc16f9.png)

[...]

PR #1144, opened 2021-07-23
Update Quality Control Category Names and Colours

Delphinus8821:
### Description of the Change
Updated the names of most of the Quality Control Categories and some of the colours to be more visible and less confusing. 
The category names have been changed from Default, Info, Warning, Severe and Fatal, which could be confusing, to Minor, Medium, Major, Severe and Critical. A category called OK has been added for when no Quality Control rules have been flagged. 

### Alternate Designs
None realised.

Updates the Quality Control categories to be less confusing for users by providing more meaningful category names. The colours are now also more meaningful by being colours that are more representative of the category names (e.g. OK is Green, Minor is Blue, Medium is Yellow, Major is Orange, Severe is Red and Critical is Dark Red).

### Benefits
Makes the Quality Control view less confusing for users. 

### Possible Drawbacks
None realised.

### Verification Process
Tested that all of the new category names and colours appear correctly, see images below.

### Applicable Issues
#1136 

OK
Minor
Medium
Major
Severe
Critical

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Changes in July 2021
 
+-   Update to Quality Control category names and colours in 
+    'QualityControlEvent' to be easier to understand. 
+
 -   Update to `openEdit`, `updateNotesUI`, `createNote` and the constructor in 
     `NotesViewPane` to include variables to hold what graph elements are 
     selected and applied to the note. Also included a right click context menu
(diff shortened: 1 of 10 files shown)


arcturus2:
I'm yet to test it but it looks absolutely amazing @Delphinus8821.

antares1470 (changes requested):
One minor thing but the Quality Category button in DAV has a white background when the category is ok. If you could change that to the same green you have in the QCV, that would be great (I did a quick test of this and I don't think there are any issues of this against the Go button).

Also if you're able to add a whatsnew and changelog entries, that would be good.

antares1470 (approved):

arcturus2 (approved):
Just brilliant :star:

Issue #1120, opened 2021-06-23
Change Default Label Text Colour


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The default label colour at the moment is blue. While it looks cool though, blue text on a black background can at times be difficult to read (or at the very least more difficult compared to other possible combinations). I'm proposing changing the default label colour to white as this is much more readable and there really isn't much reason to retain the blue as a default anyway.

### Steps to Reproduce

1.  Create a node and view label

**Expected behaviour:** Label is white by default

**Actual behaviour:** Label is blue by default

**Reproduces how often:** 100% (if you haven't changed from the defaults)

### Additional Information

This will help with accessibility and work being done as a part of #46

GammaVel:
This looks awesome ! I have looked at new graphs, confirmed white is the default for the top label text and checked existing graphs (nodes added to these are still blue). thanks @Delphinus8821 !

PR #1146, opened 2021-07-26
Change Default Label Colour to White

Delphinus8821:
### Description of the Change
Changed the default node and transaction label colour to white as it is easier to see on the black graph background.

### Alternate Designs
None realised

Makes the nodes and transactions labels easier to read by default.

### Possible Drawbacks
None realised

### Verification Process
Created both nodes and transactions to check the label colour had changed to white. 
Checked that the sphere graph builder also has updated to use white labels for nodes and transactions.

### Applicable Issues
#1120 

image


antares1470 (changes requested):
Just a whatsnew entry and this should be good to go

antares1470 (approved):

arcturus2 (approved):
Looks good.

image

Issue #1128, opened 2021-07-01
Keep Attribute Editor properties expanded


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Whenever the graph selection changes, any expanded node or transaction attributes contract to display "<Multiple Values>", rather than remain expanded and populate newly selected values. A toggle next to each attribute allowing users to keep the values expanded (or expand it for them when the selection changes) would save a lot of time doing it manually.

### Steps to Reproduce

1. Open a graph and the Attribute Editor
2. Select a set of nodes or transactions such that an attribute has more than one value for the selection
3. Expand an attribute with multiple values
4. Change the graph selection such that the chosen attribute still has multiple values

**Expected behaviour:** There is an option to keep the attribute values expanded

**Actual behaviour:** Attribute values always contract

**Reproduces how often:** 100%

Auriga2:
I added an option next to the Expand/collapse button, with a too tip text as below: 
1.  Any suggestion to improve it's appearance and location?
2. Should this option work across all graphs? Current version only works within a graph.
3. When zoom in/out currently it only works for the attributes this new option is checked. I guess this should be for everything. (Currently working on it- it's triggering attibuteValueHasChanged = true for at least one attribute)
![keep attributes expanded](https://user-images.githubusercontent.com/64826928/126583614-37dd3af7-6c07-47a1-9f14-4f3acc2e1f07.gif)

GammaVel:
Thanks @Auriga2 for making these improvements. I noticed that the checkbox didn't get implemented as initially mentioned in the issue description. As you found, if you change graphs the attributes that are expanded don't retain state when you go back to the original graph. 

I also noticed if you are looking at multiple values in the transaction section and then select multiple transactions and nodes, Constellation will auto scroll back up to the Nodes section, not the transaction section ( I believe this is because of the priority of the nodes.   Cheers :-)

Auriga2:
[...]

PR #1147, opened 2021-07-26
Keep Attribute Editor multiple values expanded

Auriga2:
### Description of the Change
When the graph selection was changing, or graph was zooming in or out , any expanded node or transaction attributes were contracting to display "Multiple Values", rather than remaining expanded and populating newly selected values. 

### Alternate Designs
A possible extension is to keep the selection of attributes across graphs, and/or retain the selection upon returning to the graph. Currently it works only within a graph and loses the selection if the user leaves and returns a graph.

Better UX and saves time

### Benefits
Better UX and saves time

### Possible Drawbacks
N/A

### Verification Process

1. Open a graph and the Attribute Editor
2. Select a set of nodes or transactions such that an attribute has more than one value for the selection
3. Expand an attribute with multiple values
4. Change the graph selection such that the chosen attribute still has multiple values and verify the expanded attribute/s do not collapse. 
5. Select a single Node or Transaction and verify the attribute/s collapse. Select multiple Nodes or Transactions and verify the previously viewed attribute/s expand again.
6. Zoom in and out the graph and verify the attributes are still expanded. 

 
### Applicable Issues

https://github.com/constellation-app/constellation/issues/1128


CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeData.java
```diff
@@ -128,4 +129,18 @@ public boolean isKey() {
...
+     * Should multiple values of this attribute be always expanded .
+     *
+     * @return True if the attribute values should always be expanded, false
+     * otherwise.
+     */
+    public boolean isKeepExpanded() {
+        return isKeepExpanded;
+    }
+
+    public void setKeepExpanded(boolean isKeepExpanded) {
```
antares1470, line 143: final

antares1470 (changes requested):
If you're able to add a whatsnew entry as well that would be great

antares1470 (approved):

CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorPanel.java
```diff
@@ -675,6 +676,10 @@ private void createMultiValuePane(final AttributeData attribute, final TitledPan
             }
         });
         attributePane.setContent(dataAndMoreButtonBox);
+
+        //attributePane TitledPane EXPAND and COLLAPSE events
+        attributePane.expandedProperty().addListener((observable, wasExpanded, isNowExpanded)
```
arcturus2, line 681: I'm not sure if there is a way to remove the listener - I couldn't find a place to do it in this class.

arcturus2 (approved):
Works great.

PR #1148, opened 2021-07-27
Layers view - Remove a layer

aldebaran30701:
### Description of the Change
This change adds a delete button to each layer. When pressed, it allows the collection within the controller to be updated with that element removed. The query collection will then remove that element, sort the array and also ensure that the index and bitmask values are updated with the new layer positions.

The Layers View Pane will also write those changes to state, and execute the queries again to ensure that by deleting a selected layer, that the current view is consistent with what is visible on the graph.

In doing so, I had to remove some final modifiers and make a new method which removed an item and shifted the array positions.

### Alternate Designs
N/A

Allows removal of a layer

### Benefits
Allows removal of a layer

### Possible Drawbacks
N/A

### Verification Process
Open up a graph with layers, open up a graph without layers. 
on each graph,
Add more layers, and remove layers on the start, middle and end of the list of layers. Ensure that the layers are consistent. 
Add those layers and select one or many layers as visible. Then proceed to remove the start, middle and end layers.
Switch between graphs to ensure it has been saved in state.
Toggle some layers as visible, then delete layers before and after it. check to ensure graph is consistent with the current selections.

### Applicable Issues
Should close - #970


antares1470 (changes requested):
If you're able to add a whatsnew entry, that would be great

arcturus2 (changes requested):
I found an issue where by when I create 3 layers and delete layer 2, it assigns the selection in layer 3 to 2 instead. See video for a better explanation.


https://user-images.githubusercontent.com/39325530/127243484-ef454f30-4696-4894-9498-d5830eba4c25.mp4

antares1470:
@aldebaran30701 To add to what @arcturus2 saw, I did a similar test except I used queries for the layers. I can confirm there were no issues in that situation so I think the issue is specific to creating layers via the context menu

aldebaran30701:
Good catch @arcturus2 ,
I missed that in the initial testing as mentioned in the ticket info.
I'm thinking I'll have to use a plugin to loop all elements and then check for that mask index and then remove it from the element. 
Fingers crossed that it goes as smoothly as that!

aldebaran30701:
I've made a new plugin which will shuffle around the bitmasks to ensure they are kept in sync when a manual layer is deleted.
I've written up a few tests to ensure it is working as intended also. There are a few other tests I would like to get to eventually but I also see the value in getting the feature in before the next release + testing.

antares1470 (changes requested):
Good news, deleting manually created layers works properly now.

Bad news, I got an exception trying to use the keyboard shortcuts for either adding a new layer or deselecting all layers. This the exception for add new layer (analogous for deselecting all layers):
```
    java.lang.IllegalArgumentException: No such plugin as au.gov.asd.tac.constellation.views.layers.shortcut.NewLayerPlugin!
    at au.gov.asd.tac.constellation.plugins.PluginRegistry.get(PluginRegistry.java:185)
    at au.gov.asd.tac.constellation.plugins.PluginExecution.withPlugin(PluginExecution.java:127)
    at au.gov.asd.tac.constellation.views.layers.shortcut.LayersViewShortcuts.actionPerformed(LayersViewShortcuts.java:99)

antares1470 (approved):

aldebaran30701:
Just added a whats new entry

aldebaran30701:
Fixed the CI failing issue, it was the initialization of `LayersViewTopComponent`. I've removed that test as it does not add much value. I've also taken out the `LayersPluginRegistry` as I did not realise that it is used for plugins with no arg constructors. It did not add value. Did a few of the major sonar fixes also. Functionality is still the same, just some cleanup.

arcturus2:
I got this when I opened the layers view and added a node to a graph

image

aldebaran30701:
Thats hopefully the last of the residual service provider implementation I was testing. Sorry about that! Should be good now

arcturus2 (approved):

Issue #1005, opened 2021-03-29
More Granular Filters in Notes View


antares1470:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently you can only filter on all of the auto notes and all of the user notes. It would be good to have more granular filters e.g. only show auto notes for plugins with a specific tag.

**Expected behaviour:** Have more granular filters to be able to filter notes on

**Actual behaviour:** Only able to filter on all auto notes and all user notes

**Reproduces how often:** 100%

### Additional Information

If done with or after #1004, this could also include filtering on notes relating to specific nodes or transactions.

GammaVel:
Suggestion to add in a time based filter for the Notes

antares1470:
Being able to filter by tags may allow more appropriate default filters to be set for the auto notes as well

GammaVel:
Thanks for implementing the new filters @Delphinus8821 . They are working a treat!

PR #1152, opened 2021-08-02
Added More Filters for Auto Notes

Delphinus8821:
### Description of the Change
Have added another check combo box to the top of the Notes View to filter on what Auto Notes are displayed. This check combo box is only visible when the Auto Notes filter is selected. The second check combo box is populated with the same tags that appear in the Plugin Reporter. 

### Alternate Designs
None realised

This change allows for users to filter more thoroughly on the Auto notes that are generated. 

### Benefits
Gives more users more options for filtering on what notes appear on the UI as not all auto notes that are generated would be relevant for users to view.

### Possible Drawbacks
None Realised

### Verification Process
Old graphs do not have plugin reporter tags attached to auto notes which results in the new check combo box being empty until new plugins are run. If nothing is selected in the new check combo box then all Auto notes are displayed and this occurs by default. 

### Applicable Issues
#1005 

New Check Combo Box


arcturus2 (approved):
@Delphinus8821 this works really well and feels quick.

I noticed that when I loaded a saved graph with a filter (i.e GENERAL) that it did not load with just the "GENERAL" notes but with all of them.

I also noticed that switching graphs did not load the filters from the view's state but rather the last thing that was there on the notes view.

I think both of these are intentional so not an issue. The plugin reporter has the same behavior.

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -160,6 +168,40 @@ public void onChanged(final ListChangeListener.Change event) {
...
+                }
+            }
+        });
+
+        tagsFiltersList = FXCollections.observableArrayList(tagsUpdater);
+
+        // CheckComboBox for the Auto Note filters
+        autoFilterCheckComboBox = new CheckComboBox(tagsFiltersList);
+        autoFilterCheckComboBox.setStyle(String.format("-fx-font-size:%d;", FontUtilities.getApplicationFontSize()));
+        autoFilterCheckComboBox.getCheckModel().getCheckedItems().addListener(new ListChangeListener() {
```
antares1470, line 188: convert to a lambda (looks like you'll need to specify the type of the parameter though due to ambiguity)

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -339,6 +381,15 @@ protected void addPluginReport(final PluginReport pluginReport) {
                     false
             );
 
+            final String[] tags = pluginReport.getTags();
+            final List<String> tagsList = new ArrayList<>();
+            for (String tag : tags) {
```
antares1470, line 386: final

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -429,9 +486,15 @@ protected synchronized void updateNotesUI() {
...
                 notesViewEntries.forEach(entry -> {
                     // Add note to render list if its respective filter is selected.
-                    if ((selectedFilters.contains(USER_NOTES_FILTER) && entry.isUserCreated())
-                            || (selectedFilters.contains(AUTO_NOTES_FILTER) && !entry.isUserCreated())) {
+                    if ((selectedFilters.contains(USER_NOTES_FILTER) && entry.isUserCreated())) {
                         notesToRender.add(entry);
+
+                    } else if (selectedFilters.contains(AUTO_NOTES_FILTER) && !entry.isUserCreated()) {
+                        final boolean addNote = updateAutoNotesDisplayed(entry);
+                        if (addNote) {
```
antares1470, line 494: I'd probably just merge these together given this is the only spot the boolean expression is used

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/state/NotesViewStateIoProvider.java
```diff
@@ -90,6 +91,17 @@ public void readObject(final int attributeId, final int elementId, final JsonNod
                                 }
                                 noteViewEntries.get(i).setTransactionsSelected(selectedTransactions);
                             }
+                        } else if (notesArray.get(i).get(3).asBoolean() == false) {
+                            // Create auto notes with the tags they have assigned to them
+                            final JsonNode tagsArrayNode = notesArray.get(i).get(5);
+                            if (tagsArrayNode != null) {
+                                List<String> tagsArray = new ArrayList<>();
```
antares1470, line 98: final

antares1470 (changes requested):
The only thing I got from testing was that old graphs with existing auto notes don't have any tags (only newly generated auto notes will have tags) meaning you won't see any old auto notes when applying filters (and filters won't be available until after new auto notes are generated). Not sure though whether that is something to fix here or as a separate task

If you're also able to add a whatsnew as well as a changelog entry for some of the function definition changes, that would be great

antares1470 (approved):

PR #1157, opened 2021-08-03
DraggableCell Unit Test

antares1470:
### Description of the Change

Have added some unit tests for `DraggableCell`. Worth noting that a `DraggableCell` object requires the JFX Toolkit to be initialised so consequently all the unit tests sit in `if (!GraphicEnvironment.isHeadless())` blocks and so won't run in CI (nor be picked up by Sonar).

I also omitted testing of the `onDragDetected` listener as the call to `startDragAndDrop` ended up making it too difficult to be able to run in a unit test (there also weren't many asserts I could've added for it anyway).

Unit testing Core

### Benefits

More unit tests

### Possible Drawbacks

Nil


CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/io/DraggableCellNGTest.java
```diff
@@ -0,0 +1,241 @@
...
+        // (so CI won't actually tell you whether this is passing or not)
+        if (!GraphicsEnvironment.isHeadless()) {
+            new JFXPanel();
+            
+            final DraggableCell<String> instance = new DraggableCell<>();
+            instance.updateItem(null, false);
+
+            assertEquals(instance.getText(), null);
+            assertEquals(instance.getItem(), null);
+            assertEquals(instance.isEmpty(), false);
```
aldebaran30701, line 81: Why is it that when you pass it null, its not empty, and when you pass it an object it returns true to the call is.Empty() ?

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/io/DraggableCellNGTest.java
```diff
@@ -0,0 +1,241 @@
...
+            
+            // mocking the required dragboard
+            final Dragboard dragboard = mock(Dragboard.class);
+            when(dragboard.hasString()).thenReturn(true);
+            
+            final DragEvent event  = new DragEvent(null, dragboard, 0, 0, 0, 0, null, null, null, null);
+            
+            instance.getOnDragExited().handle(event);
+            
+            assertEquals(instance.getOpacity(), 1.0);
```
aldebaran30701, line 195: Would it be beneficial to have a call to one of the other tests to set the opacity to something other than 1.0, so that you can verify that it has actually changed between events and not just sitting at its default value?

ie. 
```
setOpacity(0.3);
assertEquals(getOpacity(), 0.3f);
doOnDrag();
assertEquals(getOpacity(), 1.0f);
```

aldebaran30701 (changes requested):
Looks good, just had some queries mentioned in the review. 👍

arcturus2 (approved):
Thank you @antares1470 for doing this. I've approved this but would like to get an answer for @aldebaran30701's questions and also @aldebaran30701 can you see a way for this class to be tested avoiding the use of headless and trying things like `verify` like you did with your PR https://github.com/constellation-app/constellation/pull/1165 ?

aldebaran30701:
@arcturus2 I think the catching of the exception in my PR is a special case, because the actual code which needed to be executed was on the FX Platform thread. I couldn't test it unless I initialized the toolkit. In my PR I haven't actually tested any code within the JFX Platform thread. I just call a verify to make sure that the line right before the platform thread is called.
I can't see a way other than mocking every single JFX element in this PR, to do it without the use of the if headless() statement.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/io/DraggableCellNGTest.java
antares1470: @aldebaran30701 DraggableCell (or rather its super class Cell) has a property called `empty` so `isEmpty()` is actually checking whether that property value is false, not whether the cell contains any content (admittedly confusing given the context of `isEmpty()` used for just about any other Java object). So the 2nd parameter in `updateItem()` is setting the empty property so that's the value that is determining my assert

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/io/DraggableCellNGTest.java
antares1470: Good pickup. I'll add that

aldebaran30701 (approved):

Issue #413, opened 2020-04-07
Unit testing improvements


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

This is a list of unit testing improvements I'd like to see considered. These should be separate tickets but there are too many already:

 * Use mockito or a good mocking library
 * Write unit tests to open test graphs and make sure they work
 * We could make use of the NetBeans harness module
 * Automated functional tests somehow
 * Modify existing unit tests expecting an exception to occur to annotate the expected Exception rather than use the convention to say the unit test "throws Exception"
 * All public methods should have a unit test definition and if its not implemented yet then set it to "skip" the test. This will help track how many unit tests are outstanding. 
 * Ensure that schemas are thoroughly tested

On top of this, there is a large number of classes and functions currently lacking any unit testing so this should also be fixed.

arcturus2:
If you go to https://sonarcloud.io/component_measures?id=constellation-app_constellation&metric=Coverage&view=list you can also get a feel for which classes have no unit testing coverage so perhaps focusing on classes from the top right and work to the top left would be good.

![image](https://user-images.githubusercontent.com/39325530/91870641-5e657100-ecba-11ea-8b81-b39b3a863d4c.png)

This is a link to the list view of this https://sonarcloud.io/component_measures?id=constellation-app_constellation&metric=uncovered_lines&view=list so we could co-ordinate a surge on this when people are ready.

GammaVel:
Moved ticket back into To do for prioritisation.

PR #1165, opened 2021-08-05
✅ Test DataAccessSearchProvider

aldebaran30701:
Test using Mockito as tested methods were overrides and returned void. 
Mockito was used to test if key methods were called and their parameters. 

Applicable issues:
#413


CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,324 @@
...
+            when(daPane.getCurrentTab()).thenReturn(new Tab());
+
+            // Mock the static method getDataAccessPane and return the mocked DataAccessPane
+            try (MockedStatic<DataAccessUtilities> mockedStatic2 = Mockito.mockStatic(DataAccessUtilities.class)) {
+                mockedStatic2.when(() -> DataAccessUtilities.getDataAccessPane()).thenReturn(daPane);
+
+                try {
+                    PluginDisplayer pd = new PluginDisplayer("SelectTopN");
+                    pd.run();
+                } catch (final IllegalStateException ex) {
```
aldebaran30701, line 236: @antares1470 I've just caught the exception for the toolkit not instantiated and allowed the code to continue in this situation as I'm only able to test the verify() calls as opposed to a return value.
I was thinking about surrounding it in the if headless block but that would not test anything in the method that way.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * Test of evaluate method, of class DataAccessSearchProvider. Pass in an
+     * unmatchable search string
+     */
+    @Test
+    public void testEvaluateFail() {
+        System.out.println("evaluate fail");
+
+        // Creating mocks
+        MockitoAnnotations.initMocks(this);
```
arcturus2, line 95: `initMocks` is deprecated. Use `openMocks` instead.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+        // Mock the request text to be an unmatchable string
+        when(request.getText()).thenReturn("nothignshouldmatchthisstring");
+        when(response.addResult(Mockito.anyObject(), Mockito.anyString())).thenReturn(true);
+        DataAccessSearchProvider instance = new DataAccessSearchProvider();
+        instance.evaluate(request, response);
+
+        // Verify that addResult was never called.
+        // This should mean that no plugin name matched the input
+        verify(response, never()).addResult(Mockito.anyObject(), Mockito.anyString());
```
arcturus2, line 121: `anyObject()` is going to be deprecated in Mockito 4. Can we use `new Object()` or something like this to avoid this becoming a problem in the future upgrades given at the time of deving this we know it is marked for deprecation. See https://www.javadoc.io/static/org.mockito/mockito-core/3.11.2/org/mockito/ArgumentMatchers.html#anyObject-- for more information.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,324 @@
...
+        };
+
+        // Setting up a mock for DataAccessUtilities to return null when fetching the pane.
+        try (MockedStatic<DataAccessUtilities> mockedStatic1 = Mockito.mockStatic(DataAccessUtilities.class)) {
+            when(daPane.getCurrentTab()).thenReturn(null);
+            mockedStatic1.when(() -> DataAccessUtilities.getDataAccessPane()).thenReturn(daPane);
+
+            // Mock the static method getDefault() to return the mock of NotificationDisplayer
+            try (MockedStatic<NotificationDisplayer> mockedStatic3 = Mockito.mockStatic(NotificationDisplayer.class)) {
+                when(notifDisplayer.notify("Data Access view", UserInterfaceIconProvider.WARNING.buildIcon(16, ConstellationColor.DARK_ORANGE.getJavaColor()), PluginDisplayer.STEP_STRING, null)).thenReturn(notif);
```
arcturus2, line 275: To make things easier to read can you extract out `UserInterfaceIconProvider.WARNING.buildIcon(16, ConstellationColor.DARK_ORANGE.getJavaColor()` as its used twice here?

Ditto `testRun3`

arcturus2 (changes requested):
This is really impressive as you have managed to test a lot of the class using a range of Mockito concepts like `verify` and others I have not used like `Mockito.mockStatic`.

There are some deprecated methods being used which I think should be fixed given we know they are deprecated at the time of adding these unit tests in. OK with these suggestions?

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
arcturus2: Perhaps this is a better solution than checking for headless. :+1:

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
aldebaran30701: any() seems to be the next in line to use, I originally had it as any() but I was having issues elsewhere and was changing things here to check if they would help me.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
aldebaran30701: Easy!

antares1470 (commented):
Maybe this is just me misunderstanding all the mockito functions but it doesn't seem like that any asserts are being made in the unit tests about the behaviour of the function being tested. From what I can tell, `verify()` is checking whether certain lines are being called (correct me if I'm wrong there). 

While checking which lines are called how can sometimes correlate to certain behaviour being achieved, I don't think this is always the case. If we hypothetically re-wrote one of these functions (without changing its intended behaviour), I would expect a unit test to still pass (usually) but I'm not sure that would be the case in this instance.

Maybe you are able to clarify @aldebaran30701?

aldebaran30701:
@antares1470 You are correct in the assumption, I could not find another way to test this class, other than how I've done it here. 
I'm all ears to change it to something that tests it better, I couldn't find a suitable refactor to make it easier to test.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,327 @@
...
+    @Test
+    public void testEvaluateFail() {
+        System.out.println("evaluate fail");
+
+        // Creating mocks
+        MockitoAnnotations.openMocks(this);
+        daPane = mock(DataAccessPane.class);
+        qpPane = mock(QueryPhasePane.class);
+        request = mock(SearchRequest.class);
+        response = mock(SearchResponse.class);
```
antares1470, line 102: The `@Mock` annotation with `openMocks()` call is equivalent to calling the `mock()` on all the mock object so you'll only need to do one or the other (similar in other tests)

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessSearchProviderNGTest.java
```diff
@@ -0,0 +1,327 @@
...
+    protected NotificationDisplayer notifDisplayer;
+
+    private static final Icon WARNING_ICON = UserInterfaceIconProvider.WARNING.buildIcon(16, ConstellationColor.DARK_ORANGE.getJavaColor());
+
+    public DataAccessSearchProviderNGTest() {
+    }
+
+    @BeforeClass
+    public void setUpClass() throws Exception {
+        MockitoAnnotations.openMocks(this);
```
antares1470, line 72: You've got this called in all your tests. You could then move this to `setUpMethod()` and remove it from all the tests (if you decide to go down the mock annotations path)

antares1470 (changes requested):

antares1470 (approved):

PR #1170, opened 2021-08-06
Unit Tests for ExtractTypes and ExtractWords Plugins

Delphinus8821:
### Description of the Change
Added unit tests for ExtractTypesFromTextPlugin.
Also added additional tests for ExtractWordsFromTextPlugin

Unit testing Core

### Benefits
More unit tests

### DIFF
--- a/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractTypesFromTextPluginNGTest.java
+++ b/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractTypesFromTextPluginNGTest.java
@@ -17,9 +17,13 @@ package au.gov.asd.tac.constellation.views.dataaccess.plugins.importing;
 
 import au.gov.asd.tac.constellation.graph.StoreGraph;
 import au.gov.asd.tac.constellation.graph.processing.GraphRecordStore;
+import au.gov.asd.tac.constellation.graph.processing.GraphRecordStoreUtilities;
 import au.gov.asd.tac.constellation.graph.processing.RecordStore;
 import au.gov.asd.tac.constellation.graph.schema.SchemaFactoryUtilities;
 import au.gov.asd.tac.constellation.graph.schema.analytic.AnalyticSchemaFactory;
+import au.gov.asd.tac.constellation.graph.schema.analytic.concept.AnalyticConcept;
+import au.gov.asd.tac.constellation.graph.schema.visual.concept.VisualConcept;
+import au.gov.asd.tac.constellation.plugins.PluginException;
 import au.gov.asd.tac.constellation.plugins.PluginInteraction;
 import au.gov.asd.tac.constellation.plugins.parameters.PluginParameters;
 import au.gov.asd.tac.constellation.plugins.text.TextPluginInteraction;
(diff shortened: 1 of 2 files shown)


CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractTypesFromTextPluginNGTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @AfterMethod
+    public void tearDownMethod() throws Exception {
+    }
+
+    /**
+     * Test of getType method, of class ExtractTypesFromTextPlugin.
+     */
+    @Test
+    public void testGetType() {
```
antares1470, line 67: can keep this in but in my personal opinion, I don't we get much value from testing the likes of `getType`, `getPosition`, and `getDescription`

Delphinus8821: They were auto generated when I made the test files. They probably don't give much value but I don't think it hurts to include them. I can remove them if it is preferred.

antares1470 (changes requested):
I think there are a few more test cases for the query function in ExtractTypesFromText (e.g. null string, string where a detection regex gets picked up) so it would be good to include test cases for those.

Your test for updateParameters in ExtractWordsFromText also appears to be lacking any asserts so it would be nice to include some if possible.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractWordsFromTextPluginNGTest.java
```diff
@@ -549,4 +585,35 @@ public void testExtractWordFromContentRegexOnlyLowerCaseFalseParameter() throws
...
+        graph.setStringValue(transactionContentAttributeId, txId, "Can you do this?\nWill you  DO  that?");
+
+        ExtractWordsFromTextPlugin instance = new ExtractWordsFromTextPlugin();
+        Graph graph1 = new DualGraph(graph.getSchema(), graph);
+
+        PluginParameters parameters = instance.createParameters();
+        parameters.getParameters().get(ExtractWordsFromTextPlugin.ATTRIBUTE_PARAMETER_ID).setStringValue(ContentConcept.TransactionAttribute.CONTENT.getName());
+
+        instance.updateParameters(graph1, parameters);
+        assertTrue(parameters.hasParameter(ExtractWordsFromTextPlugin.ATTRIBUTE_PARAMETER_ID));
```
antares1470, line 617: Having had a quick look at the function being tested here, I think it would be good to have some asserts confirming the function worked as intended. Currently, I don't think the assert here is adding anything as the function doesn't appear to be adding or removing that parameter (so it will always be true)

antares1470 (changes requested):

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractWordsFromTextPluginNGTest.java
```diff
@@ -549,4 +595,40 @@ public void testExtractWordFromContentRegexOnlyLowerCaseFalseParameter() throws
...
+        parameters.addParameter(useRegexParameter);
+
+        Graph graph1 = new DualGraph(graph.getSchema(), graph);
+        instance.updateParameters(graph1, parameters);
+
+        assertEquals(parameters.getParameters().size(), 3);
+        assertTrue(parameters.hasParameter(ExtractWordsFromTextPlugin.ATTRIBUTE_PARAMETER_ID));
+        assertEquals(parameters.getParameters().get(ExtractWordsFromTextPlugin.WORDS_PARAMETER_ID).getStringValue(), "text");
+        assertEquals(parameters.getParameters().get(ExtractWordsFromTextPlugin.USE_REGEX_PARAMETER_ID).getBooleanValue(), true);
+        assertTrue(!parameters.hasParameter(ExtractWordsFromTextPlugin.SELECTED_ONLY_PARAMETER_ID));
```
antares1470, line 626: `assertFalse(condition)` probably better used here (rather than `assertTrue(!condition)`)

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):

PR #1181, opened 2021-08-10
`GlobalParametersPane` Tests

formalhaunt:
### Description of the Change

These tests will only run when a display is present. If the build is headless, a skipping exception will be thrown which will cause the class to be skipped.

Tests are awesome!

### Benefits

Improves test coverage.

### Possible Drawbacks

Opens a super annoying frame when the tests run.

### Applicable Issues


antares1470 (approved):

arcturus2 (approved):
@formalhaunt we've generally started our test methods with the word "test' so `testGetParams` but given its not properly documented its not a major concern.

Also if you didn't know, you can right click on a class -> navigate -> go to test class and that will create a boiler plate for you if the test class does not exist. You just need to select TestNG from the list.

PR #1183, opened 2021-08-11
`MergeNodesPlugin` Unit Tests

formalhaunt:
### Description of the Change

Unit tests for `MergeNodesPlugin` class. Added a test concrete implementation of `MergeNodeType`.  This gets picked up in the scan by the NB `Lookup`. As it is only in the test packages it will not be included in the final distribution and so is only available in the tests.

### Alternate Designs

Tried mocking the static call to `Lookup` but it proved difficult when the test was run with the other test classes. Something else must call the `MergeNodesPlugin` class before I set up the mocks and pre-populate the `static final` global variables essentially preventing the mocking from working.

Tests are awesome!

### Benefits

Improves code coverage.

### Applicable Issues

### DIFF
--- a/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPluginNGTest.java
+++ b/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPluginNGTest.java
@@ -1,0 +1,274 @@
+/*
+ * Copyright 2010-2021 Australian Signals Directorate
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.views.dataaccess.plugins.clean;
+
+import au.gov.asd.tac.constellation.graph.GraphWriteMethods;
+import au.gov.asd.tac.constellation.graph.schema.visual.VisualSchemaPluginRegistry;
+import au.gov.asd.tac.constellation.plugins.PluginException;
+import au.gov.asd.tac.constellation.plugins.PluginExecution;
+import au.gov.asd.tac.constellation.plugins.PluginInteraction;
+import au.gov.asd.tac.constellation.plugins.parameters.PluginParameter;
+import au.gov.asd.tac.constellation.plugins.parameters.PluginParameterController;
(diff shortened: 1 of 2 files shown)


arcturus2 (approved):
Really neat and a lot of effort put in here. Thank you @formalhaunt.

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPluginNGTest.java
```diff
@@ -0,0 +1,274 @@
...
+        final PluginParameters actual = mergeNodesPlugin.createParameters();
+        
+        final Map<String, PluginParameter<?>> parameters = actual.getParameters();
+        
+        assertEquals(parameters.keySet(), Set.of(
+                "MergeNodesPlugin.merge_type",
+                "MergeNodesPlugin.threshold",
+                "MergeNodesPlugin.merger",
+                "MergeNodesPlugin.lead",
+                "MergeNodesPlugin.selected"));
```
antares1470, line 102: I think it would be better to use the constant strings, e.g. `MergeNodePlugin.MERGE_TYPE_PARAMETER_ID`

CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPluginNGTest.java
```diff
@@ -0,0 +1,274 @@
...
+        
+        final PluginParameter<SingleChoiceParameterValue> mergeTypeParameter = 
+                (PluginParameter<SingleChoiceParameterValue>) parameters.get("MergeNodesPlugin.merge_type");
```
antares1470, line 105: similar here and other places where the parameter name is used

antares1470 (changes requested):

aldebaran30701 (approved):

arcturus2:
@formalhaunt would you be able to have a look at the feedback given and respond when you can?

formalhaunt:
Just saw these. My opinion is that you shouldn't use production code to in your code test. Basically if something changes in the production code the tests should break. By referencing the constants in the production code you prevent this from happening.
I found this article that makes my point in probably better words https://dev.to/scottshipp/don-t-use-non-test-constants-in-unit-tests-3ej0

antares1470:
> 
> 
> Just saw these. My opinion is that you shouldn't use production code to in your code test. Basically if something changes in the production code the tests should break. By referencing the constants in the production code you prevent this from happening.
> I found this article that makes my point in probably better words https://dev.to/scottshipp/don-t-use-non-test-constants-in-unit-tests-3ej0
> In this case its probably not a big deal as they are id strings but imho I think its good practice to follow where possible.

I guess for me personally it comes down to the purpose of the test. If the value of the constant is important for the test then it makes sense to reference the value itself rather than the constant. On the other hand, if the value of a constant isn't that important to the test then I think using the constant may be better in the long term because in those instances, if you're changing the value of the constant you probably want that reflected in the test anyway.

But I get your point. I'd be the first to admit that I'm definitely getting picky here with my suggestion so I can approve this as is

antares1470 (approved):

PR #1200, opened 2021-08-13
✅ Test boolean parameter type

aldebaran30701:
### Description of the Change
Added some testing of the boolean parameter type
No issues found with the current implementation.

### DIFF
--- a/CorePluginFramework/test/unit/src/au/gov/asd/tac/constellation/plugins/parameters/types/BooleanParameterTypeNGTest.java
+++ b/CorePluginFramework/test/unit/src/au/gov/asd/tac/constellation/plugins/parameters/types/BooleanParameterTypeNGTest.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright 2010-2021 Australian Signals Directorate
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


CorePluginFramework/test/unit/src/au/gov/asd/tac/constellation/plugins/parameters/types/BooleanParameterTypeNGTest.java
```diff
@@ -0,0 +1,240 @@
...
+    public void tearDownMethod() throws Exception {
+    }
+
+    /**
+     * Test of build method, of class BooleanParameterType.
+     */
+    @Test
+    public void testBuild_String() {
+        System.out.println("build_string");
+        String id = "booleanParameter";
```
arcturus2, line 62: You could reuse the `BooleanParameterType.ID` instead but did you pick a different string on purpose to show that it should still work?

arcturus2 (approved):

CorePluginFramework/test/unit/src/au/gov/asd/tac/constellation/plugins/parameters/types/BooleanParameterTypeNGTest.java
aldebaran30701: I just passed any string in to ensure it was set correctly in the constructor. Could be a good idea to use the .ID but I would also want to test this to ensure it wasn't hardcoding anything. (Choosing a random string like "booleanParameter" makes it unlikely)

antares1470 (approved):

PR #1210, opened 2021-08-16
Adding tags to Arrangement plugins

arcturus2:
### Description of the Change

Adding tags to arrangement plugins. Note that I have removed the "LOW LEVEL" tag for some of them - keen to know what you think?

Some minor code smells in a couple of places have been fixed.


antares1470 (approved):
I agree with removing the low level tag. It didn't fit

aldebaran30701 (approved):

PR #1214, opened 2021-08-16
add tagging for core graph node

aldebaran30701:
Add tagging for core graph node.


arcturus2 (commented):
Not sure on this one. This feels border line `UTILITY` as what we are doing is loading and saving a template which is a preference to a defined graph schema.
Thoughts?

aldebaran30701:
Agreed, much better suited. I've changed the pluginType to also better reflect things as it essentially exports it to a preference / file or loads it (imports)

antares1470 (approved):

arcturus2 (approved):

PR #1225, opened 2021-08-16
Update plugin tags for MapView

Delphinus8821:
### Description of the Change
Update to plugin tags in CoreMapView.

### DIFF
--- a/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/plugins/CopyCustomMarkersToGraphPlugin.java
+++ b/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/plugins/CopyCustomMarkersToGraphPlugin.java
@@ -27,7 +27,6 @@ import au.gov.asd.tac.constellation.plugins.PluginExecution;
 import au.gov.asd.tac.constellation.plugins.PluginInfo;
 import au.gov.asd.tac.constellation.plugins.PluginInteraction;
 import au.gov.asd.tac.constellation.plugins.PluginNotificationLevel;
-import au.gov.asd.tac.constellation.plugins.PluginType;
 import au.gov.asd.tac.constellation.plugins.parameters.PluginParameters;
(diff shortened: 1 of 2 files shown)


CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/plugins/CopyCustomMarkersToGraphPlugin.java
```diff
@@ -49,7 +48,7 @@
  * @author cygnus_x-1
  */
 @ServiceProvider(service = Plugin.class)
-@PluginInfo(pluginType = PluginType.COPY, tags = {"COPY"})
+@PluginInfo(tags = {"MODIFY"})
 @NbBundle.Messages("CopyCustomMarkersToGraphPlugin=Copy Custom Markers to Graph")
 public class CopyCustomMarkersToGraphPlugin extends SimpleEditPlugin {
```
antares1470, line 53: You've gone with MODIFY which could be fine but in my head I could also make an argument for IMPORT (from map) or CREATE. Thoughts @arcturus2?

aldebaran30701: I can see that method, happy with either from my end.

aldebaran30701 (approved):

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/plugins/CopyCustomMarkersToGraphPlugin.java
arcturus2: Hi @antares1470 @aldebaran30701! The input to this seems to be the custom marks from the Map View which is still internal to Constellation so I would keep "MODIFY". If it was a file import then that's where I'd prefer using IMPORT.

arcturus2 (approved):

antares1470 (approved):

Issue #1126, opened 2021-06-29
Add tags to plugins


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Some plugins have had tags added to them (e.g. Low Level, Selection) but the vast majority of plugins are missing these. For those plugins, we should look to add tags as tags enable better filtering in views like the Plugin Reporter

### Steps to Reproduce

1.  \[First Step\]

2.  \[Second Step\]

3.  \[and so on…\]

**Expected behaviour:** All plugins are tagged

**Actual behaviour:** Only some plugins are tagged

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

GammaVel:
The new tags in the plug in reporter are great. Just wondering if down the track there might be too many to display nicely in the drop down filer (as it expands with more tags) but that is a separate issue.

arcturus2:
@GammaVel the intention would be to limit the amount of tags we have so this shouldn't be a major problem. Thank you for the question though.

PR #1237, opened 2021-08-18
Adding Tags to plugins in Core Import Export

Auriga2:
### Description of the Change
Added static classes with plugin Tags to anonymous plugins in the Core Import Export Plugins module

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1126

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
@@ -83,7 +83,7 @@ public class FileDropper implements GraphDropper {
         return null;
     }
 
-    @PluginInfo(pluginType = PluginType.IMPORT, tags = {"LOW LEVEL"})
+    @PluginInfo(pluginType = PluginType.IMPORT, tags = {"IMPORT"})
     public static class DragAndDropFilePlugin extends SimplePlugin {
 
         private final File file;
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
@@ -71,30 +71,7 @@ public class FileDropper implements GraphDropper {
                     final List<File> files = (List<File>) data;
                     files.stream().forEach(file -> {
                         if (file.isFile()) {
-                            PluginExecution.withPlugin(new SimplePlugin("Drag and Drop: File to Graph") {
-                                @Override
-                                public void execute(final PluginGraphs graphs, final PluginInteraction interaction, final PluginParameters parameters) throws InterruptedException, PluginException {
-                                    try {
-                                        final Graph g = new GraphJsonReader().readGraphZip(file, new HandleIoProgress(String.format("Reading %s...", file.getName())));
-                                        GraphOpener.getDefault().openGraph(g, file.getName());
-
-                                        final ReadableGraph rg = g.getReadableGraph();
-                                        try {
-                                            ConstellationLoggerHelper.importPropertyBuilder(
-                                                    this,
-                                                    GraphRecordStoreUtilities.getVertices(rg, false, false, false).getAll(GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.LABEL),
-                                                    Arrays.asList(file),
-                                                    ConstellationLoggerHelper.SUCCESS
-                                            );
-                                        } finally {
-                                            rg.release();
-                                        }
-                                    } catch (final IOException | GraphParseException ex) {
-                                        Logger.getLogger(FileDropper.class.getName()).log(Level.WARNING, String.format("Error loading file %s: %s", file.getPath(), ex.getMessage()));
-                                        NotifyDisplayer.display("Error loading graph: " + ex.getMessage(), NotifyDescriptor.ERROR_MESSAGE);
-                                    }
-                                }
-                            }).executeLater(graph);
+                            PluginExecution.withPlugin(new DragAndDropFilePlugin(file)).executeLater(graph);
                         }
                     });
                 };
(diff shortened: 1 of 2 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
```diff
@@ -105,4 +82,42 @@ public void execute(final PluginGraphs graphs, final PluginInteraction interacti
 
         return null;
     }
+
+    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"MODIFY"})
+    public static class DragAndDropFilePlugin extends SimplePlugin {
```
antares1470, line 87: given this plugin is opening a file via drag and drop, I'm not sure MODIFY is the tag to use. Maybe UTILITY (or GENERAL if nothing else fits)? Plugin type might need to be different as well

antares1470 (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
aldebaran30701: not import ?

Auriga2: The parent plugin is an IMPORT. Should these also be Import Type and IMPORT
tag?  
PluginType.CREATE, tags = {"IMPORT"} Or 
PluginType.IMPORT, tags = {"IMPORT"}

Auriga2: OR PluginType.IMPORT, tags = {"UTILITY "}?

aldebaran30701: Personally I'd say PluginType.IMPORT, tags = {"IMPORT"} fits best

antares1470: @Auriga2 I'd go with however we have tagged opening graphs elsewhere (bar WELCOME since that wouldn't make sense in this instance). I did think about IMPORT but thought that this really opening a graph rather than importing one. If IMPORT though is how we've tagged it elsewhere then go with that

Auriga2: I am not sure how to find these. I can see some with PluginType.DISPLAY, tags = {"MODIFY"}.
Shall I add multiples in that case? Say PluginType.CREATE, tags = {"IMPORT", "UTILITY "}?

antares1470: `OpenFilePlugin` is tagged as `@PluginInfo(pluginType = PluginType.IMPORT, tags = {"LOW LEVEL"})`. I'd say that is probably the most similar action so I would go with that

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/JsonDropper.java
```diff
@@ -86,10 +86,10 @@
         return null;
     }
 
-    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"MODIFY"})
+    @PluginInfo(pluginType = PluginType.IMPORT, tags = {"LOW LEVEL"})
     public static class DragAndDropJSONPlugin extends SimpleEditPlugin {
```
antares1470: I think this function is actually adding to an existing graph (rather than opening a graph like the other function) so what you tagged it with previously is ok

antares1470 (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
```diff
@@ -105,4 +82,42 @@ public void execute(final PluginGraphs graphs, final PluginInteraction interacti
 
         return null;
     }
+
+    @PluginInfo(pluginType = PluginType.IMPORT, tags = {"LOW LEVEL"})
```
arcturus2, line 86: I'd go with `IMPORT` as this is adding data from an external file (via drag and drop) onto the graph.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/JsonDropper.java
```diff
@@ -110,4 +85,44 @@ public void edit(final GraphWriteMethods wg, final PluginInteraction interaction
 
         return null;
     }
+
+    @PluginInfo(pluginType = PluginType.IMPORT, tags = {"LOW LEVEL"})
```
arcturus2, line 89: Ditto `IMPORT` for me on this one too.

arcturus2 (changes requested):
Sorry but I would go with `IMPORT` for both of them

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
arcturus2: Sorry but I would go with `IMPORT` because they are part of the same plugin. It would also mean that when you filter in the notes view for `IMPORT` then there will still be a lingering plugin for `MODIFY` which was for the same action. That is, effectively you have both `IMPORT` and `MODIFY` as tags.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/JsonDropper.java
arcturus2: Similarly, I would suggest `IMPORT` as otherwise you have both an `IMPORT` and `MODIFY` tags effectively for this action.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
Auriga2: Not Clear, do you mean ` PluginType.IMPORT, tags = {"IMPORT"}`?

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/JsonDropper.java
Auriga2: Sorry do you mean ` PluginType.IMPORT, tags = {"IMPORT"}`?

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/FileDropper.java
antares1470: @arcturus2 if that is the case, should we change the `OpenFilePlugin` tags then? These are more or less doing the same thing (one opens graph via file explorer, and the other via drag and drop), so having different tags for the pair them doesn't make sense

arcturus2: Yep @antares1470, opening a file would be `IMPORT` to my mind.

> Not Clear, do you mean PluginType.IMPORT, tags = {"IMPORT"}?

Yes @Auriga2

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/dragdrop/JsonDropper.java
arcturus2: Yeah @Auriga2

antares1470 (approved):

arcturus2 (approved):

Issue #1022, opened 2021-04-28
Fix SonarCloud Issues


antares1470:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

With the recent work done to move to Github Actions, we now have CI again which means we once again have SonarCloud picking up issues with the code base. Since the last time we had SonarCloud regularly reviewing the code, we have seen an increase in detected issues (note that this is partially due to new rules being added). It would be good then to take some time to go through and reduce these numbers again.

Ideally we will aim for the following:
- as close to 0 bugs as we can
- as close to 0 vulnerabilities as we can 
- \>= 80% security hotspots reviewed
- code smells down as low as possible
- < 3% code duplication

https://sonarcloud.io/dashboard?id=constellation-app_constellation

### Additional Information

Anyone working on this ticket should add a comment stating which rules they are working on so that there isn't any accidental doubling up

antares1470:
Working on cleaning up the following rules:
- Math operands should be cast before assignment

Delphinus8821:
Working on cleaning up code smells in CoreHelp.

aldebaran30701:
I'm working my way through the MapView package for any code smells I can resolve

aldebaran30701:
Working on:
- Add a private constructor to hide implicit public

Delphinus8821:
Working on:
- final classes should not have protected members 
- String literals should be placed on the left side when checking for equality

Delphinus8821:
Working on:
- Redundant casts should not be used

Auriga2:
Working on:

- Refactor the code to eliminate nested `switch`
- Exceptions should be either logged or rethrown but not both
- Open curly brace locate at the end of a line
- Local variables should not be declared and then immediately returned or thrown

Auriga2:
- Remaining: static base class members should not be accessed via derived types (Except Map.Entry)
- String literals should not be duplicated
- New:"String#replace" should be preferred to "String#replaceAll"

PR #1253, opened 2021-08-20
Sonar cleanup

antares1470:
### Description of the Change

Cleaned up the following Sonar rule hits:
- Math operands should be cast before assignment
- "static" base class members should not be accessed via derived types
- "String#replace" should be preferred to "String#replaceAll"
- Whitespace and control characters in literals should be explicit
- "static" members should be accessed statically
- "entrySet()" should be iterated when both the key and value are needed

It looks like a lot but a few of those only had 1 or 2 hits.

In the replace v replaceAll instance, it appears that the strings"\\\\;" and "\\\\:" were meant to be literal rather than regex (based on other parts of the functions they were contained in and the fact that replacing a string with itself makes no sense).

Removing bugs and code smells from Core

### Benefits

Removing bugs and code smells from Core

### Applicable Issues

#1022

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/visual/renderables/HitTester.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/visual/renderables/HitTester.java
@@ -22,6 +22,8 @@ import au.gov.asd.tac.constellation.visual.opengl.renderer.GLRenderable;
 import au.gov.asd.tac.constellation.visual.opengl.renderer.GLVisualProcessor;
 import au.gov.asd.tac.constellation.visual.opengl.utilities.GLTools;
 import com.jogamp.common.nio.Buffers;
+import com.jogamp.opengl.GL;
+import com.jogamp.opengl.GL2ES3;
 import com.jogamp.opengl.GL3;
 import com.jogamp.opengl.GLAutoDrawable;
 import java.nio.FloatBuffer;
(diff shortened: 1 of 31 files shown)


CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/visual/renderables/HitTester.java
```diff
@@ -64,7 +66,7 @@
     // The buffer to read from. This is hardcoded as their seems to be no real reason to change it,
     // but perhaps it should be looked up from the corersponding GraphRenderable. At the moment it simply
     // matches the buffer name used inside the if(doHitTesting) {} block of GraphRenderable's display method.
-    private final int hitTestBufferName = GL3.GL_COLOR_ATTACHMENT0;
```
aldebaran30701, line 67: More curious as to what is happening here, was it picked up in Sonar?
Is it the below rule?
"static" base class members should not be accessed via derived types

aldebaran30701 (approved):

arcturus2 (approved):

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/visual/renderables/HitTester.java
antares1470: @aldebaran30701 Yes that is the rule

arcturus2: yep

PR #1265, opened 2021-08-23
Move Notes View to Views menu

sol695510:
### Description of the Change
Moved Notes View to Views menu.

### Alternate Designs
N/A

This is a change to a core view.

### Benefits
The Notes View will be more visible to users.

### Possible Drawbacks
N/A

### Verification Process
N/A

### Applicable Issues
N/A


arcturus2 (approved):
Thanks @sol695510

PR #1276, opened 2021-08-24
:white_check_mark: Add tests for GraphicsCardUtilities

groombridge34a:
### Description of the Change

Add unit tests for the GraphicsCardUtilities class.

### Alternate Designs

The previous one - not having tests. It's worse.

Testing is awesome.

### Benefits

Improve code coverage in the core.

### Possible Drawbacks

Full test suite takes marginally more time to run.

### Verification Process

Ran the tests.

### Applicable Issues

#413

### DIFF
--- a/CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/GraphicsCardUtilitiesNGTest.java
+++ b/CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/GraphicsCardUtilitiesNGTest.java
@@ -1,0 +1,313 @@
+/*
+ * Copyright 2010-2021 Australian Signals Directorate
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.utilities;
+
+import au.gov.asd.tac.constellation.utilities.text.SeparatorConstants;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import org.mockito.MockedStatic;
+import org.testng.annotations.Test;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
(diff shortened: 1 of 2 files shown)


CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/GraphicsCardUtilitiesNGTest.java
```diff
@@ -0,0 +1,313 @@
...
+import org.testng.annotations.AfterMethod;
+
+/**
+ * Unit tests for the GraphicsCardUtilities class.
+ * 
+ * @author groombridge34a
+ */
+public class GraphicsCardUtilitiesNGTest {
+    
+    private static final String DXDIAG_FILENAME = System.getProperty("user.home") + "/dxdiag.txt";
```
arcturus2, line 45: Given this is the same location Constellation writes to `dxdiag.txt` when its run I was wondering if it would be better to isolate the test file and instead create a temp file like ` File.createTempFile("dxdiag", "txt").getName()`? Or something to that affect?

arcturus2 (changes requested):
Thank you @groombridge34a for this. I've added a comment for you to consider. Also there was a failed unit test that I accidentally introduced to `master` but that has been fixed so if you merge `master` your branch should run fine.

antares1470 (changes requested):
The only thing I would suggest is moving from catch-and-fail to simply throwing the exceptions. This is what we've done elsewhere and has the same effect

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/GraphicsCardUtilitiesNGTest.java
groombridge34a: Unfortunately `System.getProperty("user.home") + "/dxdiag.txt"` is hard-coded in the class under test (`GraphicsCardUtilities`) and Mockito doesn't let allow mocking static methods in `System`. Changing the class is the only solution I can think of. Since the class is never instantiated (ie never uses a constructor) I think a setter is the only way to avoid a deep refactor? The simplest thing could be to make the location of `dxdiag.txt` a member variable and add a setter method?

Alternatively we could try changing `GraphicsCardUtilities` to create `dxdiag.txt` in a temp file. But for some reason the class explicitly creates only one `dxdiag.txt` file per life of the application VM (see the `loaded` variable). It's unlikely, but conceivable, that a temp file might be removed by the OS while Consty is still running. And the class doesn't document why this behaviour is important... although it's probably because it's a low chance that the user would be hot-swapping video cards :-)

arcturus2: Thanks @groombridge34a. For some reason I thought you were mocking what was in that file so it didn't matter where the location of the file was. It sounds like your tests is expecting that file to exist.

Though not documented, the reason that location is fixed is because that is the default location when `dxdiag /64bit /t` writes the file writes to when executed.

Given you need the file there for your tests I think its fine as you have it.

Ah yes, an optimisation I wanted to do for a while was to not run `dxdiag` every time you run Constellation based on the user not likely to swap cards!

arcturus2 (approved):

PR #1280, opened 2021-08-24
Headless Build

formalhaunt:
### Description of the Change

Added TestFX and Monocle to the deps. This should hopefully allow for a headless build to run over JavaFX code.  Added required system props to the run test script so that the build runs in headless mode.

The PR contains one example of how to write the tests as a proof of concept.

This will mean that all the tests we have written that involve JavaFX code can now be run on the CI and included in coverage stats.

### Benefits

All tests run as a part of the CI build.

### Possible Drawbacks

The Monocle dep is based on the JDK version. This could cause some issues when we look at upgrading to the next LTS. No one seems to have touched the GitHub repo for a while but I am not sure what is involved in making a new release. It is part of the TestFX package so hopefully it will be kept up to date as needed.

### DIFF
--- a/.github/workflows/build-and-scan.yml
+++ b/.github/workflows/build-and-scan.yml
@@ -14,7 +14,7 @@ jobs:
     name: Unit Tests
     runs-on: ubuntu-latest
     env:
-      RUNNER_IMAGE: "constellationapplication/netbeans-runner:12"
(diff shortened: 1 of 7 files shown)


antares1470 (approved):

arcturus2 (approved):
Many thanks for this one @formalhaunt !

PR #1282, opened 2021-08-24
Enable Display Tests

formalhaunt:
### Description of the Change

Removed the isHeadless check from tests so that they will now run as the headless CI issue has been resolved.

Tests are awesome and they should be run.

### Benefits

Improve code coverage.


arcturus2 (approved):

arcturus2:
This really helps give a more accurate picture of our code coverage. Thank you for going the extra mile and adding more tests as well.

antares1470 (approved):

PR #1289, opened 2021-08-25
Five sec write lock action fix

antares1470:
### Description of the Change

The Five second write lock action was calling the five second read lock plugin instead of its namesake. Have fixed this.

Fix bug in core

### Benefits

Fix bug in core


arcturus2 (approved):
Thank you.

PR #1291, opened 2021-08-25
restored graph construction demo to build graph menu

antares1470:
### Description of the Change

The graph construction demo had its menu item link commented out an was therefore only available via search. Since nothing appears to be superseding it, this PR is putting it back on the menu

This is already a thing, now it's properly accessible

### Benefits

Easier to access

### Possible Drawbacks

Maybe some people liked that it was a hidden easter egg?


arcturus2 (approved):
Thank you @antares1470

PR #1303, opened 2021-08-26
:fire: Remove unused code from ImmutableObjectCache

groombridge34a:
### Description of the Change

ImmutableObjectCache included a second, more complex version of the code which is unreachable.

### Alternate Designs

Whether the simple or complex code is enabled could be controlled by changing the VERBOSE field to a flag controlled by configuration.

This code has never been reachable during the commit history of this project. Keeping dead code "just because" increases the complexity of the code.

### Benefits

Reduced code complexity and technical debt.

### Possible Drawbacks

This code was not covered by unit tests, it's possible that regression errors could be introduced.

### Verification Process

Created regression tests

### Applicable Issues

#413

### DIFF
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/ImmutableObjectCache.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/ImmutableObjectCache.java
@@ -15,10 +15,8 @@
  */
 package au.gov.asd.tac.constellation.utilities.datastructure;
 
-import au.gov.asd.tac.constellation.utilities.text.SeparatorConstants;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Map.Entry;
 
 /**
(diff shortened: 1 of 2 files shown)


groombridge34a:
> 
> 
> I think the `VERBOSE` was left there for debugging when deving locally. @cygnus-x-1 let me know your thoughts on this PR.

arcturus2 (changes requested):
@groombridge34a wondering if you don't mind if we keep the code and use `java.util.logging.Logger` and `FINE` them instead? It will help with debugging and it is used by the IOProvider's.

arcturus2:
@groombridge34a  would you be able to have a look passing the "verbose" as a parameter or something like that to test it out? I am looking at a performance issue and the verbose logging in here is useful so I would like to keep it if thats ok?

groombridge34a:
As discussed, this code is useful for debugging and should be exposed with a code conditional of some kind. Closing this PR and deleting the branch, will proceed in a new PR.

Issue #1264, opened 2021-08-23
Running Throb Nodes on empty graph causes a crash


Atlas139mkm:
### Description

Running Throb Nodes on an empty graph causes an immediate crash

### Steps to Reproduce

1.  create new graph

2.  Experimental/Animations/Throb Nodes

3.  Crash

**Expected behaviour:** Either a warning be thrown to the user or nothing

**Actual behaviour:** Constellation crashes

**Reproduces how often:** 100%

arcturus2:
@Atlas139mkm assume this was picked up during testing the latest release candidate?

Atlas139mkm:
Yeah it was

arcturus2:
Thanks @Atlas139mkm for finding this issue. I gave it a test in v2.3 and confirmed it was an issue there also. Therefore I've gone ahead and removed the v2.4.0-rc1 I added as I am going to reserve this tag for issues that only occur in the RC testing. Ideally I would prefer we don't introduce new bugs that were not there from the previous release and this tag will help to catch those.

GammaVel:
Thank you @Atlas139mkm for identifying this crash and fixing it !!! Nice work :-)

Issue #1262, opened 2021-08-23
Created nodes not visible after fly-through on empty graph


Atlas139mkm:
### Description

After doing a fly-through on an empty graph, then stopping the fly through, trying to add nodes to the graph creates them however they cannot be seen.

### Steps to Reproduce

1.  Create a new empty graph

2.  Experimental/Animations/Fly-Through

3.   Experimental/Animations/Stop Animating

4.  Toggle selection mode to draw mode

5.  click on the graph to add nodes (the node count in the bottom left should increase, however they will not visually appear)

**Expected behaviour:** It would be expected that the graph would continue with its normal behavior after stopping the fly-through

**Actual behaviour:** It doesn't show the nodes visually when creating them after stopping the fly through animation.

**Reproduces how often:** 100%

### Additional Information

If you stop the animation straight after starting it, the issue will not be produced. There is about a 2 second window to stop the animation to maintain the graph working.

arcturus2:
@Atlas139mkm assume this was picked up during testing the latest release candidate?

Atlas139mkm:
Yeah it was

arcturus2:
Thanks @Atlas139mkm for finding this issue. I gave it a test in v2.3 and confirmed it was an issue there also. Therefore I've gone ahead and removed the v2.4.0-rc1 I added as I am going to reserve this tag for issues that only occur in the RC testing. Ideally I would prefer we don't introduce new bugs that were not there from the previous release and this tag will help to catch those.

GammaVel:
I tried this out with the latest release candidate and this fixes the issue. Thanks again @Atlas139mkm for finding and fixing !!

PR #1305, opened 2021-08-27
fix for flying and throbbing animation crashes

Atlas139mkm:
### Description of the Change

I have a added a fix for both the Flying animation and the throbbing animation causing crashes with no vertex's on the graph. The solution added:
-  A check in the ThrobNodeAnimation class that ensures there is at least 1 vertex before attempting to start the animation. 
- A check in the FlyingAnimation class to check there is at least 2 vertex's present when calling the Flying animation.

### Alternate Designs
N/A

It avoids the code crashing when:
1: calling an animation with 0 nodes present
2: calling the Flying animation when there is 1 or less nodes.

### Benefits

Avoids users accidently crashing the application.

### Possible Drawbacks

The Flying animation will not run unless there is 2 nodes present. It will mean the user can not watch a flying animation of 1 node.

Throb animation will not run unless 1 node is present.

### Verification Process

I would test each animation with various numbers of nodes (0 nodes, 1, node and 2 nodes.)

### Applicable Issues

#1262 
#1264

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
@@ -49,66 +49,75 @@ public final class FlyingAnimation extends Animation {
 
     @Override
     public void initialise(GraphWriteMethods wg) {
+        // dont initilise the animation if there is less than 2 vertexs
+        if (wg.getVertexCount() <= 1) {
+            stopAnimation();
+        } else {
 
-        xAttr = VisualConcept.VertexAttribute.X.get(wg);
-        yAttr = VisualConcept.VertexAttribute.Y.get(wg);
-        zAttr = VisualConcept.VertexAttribute.Z.get(wg);
-        x2Attr = VisualConcept.VertexAttribute.X2.get(wg);
-        y2Attr = VisualConcept.VertexAttribute.Y2.get(wg);
-        z2Attr = VisualConcept.VertexAttribute.Z2.get(wg);
-        rAttr = VisualConcept.VertexAttribute.NODE_RADIUS.get(wg);
-        selectedAttr = VisualConcept.VertexAttribute.SELECTED.get(wg);
-        doMixing = x2Attr != Graph.NOT_FOUND && y2Attr != Graph.NOT_FOUND && z2Attr != Graph.NOT_FOUND;
-        final int cameraAttribute = VisualConcept.GraphAttribute.CAMERA.get(wg);
-        camera = wg.getObjectValue(cameraAttribute, 0);
+            xAttr = VisualConcept.VertexAttribute.X.get(wg);
+            yAttr = VisualConcept.VertexAttribute.Y.get(wg);
+            zAttr = VisualConcept.VertexAttribute.Z.get(wg);
+            x2Attr = VisualConcept.VertexAttribute.X2.get(wg);
+            y2Attr = VisualConcept.VertexAttribute.Y2.get(wg);
+            z2Attr = VisualConcept.VertexAttribute.Z2.get(wg);
+            rAttr = VisualConcept.VertexAttribute.NODE_RADIUS.get(wg);
+            selectedAttr = VisualConcept.VertexAttribute.SELECTED.get(wg);
+            doMixing = x2Attr != Graph.NOT_FOUND && y2Attr != Graph.NOT_FOUND && z2Attr != Graph.NOT_FOUND;
+            final int cameraAttribute = VisualConcept.GraphAttribute.CAMERA.get(wg);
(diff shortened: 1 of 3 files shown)


CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
```diff
@@ -49,66 +49,75 @@
 
     @Override
     public void initialise(GraphWriteMethods wg) {
+        // dont initilise the animation if there is less than 2 vertexs
+        if (wg.getVertexCount() <= 1) {
```
antares1470, line 53: More curious if the decision to not animate on on 1 node was fixing a bug or purely a design decision? If It's the latter, while I would anticipate it's unlikely someone would want to fly-through on a graph with only one node, I think I would prefer to leave that decision to the user rather than take it out of their hand

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
```diff
@@ -49,66 +49,75 @@
 
     @Override
     public void initialise(GraphWriteMethods wg) {
+        // dont initilise the animation if there is less than 2 vertexs
```
antares1470, line 52: vertexs -> nodes

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
```diff
@@ -49,66 +49,75 @@
...
+            }
         }
     }
 
     @Override
     public List<VisualChange> animate(GraphWriteMethods wg) {
-        if (step >= stepsPerLink) {
-            // Get the next p3 vertex.
-            final Vector3f xyz = getNextVertex(wg, camera.getMix());
+        // dont animatite unless there is more than 1 vertex
```
antares1470, line 86: animate*

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
Atlas139mkm: It was to fix a bug, constellation would crash if there was only 1 node on the fly though. I'm sure there is probably a more appropriate solution that stopped it crashing on 1 node however I figured this was a quicker and much easier fix.

antares1470: Yeah that's alright. Figured I'd check

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/FlyingAnimation.java
arcturus2: @antares1470 as in change the word to "nodes"? It's in code rather than GUI facing so I think its OK. Though grammatically it should be "vertices"

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/ThrobbingNodeAnimation.java
```diff
@@ -42,28 +42,40 @@
...
         }
-        return Arrays.asList(new VisualChangeBuilder(VisualProperty.VERTEX_RADIUS).forItems(wg.getVertexCount()).withId(throbbingNodeAnimationId).build());
+        // return an empty list if 0 nodes
+        return Arrays.asList();
     }
 
     @Override
-    public void reset(GraphWriteMethods wg) {
+    public void reset(GraphWriteMethods wg
+    ) {
```
arcturus2, line 78: @Atlas139mkm can you please auto-format the file? Alt+Shift+F

arcturus2 (changes requested):
Sorry about the delay in reviewing your PR @Atlas139mkm. Some minor comments exist from @antares1470 and me which has nothing to do with your coding though would be nice to have fixed. 

Thank you for the fixes as these are long standing issues we did not know existed :heart:

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/animation/ThrobbingNodeAnimation.java
antares1470: @Atlas139mkm I think you missed this bit of feedback. Once done this should be good to go

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):

Issue #1284, opened 2021-08-24
Opening a file from the Recent Icons on the Welcome Page causes unexpected exception


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

On the Recent section of the Welcome page there are handy tiles which can be clicked on to look at files that I have been working on. I have noticed I get a **DataObject.find on null error** for a couple of the files that I am working on. It doesn't happen for all the files, only a couple and I am not sure what makes those files unique. 

### Steps to Reproduce

1.  Open Constellation and open and save graph files
2.  Close constellation and reopen, check the Welcome Page. 
3.  Click on the Recent Items files  -> Some cause a **DataObject.find on null error** exception that appears at the bottom of the application
4. Open the exact same file by going to File > Open Recent -> There are no issues with opening the file from here
5. Make some changes to the graph and save. Close constellation and reopen. Click on the recent file icon for the graph again > the same error appears
6.  Reopen the graph from File Recent and this time make some changes to the graph and save it with a different name. Repeat the process of closing constellation and reopening. Access the graph from the recent icons > this time there is no error when opening the graph. 

**Expected behaviour:** Graph files open from the Welcome section or the DataObject.find on null error is handled before it causes an exception. 

**Actual behaviour:** DataObject.find on null error

**Reproduces how often:** - medium difficulty -  I can reproduce with a couple of graphs but not all. 

### Additional Information

![image](https://user-images.githubusercontent.com/83679004/130703306-c0f4eafb-4328-4657-9a40-011c8f6ced80.png)

Auriga2:
@GammaVel It would be great of you can attach a file that failed, as I could not recreate this error with some random files I worked with.

GammaVel:
[...]

PR #1311, opened 2021-08-27
Welcome Page Open Graph bug fix

antares1470:
### Description of the Change

Hopefully fixed an issue where an illegal argument exception was thrown trying to open some graphs from the welcome page. From the digging I was able to do, the issue appears to be the way that the file object is created before trying to open the graph so I've changed it what seems to be used elsewhere (including the Open Recent menu action) and hopefully that should do the trick (it seems to work on the graphs I had that I could open successfully previously but I didn't have any graphs that couldn't experienced this issue so am not sure for those ones). 

Bug fix

### Benefits

Less bugs

### Applicable Issues

#1284


arcturus2 (approved):
Seems to work fine for me thank you @antares1470

PR #1320, opened 2021-08-30
:fire: Remove unused "localized" parameter from locate method signature

groombridge34a:
### Description of the Change

The `ConstellationInstalledFileLocator` class in `CoreUtilities` includes a `locate()` method intended to find files in NetBeans modules. One of the method parameters is `boolean localized`, which was intended to narrow the search to a localized and branded lookup. But the method never uses this parameter, it's hard coded to `false`.

### Alternate Designs

None.

Minor update.

### Benefits

Increased clarity by removing unused code.

### Possible Drawbacks

None.

### Verification Process

- Used NetBeans to check for usages of the `locate()` method in the entire Core, and addressed those usages.
- Used NetBeans to check that `ConstellationInstalledFileLocator` has no occurrences of polymorphism (eg inheritance) and method overrides.
- Ran the unit test suite as problems with changing method signatures manifest as compile time errors.
- Ran Constellation, clicked around the Map View and the Scripting View.

### Applicable Issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Constellation Changes
 
+## Changes in September 2021
+
+-   Removed unused `localized` parameter from the signature of the `locate()`
+    method in `ConstellationInstalledFileLocator`.
+
 ## Changes in August 2021
 
 -   Added 'updateTagsFiltersAvailable', 'updateSelectedTagsCombo', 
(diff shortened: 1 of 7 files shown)


arcturus2 (approved):

antares1470 (changes requested):
Thanks for this @groombridge34a. If you're able to add an entry to CHANGELOG.md with the method signature change, that would be great

groombridge34a:
> 
> 
> Thanks for this @groombridge34a. If you're able to add an entry to CHANGELOG.md with the method signature change, that would be great

Done.

antares1470 (approved):

Issue #1290, opened 2021-08-25
UX improvement suggestions for notes view


Atlas139mkm:
### Description

While testing the notes view, it felt like it didn't flow too well from a UX perspective. 
- The pop up from the edit button feels a bit strange with the check boxes down the bottom indicating if you want to add or remove the note from specific graph elements.
    - This is also a problem if the user has not already highlighted the graph elements they want to add or remove the note from. If they click back to constellation to alter their selection the popup will disappear behind the constellation window. (gif provided at the bottom)
    - The 'keep original selection' checkbox is also not needed if designed correctly
- below is the current vs my suggested UI 
 
Actual look
![image](https://user-images.githubusercontent.com/88649439/130714039-6edd00ca-4619-4ba2-a783-e15e83518db7.png)

Suggested look
![image](https://user-images.githubusercontent.com/88649439/130714032-838ea2ce-6836-4e66-8ce2-9493378ac399.png)

- Changes
    - Editable text for title and content within the note (re-sizeable based on amount of text)
    - Show the number of graph elements the note is connected too at the bottom of the note, represented in the same manner as constellation has.
    - Add selected button - adds the note to the currently selected graph elements
    - Remove selected button - removes the note from the currently selected graph elements
    - Delete note button - deletes the note

- Reasoning 
    - Removes the need for the pop-up
    - user can still interact with the graph and add or remove the note from graph elements, without needing to worry about the pop up disappearing
    - icons down the bottom convey the same message in a consistent way with constellation, while taking up less space and removing text
    - Easier for user to see all notes functionality
    - should increases overall usability while maintaining all the same functionality

### Additional Information

 - adding a note to all graph elements
 - deciding I want to remove it from Vertex 0 and both transactions
 - however not all graph elements are selected
 - the edit pop up then disappears behind the constellation window when I need to change my selected elements

*Ignore pop up appearing on another window (that my computer putting it on the main window)
![Notes UI flow](https://user-images.githubusercontent.com/88649439/130713272-86ae4269-6ba6-4043-a161-c1cd14823f85.gif)

Note: I'm not sure even sure if this is possible so I'd love feedback and comments on my suggestions

arcturus2:
Love the suggestions @Atlas139mkm

GammaVel:
hi @Atlas139mkm. Just look a quick look at this now and when deleting a note there is a colon that I don't think belongs there after the question mark: 
[...]

Issue #1178, opened 2021-08-10
Notes view - update plural for user note that has been added to a single transaction or node


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

This is a bit of a trivial issue. In the new Notes view it possible to add notes to a single node or only one transaction. There is some really handy text that displays that tells the user that it has been added to a node or transaction at the end of the note. Would it be possible to change the plural casing to be singlar as currently if the note has been added to 1 node it has this text" Note has been added to 1 node**s**  and similarly *1 transaction**s**  (for one transaction). The screenshot below helps explain this.

![image](https://user-images.githubusercontent.com/83679004/128810907-2cfb75bf-fb29-4925-a1af-76ce00c92503.png)

### Steps to Reproduce

1.  Open existing graph and launch the Notes view from the experimental menu

2.  Select a single node > Add a note title and description and add it to the selected Node using the checkbox "Link note to selected" -> Check the text that is displayed on the note

3.  Repeat with a single transaction. 

**Expected behaviour:** _Note selection is linked to 1 node_

**Actual behaviour:** Note selection is added to 1 nodes (same result for where the is one transaction too)

**Reproduces how often:** 100%

GammaVel:
This is looking really good now, thanks again @Atlas139mkm for fixing up the plurals. I tried this out on an existing graph that I had that I stored some notes on and it worked. I also checked this on a new graph that I am making. Very nice!

GammaVel:
and here is a screen capture of the update: 
![image](https://user-images.githubusercontent.com/83679004/133956655-a7e5043f-89f6-4313-8f25-9ce7ea945adf.png)

PR #1325, opened 2021-08-31
Improve notes view UI and fix introduced bug

Atlas139mkm:
Please give lots of advice as I want to ensure what I have done is up to standard.

### Description of the Change

1:
Altered the notes view to remove the need for the pop up while retaining and improving existing usability and function.
- no more pop up
- added "add selected" and "remove selected" button to the note
- altered functionality of "edit" button to allow editing of the note directly instead of via a pop up
- note saves if its in edit or view state, even if the user closes notes view or changes the nodes/transactions it's linked to.
- altered linked to text to state "node" and "transaction" if there is only 1 element, "nodes" and "transactions" if there are multiple (requested in #1178)

2:
- fixed an issue from #1316 that was merged to master today, where creating a new graph would carry notes over from the previously created graph to the new one.

Main changes in the code are new buttons, removed pop up, added a variable to NotesViewEntry that states if the note is in edit or view mode and fixed the #1316 issue (this can be seen in master at the moment).

### Preview of New Notes View
Notes video demo

### Alternate Designs

- Potentially adding right click to edit the text areas
- Change the colours of the buttons to be a traffic light style (blue edit, green add, orange remove, red delete)
- reducing buttons on the note by combing some (I couldn't think of a way to do this without pop ups)

- It fixes the recent bug merged
- It improves usability of the notes view
- It maintains all existing functionality

### Benefits

- Increased usability
- removal of bug

### Possible Drawbacks

- The size of the editable text box on the note can sadly not be increased in height
   - There was a solution to this using a InlineCssTextArea, however this slowed creation of the first note by at least 3 seconds (unsure why so I went with speed performance over a dynamically resizing text area)

### Verification Process

- I pressed every combination of buttons on the note I could think of.
- I tested various combinations of notes with no or null text areas
- I imported graphs with existing notes
- I removed and added the selected nodes and transactions while the node was in edit mode
- I made new graphs
- I looked at the other notes features such as the filtering and ensured the notes still appeared 
- I re-tested the same process that caused the original bug for #1316 

### Applicable Issues

#1290
#1178
#1316


antares1470:
@Atlas139mkm just a request but are you possibly able to add the bug fix as a separate PR? I think it would be good to get that in before v2.4 is released

sol695510 (approved):

antares1470 (changes requested):
I think it would be nice to find a way to make the notes themselves smaller. While adding the buttons to user notes removes pop ups, it makes the user notes a lot larger than they need to be and therefore requires a user to do more scrolling through notes which is less than ideal. The sizing issue also affects auto notes which definitely isn't ideal.

NotesViewNoteSize

One option for reducing size could be to merge the 'Add Selection' and 'Remove Selection' buttons into one button. This will require adding submenus so that you can pick whether to add or remove. An alternative might be moving the selection change functionality to the right-click menu (which already contains viewing selection on the graph). Just some thoughts

Atlas139mkm:
Thanks for the feedback @antares1470! I'll get onto the changes in a bit. I'll see what I can do about adding the add and remove to a right click option. That seems like a much better way to do it too keep the size down.

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -595,53 +597,91 @@ private void createNote(final NotesViewEntry newNote) {
...
+        // Define selection label
+        String selectionLabelText = "";
+        final Label selectionLabel = new Label(selectionLabelText);
+
         // If the note is user created add the selection details.
         if (newNote.isUserCreated()) {
-            String selectionLabelText;
             if (newNote.isGraphAttribute()) {
-                selectionLabelText = "Note selection is linked to the graph.";
+                selectionLabelText = "Note is linked to: the graph.";
```
arcturus2, line 633: How about "Note linked to: Graph" to reduce the amount of words we have to see here on the note?

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -595,53 +597,91 @@ private void createNote(final NotesViewEntry newNote) {
...
             } else {
-                selectionLabelText = "Note selection is linked to " + newNote.getNodesSelected().size() + " nodes and "
-                        + newNote.getTransactionsSelected().size() + " transactions.";
+                selectionLabelText = "Note is linked to: ";
```
arcturus2, line 635: Similarly that would then make this "Note linked to: "

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewTopComponent.java
```diff
@@ -116,7 +116,7 @@ protected void handleComponentOpened() {
     @Override
     protected void handleComponentClosed() {
         super.handleComponentClosed();
-        notesViewPane.closeEdit();
+        //notesViewPane.selectAllFilters();
```
arcturus2, line 119: Given the PR is going into `master` please remove dead commented code if you don't need it.

arcturus2 (changes requested):
@Atlas139mkm this is really neat and thank you very much for thinking how this could be improved and also implementing a neat solution.

I have left some comments for you but besides them, can you please look into adding an "Are you sure?" type message when a user clicks to delete the note? I think in terms of UX this is important. Noted also that I am happy for this to be put into a ticket to be worked on later as this is technically not part of anything introduced in your PR.

Let me know your thoughts.

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
Atlas139mkm: Yep I agree, ill change it.

Atlas139mkm:
Awesome thanks @arcturus2, ill make your requested changes can look into a nice way of displaying an are you sure message.

Atlas139mkm:
Actually thoughts on this?

Delete button would convert to disabled and change to "are you sure?" (I would probably disable the edit button too until they pick yes or no)

no and yes buttons would appear and yes would delete it, no would covert the buttons back to enabled and remove the no and yes buttons.

image

arcturus2:
@Atlas139mkm that's pretty cool, I wouldn't have thought of that as a way to present it. I am open to try this idea out if you like.

For me, I was thinking of going down the traditional pop up dialog using a `new Alert(Alert.AlertType.CONFIRMATION)` to get the OK/Cancel response given that's what I am used to seeing.

Atlas139mkm:
I'm happy to do either way. I figured with this way it avoids more pop ups, the confirmation message happens directly under the users mouse and it keeps everything happening in the one location.

arcturus2:
@Atlas139mkm maybe we should get more opinions on this one.

In terms of the rest of the app there are uses of the `Alert` dialog and checking for a confirmation. I think it is ok for a popup to appear when we really want to get the users attention and it is fine for this scenario to do so.

You got me thinking further and if I think about design principles, I am not sure if showing the "are you sure" in a button is the best way to do it because it still looks like a button. I would instead use a `Label`. But this means more use of objects and setting things visible/invisible...

I totally understand your thinking about using less popups and reusing components which are great ideas.

What do you think @antares1470 @cygnus-x-1  @aldebaran30701 @Delphinus8821 ?

Atlas139mkm:
Yeah I was thinking that, I could also just temporarily remove the edit button, delete note button and replace it with a label saying are you sure? and the no and yes buttons. Takes a little bit more reconfiguring but it would probably look better

aldebaran30701:
I'm a fan - However I think you are right about the button for displaying text. I feel like so the user doesn't just miss the fact that the text changed that it should have a larger text. I'd say just needs to be a bit more noticeable. Many popups will get annoying especially if you have a few notes to clear out.
image

sol695510:
Notes Button layout suggestion
@arcturus2 I remember you pointing me here when I initially started creating the Notes View, I think this method of hiding/unhiding the buttons on hover should be considered for the Notes View, but I can understand it being moved to a new ticket.

Thoughts? @antares1470 @aldebaran30701 @Delphinus8821 @Atlas139mkm

antares1470:
@sol695510 I reckon it could be a cool addition, I think it could be worth a separate ticket for it though.

@Atlas139mkm @arcturus2 For me I think you want to force people to address the delete and that's why the pop up isn't a bad idea. It forces their eye to move to the dialog and they have to acknowledge it before they can do anything else. 

What I think we want to avoid is a user accidentally clicking the delete button and continuing on like nothing happened because they haven't noticed the buttons change. That scenario would result in the possibility of them accidentally pressing yes later on which I think leaves you not much better off than what we have currently (i.e. no confirmation check)

Atlas139mkm:
So you would prefer a confirmation pop up over implementing what @aldebaran30701 suggested?

antares1470:
> So you would prefer a confirmation pop up over implementing what @aldebaran30701 suggested?

Regarding pop ups, I think our objective is to avoid having any unnecessary ones, rather than avoiding them altogether. Sometimes, a pop up can actually be a good solution (particularly if confirming actions with notable consequences).

That being said I think the main things to be addressed are that firstly there is confirmation, and secondly that the confirmation needs to be addressed immediately. A pop up like @arcturus2 suggested is probably the easiest way to implement that but if @aldebaran30701's solution also addresses those two points then I'm not against it

Atlas139mkm:
it would satisfy the fact that there is confirmation, however it would not warranty any level of urgency. The user would not be forced to respond to it before doing anything else. They would not be able to do anything else with that specific note but they could leave the notes view and do anything else they liked. I'm also having sizing issues implementing it atm as well since a note dynamically resizes I can capture its current size so it looks a bit strange when you click delete note. I think a pop up might be the way to go if its the very next action you want the user to do.

Atlas139mkm:
Alright I have just committed some more changes. 

This is the note.
image

Confirmation message on delete. will only delete on user clicking OK.
image

Note in edit mode
image

User right click options to add and remove selected graph elements
image

Thoughts?

antares1470 (changes requested):
Looking good. I think the only thing to add now is a whatsnew entry

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,14 @@
...
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2021-09-08 Notes View UI Improvements
+<p>New features:
+<ul>
+<li><p>Replaced edit pop up with direct note editable text fields</p></li>
+<li><p>Right click note to add and remove selected graph elements</p></li>
+<li><p>Delete note confirmation pop up</p></li>
+</ul>
+
```
antares1470, line 11: The first `<p>` is missing a `</p>`. The list items themselves also don't need to be contained in `<p></p>`

antares1470 (changes requested):

antares1470 (approved):

arcturus2 (approved):
Works nicely

PR #1329, opened 2021-08-31
Fix to horizontally aligned parameter dialogs

arcturus2:
### Description of the Change

A quick fix to preventing buttons in the horizontally aligned parameter dialogs from being chopped off. Below is what the fixed version looks like and it is wider than I like it to be and I can investigate this further later on but for now I would like this to be accepted to help get v2.4 over the line.

image


GammaVel (approved):
Looks great to me - all the buttons are visible and the layout and styling has improved.

antares1470 (approved):
Approving this for 2.4 but as you say it would be nice to have the buttons closer together

PR #1331, opened 2021-09-01
Notes view notes carrying over from graph to graph

Atlas139mkm:
### Description of the Change

Separate PR for issue introduced by #1316.

- Graphs notes would carry over from previous graph to new graph.
- This was an unintended result of the notes disappearing bug
- This code should fix the issue of the notes disappearing and should not have notes transfer from one graph to the next

Notes view doesn't function properly without this fix.

### Benefits

Working notes view

### Possible Drawbacks

There shouldn't be any

### Verification Process

Tested the original bug of user created notes disappearing. (The note did not disappear) 
Tested creating a note then creating a new graph.  (The notes did not transfer over )

### Applicable Issues

#1316
#1325 (This issue was fixed here, however this separate PR has been made)


arcturus2 (approved):
Perfect, working much better thank you @Atlas139mkm

antares1470 (approved):

Issue #1308, opened 2021-08-27
Import from File - Move transactions to the middle


arcturus2:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

A simple yet effective improvement is moving the transactions attributes block to the middle so that it is a clearer representation of how a destination node hangs at the end of a transaction.

![image](https://user-images.githubusercontent.com/39325530/131076907-674dfc63-c13e-403d-8bbb-01ba81f55f83.png)

GammaVel:
The new layout is great. This makes more sense. Thanks @Atlas139mkm for making this change!!

PR #1345, opened 2021-09-03
Move transactions to middle

Atlas139mkm:
### Description of the Change

- Changed the order that the 3 columns appear on the file importer

image

As per the issue for it #1308 , it improves readability by showing a more appropriate order of how elements actually appear in the graph.

### Benefits

It looks nicer.

### Possible Drawbacks

long term users might get confused initially. 

### Verification Process

- I ensured that the + button associated with transactions was still associated with it.
- I did not change any functional code, only the order in which they appear.

### Applicable Issues

#1308


antares1470 (approved):

arcturus2 (approved):
Thank you @Atlas139mkm

Issue #1342, opened 2021-09-03
Notes View Concurrent Mod


arcturus2:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

```
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1542)
	at au.gov.asd.tac.constellation.views.notes.NotesViewPane.lambda$setNotes$8(NotesViewPane.java:432)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
[catch] at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to Reproduce

In a highly connected graph if I continue hopping out I am able to sometimes produce this error with the notes view open.

**Expected behaviour:** \[What you expect to happen\]
No Errors

**Actual behaviour:** \[What actually happens\]
Concurrent Mod Error

**Reproduces how often:** \[What percentage of the time does it reproduce?\]
40%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

GammaVel:
I'm not seeing this with Release 2.4, it is looking good. Thank you @arcturus2 :-)

PR #1346, opened 2021-09-03
Resolve Concurrent Mod Error #1342 and Notes View fixes

arcturus2:
### Description of the Change

I would like this and the related changes to go into v2.4.0

:bug: Fixed a bug with a concurrent mod issue
:art: Removed the experimental tag for Notes View
:art: Changed the UpdateDimOrHidePlugin to a "LOW LEVEL" tag
:art: Fixed a syntax error in what's new

### Alternate Designs

Every graph action triggers a plugin report which causes the notes view to read and write from the state. This is not efficient and needs to be reworked so that it can be as efficient as the plugin reporter or as close to as possible.

### Benefits

This resolves a concurrent modification issue

### Possible Drawbacks

My concern was around more memory use but I have been able to confirm that this does not increase the memory use. The spike in memory has to do with the notes being rendered which are JavaFX objects.

### Verification Process

I did a lot of subsequent actions like continuously hopping out (holding ctrl+ right arrow) on a highly connected graph with the notes view (and even the table view) open and I could no longer produce the issue.

### DIFF
--- a/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
+++ b/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
@@ -115,6 +115,12 @@ public class NotesViewController {
                 return;
             }
 
+            // TODO: A copy of the list is being passed here instead because
+            // the Notes View is reading and writing state all the time.
+            // Review why the notes view needs to be read and written to so
+            // often. If this can be reduced significantly then the chance of
+            // reading and writing happening at the same time is reduced and we
+            // can simply pass by reference again
             pane.setNotes(new ArrayList<>(currentState.getNotes()));
             pane.setFilters(currentState.getFilters());
         }
--- a/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
+++ b/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
@@ -29,6 +29,7 @@ import au.gov.asd.tac.constellation.plugins.templates.SimpleReadPlugin;
 import au.gov.asd.tac.constellation.views.notes.state.NotesViewConcept;
 import au.gov.asd.tac.constellation.views.notes.state.NotesViewEntry;
 import au.gov.asd.tac.constellation.views.notes.state.NotesViewState;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -88,7 +89,7 @@ public class NotesViewController {
     /**
      * Read the current state from the graph.
      */
-    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"EXPERIMENTAL", "LOW LEVEL"})
+    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"LOW LEVEL"})
     public static final class NotesViewStateReader extends SimpleReadPlugin {
 
         private final NotesViewPane pane;
@@ -114,7 +115,7 @@ public class NotesViewController {
                 return;
             }
 
-            pane.setNotes(currentState.getNotes());
+            pane.setNotes(new ArrayList<>(currentState.getNotes()));
             pane.setFilters(currentState.getFilters());
         }
 
@@ -127,7 +128,7 @@ public class NotesViewController {
     /**
      * Write the current state to the graph.
      */
-    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"EXPERIMENTAL", "LOW LEVEL"})
+    @PluginInfo(pluginType = PluginType.UPDATE, tags = {"LOW LEVEL"})
     private static final class NotesViewStateWriter extends SimpleEditPlugin {
 
         private final List<NotesViewEntry> notes;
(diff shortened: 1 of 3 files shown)


CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
```diff
@@ -114,7 +115,7 @@ public void read(final GraphReadMethods graph, final PluginInteraction interacti
                 return;
             }
 
-            pane.setNotes(currentState.getNotes());
+            pane.setNotes(new ArrayList<>(currentState.getNotes()));
```
sol695510, line 118: Instead of instantiating a new array with the values of currentState.getNotes(), would it possibly be better to have currentState.getNotes() return an unmodifiable list?

sol695510 (changes requested):
Looks good to merge, just wanted clarification on the comment I made.

CoreTimelineView/src/au/gov/asd/tac/constellation/views/timeline/clustering/ClusteringManager.java
```diff
@@ -402,7 +402,7 @@ protected void execute(final PluginGraphs graph, final PluginInteraction interac
         }
     }
 
-    @PluginInfo(tags = {"MODIFY"})
+    @PluginInfo(tags = {"LOW LEVEL"})
     public final class UpdateDimOrHidePlugin extends SimplePlugin {
```
antares1470, line 406: If you look at the plugin details closely, you'll see that it's actually updating the dimmed and visibility attributes of the graph elements and so I think 'MODIFY' is the correct tag here

antares1470 (changes requested):

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
antares1470: If going with the new ArrayList is the way to go, are you also able to add a comment explaining why we're creating a new list populated with an existing list

arcturus2: Good question @sol695510. If you make it an unmodifiable list you end up with a different exception (i.e `UnsupportedOperationException`) as the code elsewhere is still trying to write to the `List`. It helps confirm the issue of something reading and writing at the same time but does not rectify the root cause which is while the List is being read, something else is changing the list.

My understanding is `Collections.unmodifieble()` is used more when you want give the list to someone you don't want them to change. In this case, changing the `getNotes()` to `return Collections.unmodifiableList(notesViewEntries)` could be something we do as a protection layer and indicates to anyone trying to use `getNotes()` that they can only read from it.

I did a test with this and removed the `new ArrayList` and got the following error, to help confirm what I meant above, i.e it won't solve the root cause.

```
java.lang.UnsupportedOperationException
	at java.base/java.util.Collections$UnmodifiableCollection.clear(Collections.java:1079)
	at au.gov.asd.tac.constellation.views.notes.state.NotesViewState.setNotes(NotesViewState.java:65)
	at au.gov.asd.tac.constellation.views.notes.NotesViewController$NotesViewStateWriter.edit(NotesViewController.java:153)
	at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.describedEdit(SimpleEditPlugin.java:207)
[catch] at au.gov.asd.tac.constellation.plugins.templates.SimpleEditPlugin.run(SimpleEditPlugin.java:115)
```

I could be wrong but the underlying issue is that there are many reads and writes in quick succession which is happening on any graph change (because it triggers a plugin which then causes the state to update) and if that could be fixed I feel the concurrent mod would be a rare thing to occur. We could then potentially do away with the `new ArrayList` if its unlikely to be reproduced (but it is still an issue).

Another solution to all of this is to use `synchronized` and lock the reads and writes but that would slow the entire app down because everything would have to wait for the read and write states which is happening all the time.

Does that help clarify things?

CoreTimelineView/src/au/gov/asd/tac/constellation/views/timeline/clustering/ClusteringManager.java
arcturus2: True but whats happening here is that when the timeline view is open, it is flooding the auto notes with this plugin call whenever a selection is made which isn't useful. I can put a comment explaining why this one is a low level if you like?

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewController.java
arcturus2: @antares1470 I've updated this with a comment as suggested.

CoreTimelineView/src/au/gov/asd/tac/constellation/views/timeline/clustering/ClusteringManager.java
arcturus2: @antares1470 I've changed this to include both tags and I have also added an explanation as to why the LOW LEVEL tag was added. This change will still prevent the notes view from showing the plugin call but we can see it in the Plugin Reporter if we need to.

sol695510 (approved):

antares1470 (approved):

Issue #1307, opened 2021-08-27
Import from file - Enable sections one at a time


arcturus2:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

To help improve UX can we investigate disabling all sections of the file importer on load like the "Configruation" and "Attributes" section given none of them can be used until the "Source and Destination" section is filled out? It should help guide new users how to do an import without reading any documentation.

As a quick example see how much clearer it is when only the top section is enabled?

![image](https://user-images.githubusercontent.com/39325530/131076407-fec6f910-50c8-4cef-afb6-0ef85f66baa5.png)

Atlas139mkm:
Looking into this, Thoughts on not disabling it but having it start collapsed. Once a file is present in the files list then it would expand for the user. This would avoid annoyances for any developing of the import pane if users had to import every time they wanted to examine the view.

GammaVel:
I love the changes that have been made and it makes this functionality simpler to use.  Thanks @Atlas139mkm!

PR #1355, opened 2021-09-06
Hide import sections when no file present

Atlas139mkm:
### Description of the Change

In the import view I have collapsed the configuration section and disabled the import button by default. These will be expanded and enabled respectively when there is a file present within the imported files list.

- The user can choose to expand the configuration section without importing a (nice for developers)
    - however the import button will stay disabled while no file is present
-  When the user imports a file after manually expanding the configuration section, it will stay open and the import button will enable
- a demonstration of the changes is shown in the gif below

disabledImportView

### Alternate Designs

- we could completely disable the configuration section if requested (let me know and ill see what I can do)

It assists the user in the desired flow of the view. Guiding them to import a file first before attempting to assign attributes and import it to the graph.

### Benefits

UX enhancements

### Possible Drawbacks

the final import button is disabled without first importing a file. (there might be a use case where this is wanted but likely not)

### Verification Process

- ensured the config pane expanded on file import
- ensured the import button enabled on file import
- ensured they both stayed active after importing 2 files
- ensured the config pane remained expanded after manually expanding it then importing a file
- ensured both config pane and import button disabled when no more files were in the import list

### Applicable Issues

#1307


arcturus2 (changes requested):
A really nice implementation @Atlas139mkm and it works well. Can you please update `whatsnew.txt` to advertise this new feature please?

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ActionPane.java
```diff
@@ -77,4 +79,9 @@ private void displayAlert(final String title, final String header, final boolean
         dialog.setHeaderText(header);
         dialog.showAndWait();
     }
+
+    // sets the enabled state of the import button
+    public void enableButton(boolean isEnabled) {
```
antares1470, line 84: parameter final

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
```diff
@@ -200,4 +203,12 @@ public SourcePane getSourcePane() {
     public Window getParentWindow() {
         return this.getScene().getWindow();
     }
+
+    public void expandPane(boolean isExpanded) {
+        titledConfigurationPane.setExpanded(isExpanded);
+    }
+
+    public void enableButton(boolean isEnabled) {
```
antares1470, line 211: parameters final

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/DelimitedImportController.java
```diff
@@ -219,4 +222,16 @@ public void setImportFileParser(final ImportFileParser importFileParser) {
...
+
+    // expands or shrinks the import pane based on if there is a file present
+    // within the imported files list
+    void openConfigPane(boolean b) {
+        importPane.expandPane(b);
+    }
+
+    // enables or disables the import button based on if there is a file present
+    // within the imported files list
+    void enableButton(boolean b) {
```
antares1470, line 234: parameters final

antares1470 (changes requested):

arcturus2 (approved):
@Atlas139mkm some merge conflicts have come up so please take a look but once @antares1470's feedback is addressed I am happy for this to be merged in.

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,9 @@
 == 3030-12-31 Getting Started
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2021-09-06 Improved import view UI guidance of intended flow
```
antares1470, line 4: I think I would simplify the title here to something like 'Import From File'. Keeps the title simple and then the details can be put in the description

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,9 @@
...
+<p>In the import view, the configuration pane starts shrunk and the import button starts disabled. They will expand/be enabled respectively once a file is present the files list. The opposite will occur when the the list is empty.</p>
```
antares1470, line 5: Import from File rather than import view. I think I would also keep the description more concise. Something like "Updated the UI to only enable and expand sections once a file has added"

antares1470 (changes requested):

Atlas139mkm:
Alright I have made the requested changes. I going to work on enabling the import button on the database connection tomorrow morning. I don't want to cause issues for those that connect via database.

Atlas139mkm:
The import button should enable and the config pane should expand when data is present within the table. This should work when a query has been successful and was able to populate the table with data.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
```diff
@@ -193,7 +193,20 @@ protected void updateSampleData() {
 
         if (configurationPane != null) {
             configurationPane.setSampleData(currentColumns, currentData);
+            openConfigPane(!currentColumns[0].isBlank());
+            enableButton(currentColumns[0].isBlank());
```
antares1470, line 197: Looks like further up the function there is the possibility of `currentColumns` being length 0, in which case this will throw an `ArrayOutOfBoundsException` so you'll probably want to add in a boolean condition for that. I would also switch to using `StringUtils.isBlank()` and `StringUtils.isNotBlank()` is case the value of `currentColumns[0]` is null

antares1470 (changes requested):

Atlas139mkm:
I renamed enable button to disable button to be more inline with what it actually does. I also changed the function to check for the current data not just the columns. I also added a else statement that just would shrink and disable the elements if there is ever a case where it could occur.

arcturus2 (approved):
Tested it and it works well.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
```diff
@@ -193,20 +193,26 @@ protected void updateSampleData() {
...
+
+            // If the current columns lenght isn't 0, expand the config pane
+            // and enable the import button
+            if (!currentData.isEmpty()) {
+                openConfigPane(StringUtils.isNotBlank(currentData.get(0).toString()));
+                disableButton(StringUtils.isBlank(currentData.get(0).toString()));
+            } else {
+                openConfigPane(false);
+                disableButton(true);
+            }
```
antares1470, line 205: Given `currentData` is a `List<String[]>`, the `toString()` of an array won't ever be blank so effectively the code could be simplified to 
```
openConfigPane(!currentData.isEmpty())
disableButton(currentData.isEmpty())
```

antares1470 (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/jdbc/JDBCImportController.java
antares1470: Should probably also change the comment to refer to `currentData` rather than `currentColumns`

antares1470 (approved):

Issue #1312, opened 2021-08-27
Import from File - Reset graph view after import


arcturus2:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

After a file is imported the graph view does not reset to the full view which should be the default behaviour as you want to see everything that you just imported.

PR #1357, opened 2021-09-06
Import from File and Import from Database - Reset graph view after importing

Auriga2:
### Description of the Change
When imported by any of above importers, the graph view is now reset to the full view to display everything that's been imported.

Better UX

### Benefits
Better UX

### Verification Process
Import a big set of data using both of the above importers and verify it resets to the full view to display everything that's been imported.

image

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1312

### DIFF
--- a/CoreImportExportPlugins/nbproject/project.xml
+++ b/CoreImportExportPlugins/nbproject/project.xml
@@ -31,6 +31,14 @@
                     </run-dependency>
                 </dependency>
                 <dependency>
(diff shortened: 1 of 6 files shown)


arcturus2 (approved):

arcturus2 (changes requested):
Thank you @Auriga2 and this works really well. Can you please update `whatsnew.txt` to explain this feature improvement please?

antares1470 (approved):
Good to go one @arcturus2's feedback has been addressed

arcturus2 (approved):

Issue #1310, opened 2021-08-27
Import from File - Make menu options more obvious


arcturus2:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The "Options" menu and what it has is not always clear to users and given there is not many options there, they could just be buttons on the GUI.

![image](https://user-images.githubusercontent.com/39325530/131077510-58945fda-5393-4c5a-9758-28a5827d756f.png)

GammaVel:
Thanks @Atlas139mkm for making these new menu items more obvious. Nice work :-)

PR #1358, opened 2021-09-06
increased visibility of import views options

Atlas139mkm:
I would like feedback on this one regarding code from 118 - 135. It appeared to not provide any functionality after deleting as it was tied to the options menu drop down. However I'm not entirely sure if has a lingering effect somewhere else.

### Description of the Change

Changed the visibility of the options menu items to be more obvious by removing the options menu and placing options menu items straight on the view. (shown below)

All functionality should remain the same, the menu options have just been converted to buttons.

image

Its easier for the user to see the views functionality.

### Benefits

Easier for users.

### Verification Process

- ensured SAVE and LOAD buttons were still triggering the correct pop up
- ensured checking the show attributes check box still triggered showing more attribute options

### Applicable Issues

#1310

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
@@ -19,15 +19,12 @@ import au.gov.asd.tac.constellation.utilities.color.ConstellationColor;
 import au.gov.asd.tac.constellation.utilities.icon.UserInterfaceIconProvider;
 import java.util.List;
 import java.util.prefs.Preferences;
-import javafx.event.ActionEvent;
 import javafx.geometry.HPos;
 import javafx.geometry.Insets;
 import javafx.geometry.VPos;
 import javafx.scene.control.Button;
 import javafx.scene.control.CheckBox;
-import javafx.scene.control.Menu;
-import javafx.scene.control.MenuBar;
-import javafx.scene.control.MenuItem;
+import javafx.scene.control.Label;
 import javafx.scene.control.TitledPane;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
@@ -56,11 +53,9 @@ public class ImportPane extends BorderPane {
             ConstellationColor.BLUEBERRY.getJavaColor());
 
     protected final Preferences importExportPrefs = NbPreferences.forModule(ImportExportPreferenceKeys.class);
-    protected final Menu optionsMenu;
-    protected final MenuItem loadMenuItem;
-    protected final MenuItem saveMenuItem;
     protected final CheckBox showSchemaAttributesCheckBox;
-    protected final MenuItem showSchemaAttributesItem;
+    protected final Button loadButton;
+    protected final Button saveButton;
     protected final Button helpButton;
 
     protected ImportController importController;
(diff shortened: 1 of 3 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
```diff
@@ -115,34 +106,16 @@ public ImportPane(final ImportTopComponent importTopComponent, final ImportContr
...
-                        ImportExportPreferenceKeys.DEFAULT_SHOW_SCHEMA_ATTRIBUTES);
-                importExportPrefs.putBoolean(ImportExportPreferenceKeys.SHOW_SCHEMA_ATTRIBUTES, newPreference);
-                importController.setShowAllSchemaAttributes(newPreference);
-                importController.setClearManuallyAdded(false);
-                importController.setDestination(sourcePane.getDestination());
-                final int saveResultsItemIndex = optionsMenu.getItems().indexOf(showSchemaAttributesItem);
-                ((CheckBox) optionsMenu.getItems().get(saveResultsItemIndex).getGraphic()).setSelected(newPreference);
-            }
-            userClickedTheCheckboxFirst[0] = false;
-        });
```
antares1470, line 134: It looks like this code is what's supposed to happen when the 'Show all schema attributes' item is (un)ticked so it probably needs to be retained in some way

antares1470 (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
Atlas139mkm: I think its for the item variant of showSchemaAttributes. so showSchemaAttributesItem not showSchemaAttributesCheckBox. The item variant was only needed if it was in the options menu. I was thinking it was for if the user clicked on the show schema attribute label in the menu instead of the check box. I think it was just a safety net so the check box would get ticked even if they didn't specifically tick it. I can put all the code I deleted back in if you think it could be doing something else however?

arcturus2: @Atlas139mkm your right, it was if they clicked the label and I added that bit of code so I remember it. OK for it to be deleted now that its a button.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/ImportPane.java
```diff
@@ -82,12 +77,8 @@ public ImportPane(final ImportTopComponent importTopComponent, final ImportContr
...
         titledSourcePane.setCollapsible(true);
 
-        // Options menu
-        optionsMenu = new Menu("Options");
-        loadMenuItem = new MenuItem("Load...");
-
-        // save menu item
-        saveMenuItem = new MenuItem("Save...");
+        loadButton = new Button("Load");
+        saveButton = new Button("Save");
```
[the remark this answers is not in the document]
arcturus2: Agree @antares1470  but the non-plural form so "Load Template".

antares1470 (approved):

arcturus2 (approved):

Issue #1255, opened 2021-08-20
Import from File - Attribute Section unable to be reopened once collapsed on laptop screen


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

I like to work from a laptop as it's very convenient. The "Source and Destination" and the "Attribute Filter" on the Import from file both have an option to collapse to make the section smaller. If I collapse the "Source and Destination" I can click on the twisty icon to expand it again. If I collapse the Attribute filter there doesn't appear to be a way of expanding it again from my laptop screen. I have to close and reopen Constellation.  It would be great if this could be fixed. 

### Steps to Reproduce

1.  Go to File Import

2. Collapse the "Source and Destination" > it can be expanded again

3. Collapse the Attributes Section > there is no option to expand it. The only work around I have found is to close the File Import, close and reopen constellation. 
![image](https://user-images.githubusercontent.com/83679004/130172733-4b4faaa9-11d3-419c-83f8-22ad6d890feb.png)

**Expected behaviour:**The Attribute section can be expanded once collapsed

**Actual behaviour:** Unable to reopen it

**Reproduces how often:** 100%

Atlas139mkm:
Has this been fixed? I'm a bit confused.
For me I can expand/shrink:  source/destination and Attributes
     - in PR #1355 I have added the ability to expand/shrink the configuration section
     
I don't have a way to collapse attribute filter in the first place. I thought it was just intended to be a label and text box not a collapsible pane.

I did notice if the constellation window is too small you cannot see the attribute drop down as the screen isn't big enough to display it all.

GammaVel:
[...]

PR #1371, opened 2021-09-08
Attribute bar disappearing on collapse

Atlas139mkm:
### Description of the Change

- Added a scroll bar to the content within the run tabs. 
- Changed attribute scroll bar to be as needed instead of always.
- Changed height of sample view to be smaller to remove large blank space

This change was needed for users with small screens/monitors that would experience the attribute collapsible pane disappearing once collapsed. The new change adds a scroll bar that stops the collapsible pane from completely disappearing after being collapsed. The user will be able to see the bar after being collapsed by scrolling to the bottom of the Import From File screen.

image

image

Allows users with small monitors to use the import file feature with more ease.
Small monitor uses can also now properly see the attributes when the tab is expanded.

### Benefits

UX benefits.

### Possible Drawbacks

The sample height section is slightly smaller so users can not see as many rows as data as they originally could. (I'd like feedback on this one, but my logic was it removed a lot of blank space it was displaying after being placed in a scroll pane - this meant the user had to scroll a lot to get to the attributes. I also figured the important aspect of the table data was the column headings not the data itself.)

### Verification Process

- I ensured the views interactive aspects still worked as intended.
- I ensured the attribute bar was still present after being minimized. 

### Applicable Issues

#1255

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RunPane.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RunPane.java
@@ -121,7 +121,7 @@ public final class RunPane extends BorderPane implements KeyListener {
         public AttributeBox(final String label, final AttributeList attributeList) {
             final BorderPane borderPane = new BorderPane();
             borderPane.setTop(attributeList);
-
+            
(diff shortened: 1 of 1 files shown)


antares1470 (approved):
I was seeing the issue before and can confirm this fix works

arcturus2 (approved):

PR #1372, opened 2021-09-08
:bug: Fix setRotationMatrix to set the right values for a 3x3 matrix, and increase test coverage

groombridge34a:
### Description of the Change

`setRotationMatrix` is intended to set the top-left nine values of the 4x4 matrix to the values of the input 3x3 matrix. Incorrect values were used in the `System.arraycopy` calls to retrieve values from the source 3x3 matrix. This is especially obvious with the 3rd copy, which attempts to copy 4 values from a 9-element array, starting at the 8th element.

While I was here I also improved unit test coverage of the class.

### Alternate Designs

None - simple bugfix.

Reliance on this incorrect method could result in strange bugs when users move their viewpoint around Constellation views.

### Benefits

Improved correctness in a core utility method.

### Possible Drawbacks

Any plugins relying on this behaviour will no longer work as expected. Updated the change log to account for this.

### Verification Process

Created unit tests and ensured they run correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 -   Removed unused `localized` parameter from the signature of the `locate()`
     method in `ConstellationInstalledFileLocator`.
 
+-   Fixed `setRotationMatrix` in `Matrix44d` as it was previously placing 
+    rotation values in the wrong value positions.
+
 ## Changes in August 2021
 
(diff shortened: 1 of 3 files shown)


antares1470 (approved):

arcturus2 (approved):

arcturus2:
Thanks for fixing the bug @groombridge34a as well.

Issue #1143, opened 2021-07-20
Delimited File importer - doesn't accept full stops in the column headings


GammaVel:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The filter search option under the Run section on the Import from File Delimiter does not accept full stops in the column heading names.  This is a relatively minor issue however the training material for new starters located [here] has a sample file with a full stop in the column name(https://github.com/constellation-app/constellation-training/blob/master/Analyst%20Training/Exercise%201%20-%20Why%20Graphs/Greg%20and%20Sarah%20lunch%20network%20table.csv)

### Steps to Reproduce

1.  Download sample file for testing from the link above

2.  Click on Import from Delimited file

3.  Load specified file and confirm the file headers have a full stop in them such as source.Name

4. Click in the search field above the attributes and enter search query,  source.Name=="Greg" -> no results are shown

5. Create a copy of the file and remove the full stop. Repeat the same query search -> results are shown

**Expected behaviour:** results are returned for ile headers have a full stop in them 

**Actual behaviour:** as per repo steps - no results are returned

**Reproduces how often:**  100%

Auriga2:
The code would fail anything other than ("^[a-zA-Z][a-zA-Z0-9_]*$") in the column headers. I guess it doesn't need to validate while filtering. If there's any validation required, that should be handled before/when importing.

Auriga2:
Looks like the javax.script CompiledScript doesn't support those special characters, as it throws `ScriptException NameError: name 'Firstpart' is not defined in <script> at line number 1` for anything in the filter with `Firstpart` followed by a special character. Tried with escape characters \"Firstpart secondpart\" but get the same error although now it doesn't mark the script in red.

GammaVel:
Thank you @Auriga2 for working on this issue which will help searching of imported data. 
[...]

PR #1378, opened 2021-09-09
File and Database Importers- Fix the Configuration filter

Auriga2:
### Description of the Change
The table filter in the Configuration section now correctly filters when the filter text has column headers containing special characters like #,@,$, space etc other than _.
This filter was sometimes not filtering properly in a new Run tab. That has been fixed and now works consistently.

### Alternate Designs
A new search filter without using javax.script CompiledScript with python ScriptEngine.

Better UX

### Benefits
Better UX

### Verification Process

- Create a CSV file or edit an existing file to contain column headers with special characters  #,@,$, space except _. (source.Name, sourceTag#, Source Type etc.)
- Click on Import from Delimited file and load that file and confirm the file headers have the special characters in them.

- Click in the filter above the table and enter search query, such as source.Name=="Greg" and verify it searches correctly.

- Repeat with the database importer, adding special characters in the column headers.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1143

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RowFilter.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RowFilter.java
@@ -53,6 +53,7 @@ public class RowFilter {
 
     private String[] columns = new String[0];
     private String[] encodedColumns = new String[0];
+    private static final String columnHeaderPrefix = "uniqueColumnHeaderPrefix_";
 
     private static final Logger LOGGER = Logger.getLogger(RowFilter.class.getName());
 
@@ -82,7 +83,7 @@ public class RowFilter {
      */
     public boolean setScript(final String script) {
         try {
-            this.script = encodeScript(script);
+            this.script = encodeScript(modifyScript(script));
             compiledScript = ((Compilable) engine).compile(this.script);
 
             LOGGER.log(Level.INFO, "SCRIPT = {0}", this.script);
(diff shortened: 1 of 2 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RowFilter.java
```diff
@@ -93,9 +94,25 @@ public boolean setScript(final String script) {
             return false;
         }
     }
+    private String modifyScript(final String script) {
+        final int endOfHeader = script.indexOf("==");
+        String modifiedScript = script;
+        if (endOfHeader != -1) {
+            final String header = script.substring(0, endOfHeader);
+            if (header != null && !header.matches("^[a-zA-Z][a-zA-Z0-9_]*$")) {
+                //Append the script with the relevant ith column header containing columnHeaderPrefix
```
arcturus2, line 103: Did you mean to say "Amend" instead?

arcturus2 (changes requested):
Thanks @Auriga2 it seems to work well. Can you please have a look at the code smells and resolve at least the critical and major smell?

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RunPane.java
```diff
@@ -152,9 +154,17 @@ public RunPane(final ImportController importController, final String displayText
...
-                rowFilter = new RowFilter();
-            }, ROW_FILTER_INITIALISER).start();
+            try {
+            final CountDownLatch latch = new CountDownLatch(1);
+                new Thread(() -> {
+                    rowFilter = new RowFilter();
+                    latch.countDown();
+                }, ROW_FILTER_INITIALISER).start();
+                latch.await();
+            } catch (InterruptedException ex) {
```
antares1470, line 164: final

antares1470 (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RowFilter.java
antares1470: I don't think append is necessarily the wrong choice of word here but "Edit" may be the better choice (I usually associate append with adding to the beginning or end rather than somewhere in the middle)

Auriga2:
I can't find a way to fix this warning. Is this really an issue? 
image

antares1470:
> I can't find a way to fix this warning. Is this really an issue?

@Auriga2 you can probably ignore that one for now

Auriga2:
`replaceFirst` doesn't seem to work on a string with $ in it, although it worked with other characters I tested.   `StringUtils.replaceOnce` seems to be a better match.

Auriga2:
Noticed another minor issue, now the filter text turns red when the column header doesn't exist, only if it also contains a special character. I'll fix that too give me few mins.

antares1470 (approved):

arcturus2 (approved):

Issue #1123, opened 2021-06-24
Make changing label colours more intuitive


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently, in order to change a the colour of node labels (similar for transaction labels) you need to do the following:
1. Open the Attribute Editor
2. Search for the graph attribute 'node_labels_bottom' (or 'node_labels_top') and press Edit
3. Click on the colour square for an attribute
4. Change to the desired colour
5. Hit Ok for each dialog box

The graph attribute (e.g. 'node_labels_bottom') only displays which attribute(s) are added as labels and so unless you already knew the above process, you would be unlikely to know that's what you need to do to change the label colour.

### Steps to Reproduce

1.  Follow above process

**Expected behaviour:** Process to change label colours should be intuitive to figure out

**Actual behaviour:** Not enough information is displayed in the right places to figure out what needs to be done

**Reproduces how often:** 100%

### Additional Information

The challenge here will be figuring out what exactly what is enough to make the process clear. I suspect maybe include renaming the graph attribute (e.g. ''node_bottom_label_attributes") or maybe displaying the value as both the attribute and the colour (e.g. "Label (Blue)" or "Label (#0000ff)" rather than simply "Label") could be options.

PR #1381, opened 2021-09-10
changed label name and added colour to display text

Atlas139mkm:
### Description of the Change

I have changed:

node_labels_bottom -> node_bottom_labels_colours

node_labels_top -> node_top_labels_colours

These attributes now also display the colour name of what they are set to. Example below.

image

### Alternate Designs

I had ideas of changing the text colour of each label. Seemed alright but would cause issues with readability in some cases.

I had the idea of adding a colour rectangle in front of every label. Proved very hard due to the current code structure of the attribute editor. Also would have required a lot of effort for a very minor issue.

Improved readability. Improved explanation of what the attribute does, making it more intuitive for the user.

### Possible Drawbacks

the label now has a different name? - I would imagine it wont cause issues but maybe?

### Verification Process

Ensured the buttons still maintained the same functionality.

### Applicable Issues

#1123


antares1470 (changes requested):
A couple of things:
1) Could you add a whatsnew entry to indicate the change in attribute names
2) In order for this to work with existing graphs, you'll need to create a visual schema updater

Atlas139mkm:
The code I wrote for the latest commit was translated from the visual schema updater V4. So I'm not sure if its perfect however it appears to work. Feedback is greatly appreciated.

CoreVisualSchema/src/au/gov/asd/tac/constellation/graph/schema/visual/compatibility/VisualSchemaV6UpdateProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    protected void schemaUpdate(StoreGraph graph) {
+
+        // retrieve the attribute Id of both the graph_labels_top/bottom
+        final int oldGraphBottomlabelsAttributeId = graph.getAttribute(GraphElementType.GRAPH, GRAPH_BOTTOM_LABELS);
+        final int oldGraphToplabelsAttributeId = graph.getAttribute(GraphElementType.GRAPH, GRAPH_TOP_LABELS);
+
+        // retrieve the new values for the graph_labels_bottom and set it to the
+        // new values
+        final int newGraphBottomLabelsAttributeId = VisualConcept.GraphAttribute.BOTTOM_LABELS.ensure(graph);
+        final String bottomLabelValue = graph.getStringValue(oldGraphBottomlabelsAttributeId, DEFAULT_VERSION);
```
antares1470, line 69: I'm assuming the use of `DEFAULT_VERSION` is because sonar said to use the existing constant. In this case though I don't think it makes sense to use it as this is different context (vertex id rather than schema version). If you could change this back to 0, that would be great (we can mark the sonar rules as won't fix)

CoreVisualSchema/src/au/gov/asd/tac/constellation/graph/schema/visual/compatibility/VisualSchemaV6UpdateProvider.java
```diff
@@ -0,0 +1,84 @@
...
+        graph.setStringValue(newGraphBottomLabelsAttributeId, DEFAULT_VERSION, bottomLabelValue);
```
antares1470, line 70: similar here

CoreVisualSchema/src/au/gov/asd/tac/constellation/graph/schema/visual/compatibility/VisualSchemaV6UpdateProvider.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // retrieve the new values for the graph_labels_top and set it to the
+        // new values
+        final int newGraphTopLabelsAttributeId = VisualConcept.GraphAttribute.TOP_LABELS.ensure(graph);
+        final String topLabelValue = graph.getStringValue(oldGraphToplabelsAttributeId, DEFAULT_VERSION);
+        graph.setStringValue(newGraphTopLabelsAttributeId, DEFAULT_VERSION, topLabelValue);
```
antares1470, line 76: and these 2 lines as well

antares1470 (changes requested):
If you could add a whatsnew entry as well, that would be great

antares1470 (changes requested):
If you could add a whatsnew entry that would be great

Atlas139mkm:
yep apologies, just did.

antares1470 (approved):

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,10 @@
 == 3030-12-31 Getting Started
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2021-09-13 node_labels_bottom/top changed
+<p>The above two graph attributes have been changed to node_bottom/top_labels_colours.</p>
+<p>The attributes now also show the colours for example: Label: (White), Identifier (#9980e6)</p>
```
arcturus2, line 6: missing a `:` after the Identifier

arcturus2 (changes requested):
This works great. So doing a schema upgrade is a big deal so can you please do the transaction labels as well for consistency? Sorry this is scope creep but because we need to do a schema update i'd like to include it in now.

Is that OK?

Atlas139mkm:
@arcturus2 

While writing the code for the transaction labels I noticed. That feature does not work as intended at all. 
- No more that 1 attribute will be shown on the transaction
- it will always be the count 
- if the count is removed it can not be re-added
- the colour of the count is determined by the first element in the list not the colour of the count attribute

I can create a separate issue or can attempt to solve it in this one if I would need to do another visual schema updater?

antares1470 (approved):

arcturus2 (approved):
@Atlas139mkm perfect thank you.

Issue #1323, opened 2021-08-31
Import from DB (JDBC) - Add some helper text and default the driver on the Add Connection screen


GammaVel:
### Prerequisites

- [X ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When importing data from a database I first need to add a driver and a connection via the Manage Connection Screen. It would be great to add some helper text to this screen that would disappear once a user typed into it, to help people understand what they need to do on this screen. As a new user of this functionality I wasn't sure what to do.

Suggestions - add guidance for the connection name, add sample text for what a connection string should look like: eg - jdbc:mysql:<database_location>

It would also be really good to populate the drop down with the first driver that has been added on the driver screen. 

The import from DB is very powerful, thanks for building this. With a few extra UX updates it will be even better!!

### Steps to Reproduce

1.  In constellation click on Import from Database
2.  Click on Manage Connection > Go to the Connections Screen
3.  Review the information on screen - and try to use - I wasn't able to, I had to exit and look at the help

**Expected behaviour:**  Some helper text is available to make this functionality easier to understand

**Actual behaviour:** Didn't know how to use this screen

**Reproduces how often:** 100%

### Additional Information

![image](https://user-images.githubusercontent.com/83679004/131435771-aa92036e-82b4-40d0-bc7c-f29431b32d6f.png)

GammaVel:
I really like the new helper text, thanks for adding it @Atlas139mkm !  One slight update requested, can we please change the connection string section to be more similar to the other helper text: 
![image](https://user-images.githubusercontent.com/83679004/133969275-8cfc9441-4905-424b-a12c-5cd6b4d35021.png)

Otherwise I initially thought it had been set for me! The rest of the are great

GammaVel:
Thank you very much @Atlas139mkm  for taking into account my feedback so quickly and updating the text, I tried it out now and it worked well. The first time I thought I could just save the helper text, now it is much clearer I need to enter my own. Thanks!!

PR #1387, opened 2021-09-10
Add prompt text to jdbc add connection

Atlas139mkm:
### Description of the Change

Added prompt text to the jbdc screen.
Automatically selected the first driver in the list if there was one.

image

### Why Should This Be In Core?'

Allows users to better understand what should be in each field.

### Benefits

helpful for users

### Possible Drawbacks

none

### Applicable Issues

#1323


antares1470 (approved):

arcturus2 (approved):
Thank you @Atlas139mkm

Issue #1239, opened 2021-08-18
Add Shortcut to Data Access View on Welcome Page and toolbar


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Given the Data Access View is one the ways to get data onto a graph, it makes sense to have a shortcut for it on the Welcome Page (like the File and Database Importer) as well as the top toolbar (like Histogram and Timeline)

**Expected behaviour:** A shortcut to the Data Access View from both the Welcome Page and the top toolbar

**Actual behaviour:** No such shortcuts exist

**Reproduces how often:** 100%

GammaVel:
Thanks for adding this handy shortcut to the toolbar and welcome page.  I confirmed both of these are working and launching the view. I'm not a huge fan of the eye icon but understand why it was picked.

PR #1389, opened 2021-09-13
added data access view to welcome page and toolbar

Atlas139mkm:
### Description of the Change

I have added shortcuts for the data access view to the toolbar as well as the welcome page (shown in the image below)

image

### Alternate Designs

I would like thoughts on if the welcome page buttons should wrap around? So is row is at maximum at 6 buttons?, or if we should leave it as 7 (how it is in the image above).

So would we want.
0 0 0 0 0 0
0

or 
0 0 0 0 0 0 0

Requested for easier access to the view as it is a common way of importing/extracting data for users.

### Benefits

Data access view is easier to access.

### Possible Drawbacks

Maybe the level of neatness having an uneven number of short cuts on the welcome page?

### Verification Process

All buttons work as intended. Steps were directly followed from previous examples.

### Applicable Issues

#1239


antares1470:
@Atlas139mkm At this point my opinion is keeping them all on the one row. I think long term you may want to spread it over a couple of rows but for me that only makes sense if there are more links. Having 6 in one row and 1 in the next looks a little funny for my mind and realistically I wouldn't expect to see many more links anyway given the Welcome Page isn't really designed to have a link to absolutely everything in the application (would kind of defeat the purpose of the Welcome Page)

CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/plugins/DataAccessViewWelcomePlugin.java
```diff
@@ -0,0 +1,102 @@
...
+import org.openide.util.NbBundle;
+import org.openide.windows.TopComponent;
+import org.openide.windows.WindowManager;
+
+/**
+ * The Open Data Access View for the Welcome Page.
+ *
+ * @author Atlas139mkm
+ */
+@PluginInfo(pluginType = PluginType.CREATE, tags = {"WELCOME"})
```
antares1470, line 39: I don't think CREATE is the right plugin type here. The other welcome plugins that open views don't seem to have types yet so you could potentially just omit. NAVIGATION potentially looks like the best option if you want to keep it though

CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/plugins/DataAccessViewWelcomePlugin.java
```diff
@@ -0,0 +1,102 @@
...
+@NbBundle.Messages("DataAcessViewWelcomePluging=Data Acess View Welcome Plugin")
+public class DataAccessViewWelcomePlugin implements WelcomePluginInterface {
+
+    public static final String IMPORT = "resources/welcome_data_access.png";
+    private final ImageView importImage = new ImageView(new Image(WelcomeTopComponent.class.getResourceAsStream(IMPORT)));
+    private final Button importButton = new Button();
```
antares1470, line 45: Given the label being used later is 'Open' rather than 'Import', I wonder if it would make more sense to change these fields to `OPEN`, `openImage`, and `openButton`

antares1470 (changes requested):

CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/plugins/DataAccessViewWelcomePlugin.java
Atlas139mkm: yep sounds good, apologies I was just mimicking one of the other files.

antares1470 (approved):

arcturus2:
Agree with @antares1470.

I am also going to propose we change the eye to a database icon - the eye is too creepy for me! I can do that though in a different PR later on.

I also noticed that the unit tests failed. I ran your branch locally and it ran ok so not sure why its failed but I will approve this.

arcturus2 (approved):

arcturus2:
I added @Delphinus8821 as an approver but I see you have already put a thumbs up so I will go ahead and merge this in.

Issue #1244, opened 2021-08-18
Update name of Export > To Image...


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When you go File > Export > To Image..., Constellation takes a screenshot of the current view of the graph window and saves it as a PNG. The idea of exporting to an image though may actually suggest that a photo of the whole graph is taken and so the name may be misleading. To remove the confusion, I propose renaming it from 'To Image...' to either 'Screenshot...' or 'To Screenshot Image...' as this better gives an idea of what is actually being exported. 

### Steps to Reproduce

1.  Go File > Export

**Expected behaviour:** Image option makes clear that screenshot of current graph window is being taken

**Actual behaviour:** Image option may suggest that screenshot of entire graph (whether or not whole graph is currently in frame) is being taken

**Reproduces how often:** 60% (depends of a user's interpretation)

### Additional Information

'Screenshot...' I think may be the clearer option of the proposed 2 but I note that it does go against the convention of 'To <something>...' so am fine with either really (or if someone else comes up with an even better one that could also work)

GammaVel:
Confirmed the text update has been made and the functionality is still working. This makes it clearer to me that I'm not getting a photo export of the graph. I'm glad the help guide was also updated.

PR #1391, opened 2021-09-13
Changed To image  ->  To screenshot image

Atlas139mkm:
### Description of the Change

Just changed the naming to be more self explanatory of what the action does.

image

Better explanation of what it does.

### Benefits

UI UX

### Verification Process

tested to see if the button still works - It does

### Applicable Issues
#1244


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/docs/export-from-constellation.html
```diff
@@ -23,7 +23,7 @@ <h1>Export From Constellation</h1>
 
                 </td>
                 <td>
-                    File -&gt; Export -&gt; To Image...
+                    File -&gt; Export -&gt; To Screenshot Image...
                 </td>
                 <td style="text-align: center">
                     <img src="../image/exportToImage.png" alt="Export to Image Icon">
```
antares1470, line 29: Are you able to change the alt here to "Export to Screenshot Image Icon"

antares1470 (changes requested):
There's also a reference to Image further down the help page. If you're able to change that one to Screenshot Image as well, that would be great

arcturus2 (approved):

antares1470 (changes requested):
There's also a reference to Image further down the help page. If you're able to change that one to Screenshot Image as well, that would be great

antares1470 (approved):

Issue #1240, opened 2021-08-18
Make templates more obvious in Data Access View


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

One of the lesser known features of the Data Access View is creating templates. The way templates work is that you can create a workflow using steps and plugins and can save that as a template so you don't need to recreate that workflow everytime you load up Constellation. This is a cool feature, but currently it's hidden in an options menu drop-down with "Load" and "Save" not exactly being clear as to what is loading or saving.

What I propose is 2 things:
1) Move this out of the Options menu so that users can see this option from the get go
2) Rename Load/Save to something like Load/Save Template or Load/Save Workflow

### Steps to Reproduce

1. Open up Data Access View
2. Setup a workflow
3. Try and save it

**Expected behaviour:** Option to load/save template can be accessed without digging through any drop-downs

**Actual behaviour:** Option to load/save template is accessed via Options menu

**Reproduces how often:** \[What percentage of the time does it reproduce?\]

### Additional Information

The help pages for Data Access View will probably need to be updated with this change

GammaVel:
Thank you @Atlas139mkm for making these improvements to the Data Access View options. I think it is much clearer now what these are for. I noticed my results are not saving as I would expect however I don't think this is related as I checked the previous version and it is also happening there. I'll do some more digging and raise a new issue if required.

Issue #1241, opened 2021-08-18
Add Quality Button and Go Button to bottom of Data Access View


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The current workflow in Data Access View is select all the plugins you want to run (configuring parameters along the way) and then going back to the top of the view and pressing Go. This workflow while not terrible does not against the more common workflow found in applications which would have the Go button at the bottom (i.e. fill in everything top-to-bottom rather than top-to-bottom-to-top). I don't think we need to move the Go button and the Quality button (which is good to check before hitting go) to the bottom of the view but it might help with the user experience to have the option on both the top (current workflow) and the bottom (more common workflow)

**Expected behaviour:** User can press Go and check the quality at the bottom of the view

**Actual behaviour:** Have to go back to the top of the view to do these things

**Reproduces how often:** 100%

GammaVel:
Thanks for all the hard work that went into solving this one @Atlas139mkm. I like the new buttons on the bottom of the pane and they do the same thing as the one on the top. I'll be curious to see which option people like the most.

PR #1395, opened 2021-09-13
Feature/make templates more obvious

Atlas139mkm:
### Description of the Change

Removed the options menu and placed its contents on the data access view.

image

This was a little bit hard to look super neat due to the existing elements on the right hand side of the screen shot and the check box texts being different sizes. However this was the best I could think of (please give feedback or suggestions on different layouts)

### Alternate Designs

I just researched and I could use toggle buttons instead of check boxes? it might look nicer however it is quite hard to tell when its toggled on as the colour change is quite hard to see. 

(I would format it nicer)
image

Easier for users to see functionality

### Benefits

UX UI

### Possible Drawbacks

potentially screen clutter? however I think its better to easily display all user options rather than hiding them under a menu.

### Verification Process

All features have maintained the same functionality. Only the layout has changed.

### Applicable Issues
#1240


arcturus2:
@Atlas139mkm  thanks for this. I have some ideas now that you've proposed an idea.

1) how about having the 2 new buttons and 2 checkbox buttons on their own line so they all appear horizontally in 1 line underneath the plus button, star button etc.

2) how about we convert the 4 options to buttons like the new step button (the + button) and have tooltips for the descriptions. I reckon we could use some buttons we have already in our icon set and also add any from https://fonts.google.com/icons that we need to. The icons could be as follows:
  * load templates - https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Afolder_open%3A
  * save templates - https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Acheck_box_outline_blank%3A
  * save results - https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Acheck_box_outline_blank%3A ??
  * deselect plugins on go - https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Acheck_box_outline_blank%3A

Note that if we go down this option then the load and save templates "From File" and "From Database" should be updated too to match.

3) I think the main problem with the existing solution was that it was still not clear there there was an options menu and what you could do. I feel that adding all these options as buttons - though is necessary, it does clutter the UI. My personal view is to go back to a menu option but besides the "Options" label on the button there is an icon too - perhaps the "settings" icon (https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Acheck_box_outline_blank%3A) which makes it more likely it is visible and curiosity to click it to see the options it contains

Besides this I feel that for the "Deselect Plugins On Go" we can omit the "Plugins" word. Essentially everything is a plugin and it probably is leading more to the technical side what developers call the bits of functionality. I added this option so its my fault and in hindsight I would omit the plugins word. If you agree then hopefully that reduces the characters and fits in nicer if you implement option 1.

antares1470:
With the current solution, it doesn't look so great on a small enough screen as demonstrated here:
DataAccessClutter

Atlas139mkm:
Data access changes
image

File import export changes
image

Atlas139mkm:
latest push
image

antares1470 (commented):
Looks a lot nicer in my mind. It still struggles with a smaller screen though:
DataAccessClutter2

arcturus2 (approved):
:heart: @Atlas139mkm. Looks so much better.

Regarding images used it looks like its only the Google material UI fonts which we have already attributed so all good there too.

Atlas139mkm:
Ill have a little play around, ill see if I can make some of the buttons go on a different row based on the width of the pane.

Atlas139mkm:
Thoughts on this?

DataAccessResizingButtons

arcturus2:
That's pretty cool @Atlas139mkm with the use of responsive design concepts. I like it. Thoughts @antares1470 ?

antares1470:
So some of the other views will have arrows that represent a "show more"
ShowMore

Don't know if that is perhaps an option (though I'd probably wanted to do that on the left hand side so as to not hide the Go button...)

I'll give the current proposed solution a test though and see how I feel about it

arcturus2:
Yeah that's neat @antares1470 forgot about that. I would say it should exclude the Go button and also the Quality Control Button Widget too.

Atlas139mkm:
I also like the way the scatter plot does it. I think the top bar on the data access view would need a redesign to make that possible as atm. those 5 buttons are just anchored to the top right of the pane, hence why they over lap other element. I would need to place them all within a HBox along with the Workflow options button. I can do that if we think that is better? I dont know if it would also be worth doing something like that for the Days in the section below?

antares1470:
DataAccessClutter3

I don't hate the buttons moving to a new row but I think having them overflow over the search bar isn't ideal. The way the scatter plot does this is nice in that it doesn't interrupt the rest of the view but as you point out @Atlas139mkm, it may be more ideal to be able to see all options (and hiding is probably the opposite of "making more obvious"). Happy for you to go with your gut on what you think is the better approach

Atlas139mkm:
Alrighty, I think I might leave it in its current state of resizing. Seeing as I imagine most users will not be running constellation on a laptop the feature will hardly be used, so probably shouldn't put too much fuss into it. The resizing works and will still be a nice feature for those on smaller screens. I attempted to do what the scatter plot had done and it was not playing super nicely either. I think the overall layout structure of the scatter plot view made it easier to implement. Also @antares1470 I'm not sure why yours was still overlapping. I'll do this push and let me know if its still occurring.

Atlas139mkm:
I just read issue #1241 I imagine this will probably change again if I implement this.

antares1470:
> I just read issue #1241 I imagine this will probably change again if I implement this.

That ticket only wants to add stuff to the bottom of the view, not remove it from the top so the work you've done here shouldn't affect that in theory

antares1470:
DataAccessClutter4

@Atlas139mkm the changes didn't help. I actually think it looked better before those changes so maybe revert that and then I think this is good to go

Atlas139mkm:
hmmm very weird. It seemed to work on mine. I'll look into it.

Also Increased the width at which it would fold because of the "Quality Category: CRITICAL" it increases the size of that button so I had to account for that when moving the content to the next row.

arcturus2:
@Atlas139mkm I'm yet to test this but based on the screenshots, the gap between the QC Widget and Go button seems to have gotten closer. We need to have these apart further like before to prevent accidental clicking of the "Go" button. I think the update with `rabRegionExectueHBox` has introduced this.

Atlas139mkm:
Apologies I can change that one back. Still unsure why Antares is overlapping the text box so much however.

Atlas139mkm:
Found why it happens, ill look for a work around.

Atlas139mkm:
This should be much better

antares1470 (approved):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessPane.java
```diff
@@ -445,7 +445,7 @@ private String getExceptionDescription(Throwable e) {
 
         // add some padding between the Go button and the previous button to avoid accidental clicking
         final Region region = new Region();
-        region.setMinSize(10, 0);
+        region.setMinSize(20, 0);
```
arcturus2, line 448: :+1:

arcturus2 (approved):

Issue #1160, opened 2021-08-03
Import from File - Search queries are not clearing when file is removed


GammaVel:
### Prerequisites

- [X ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Files and data are now being removed from the "Import from File" nicely when I click on the red remove button which is great. Thanks for fixing that. I have noticed that if there is  search criteria in the Filter section, this is not being cleared when I remove the file.  I have to go an backspace it out.  The same thing happens with the Attribute Filter. 

Now that the search is more obvious it would be really good if these two searches were automatically cleared once I removed the file. 

### Steps to Reproduce

1.  Open Import from file
2.  Select a file for importing and watch the data populate on screen
3.  Enter search criteria in the **Filter**
4.  Enter in a search term into the Attributes section
5. Now I have decided that my data is incorrect so I remove the file using the red button
6. Check the Filter and the Attribute Filter sections > these are not cleared. The file data has been cleared
7. Close the whole screen section for the import from file
8. reopen - the search data is still populated

**Expected behaviour:** when I clear the data the search criteria is cleared

**Actual behaviour:** The search criteria has to be manually deleted

**Reproduces how often:**100%

Auriga2:
@GammaVel, the filters in the table search are linked to the Run#, do you think we should clear them all, if there are few runs? I guess the multiple runs support to preserve a search, so if a user doesn't like the idea of clearing the search or if they want to reuse the query again later, they can add a new `Run` and proceed. What do you think?

The `Attribute Filter` is functional before any file is added, as it just filters the attributes irrespective of the files added. So I don't see a need to clear the search here when the files are removed.

GammaVel:
Thanks @Auriga2  and you raise a good point about the Attribute filter.   From a user perspective, once all files are cleared from the import section the Filter section should clear automatically as this makes the most sense, if I am importing a different file I won't need the filter I had set.   Down the track it might be nice to have a recent search history if possible once I start typing it could remember previous filter criteria.

[...]

PR #1396, opened 2021-09-13
Import from File - Clearing the Search filters when all files are removed

Auriga2:
### Description of the Change
In the File importer, the table filter in the configuration Run panes and the Attributes filter will now be cleared when all the files are removed.

### Verification Process

1. Open Import from file
2. Select a file for importing and watch the data populate on screen
3. Enter search criteria in the Filter in a few Run tabs
4. Enter in a search term into the Attributes section
5. Now I have decided that my data is incorrect so I remove the file using the red button
6. Check all the Filters above and verify they are cleared. 

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1160


arcturus2 (approved):

antares1470 (changes requested):
Testing looks good. If you could possibly add a whatsnew entry for this, that would be great

antares1470 (approved):

arcturus2 (approved):

Issue #1242, opened 2021-08-18
Allow editing attributes in Attribute Editor by clicking on the field


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

To edit an attribute in the Attribute Editor, you currently have to click on the Edit button next the field, rather than the field itself. This might go against intuition which may suggest you can edit the field by merely clicking on it. I think it would be good to implement this more intuitive way of editing attributes

### Steps to Reproduce

1.  Open a graph and Attribute Editor
2. Click on the field of one of the attributes
3. Click on the edit button next to the field

**Expected behaviour:** Clicking on the attribute field brings up the interface to edit the value

**Actual behaviour:** Clicking on the attribute field does nothing. Clicking on the Edit button next to it brings up the interface.

**Reproduces how often:** 100%

### Additional Information

If this is successfully implemented, this could remove the need for the Edit button (and so could probably be removed alongside this change)

Auriga2:
Should we open it at a single click or double click? Also removing the edit button will make it less intuitive for some users, at least until they get used to it?

GammaVel:
I like this little nifty change, it is much easier to edit values now. Thank you @Auriga2 for fixing this up. I like the way it works with multiple click combinations; single, double, triple.

PR #1397, opened 2021-09-14
Clicking on the attribute field brings up the interface to edit the value

Auriga2:
### Description of the Change
Clicking on the attribute field brings up the interface to edit the value. Since it allows to copy values by selecting the string with dragging the mouse, a single click can be used to pop up the edit window. Edit button can be removed here if it would not confuse users.

### Alternate Designs
opening up the edit window on the double click event

Better UX

### Benefits
Better UX

### Verification Process

1. Open a graph and Attribute Editor
2. Select a node and click on the field of one of the node attributes and verify it open up the interface to edit the value
3. Verify the value can be copied by selecting the string with dragging the mouse
4. Deselect everything and verify clicking on the node attribute doesn't populate anything

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1242


arcturus2 (changes requested):
Works so well and loving it. Nice work @Auriga2. A couple of things please:

* what's new
* an update to the attribute editor help page about the double click feature

antares1470 (commented):
It's nice to be able to simply click on the field itself to open up the edit dialog. One thing I noted though was that if you click and drag to highlight the value (or part of the value), it still opens the dialog after the mouse is released which could be a source of irritation. Don't know if there is a way to be able to distinguish between a click and a drag (so that on a drag, the edit dialog doesn't pop up)?

Auriga2:
@antares1470 Does it occur to you all the time? It happens to me only occasionally. Specifically if the mouse drag is too small. Agree it could be annoying though.

When debugging it fired a mouse click event with every drag, and event.getClickCount() was useless where it was a series of 1 and 2's for double click and it was 2 for a single click! I can use ` event.isStillSincePress()` but the getEditValueHandler() method doesn't seem to work when called like this.

```
attributeValueNode.setOnMouseClicked(event -> {
                if (event.getButton() == MouseButton.PRIMARY && event.isStillSincePress()) {
                    getEditValueHandler(attribute, editorFactory, values);
                }
            });
```
It works here:
`attributeValueNode.setOnMouseClicked(getEditValueHandler(attribute, editorFactory, values));`

Any idea why/how to overcome this?

antares1470:
@Auriga2 so I did some further testing and discovered what you were seeing, it only happened sometimes and was mainly only occurring on small enough highlights. Strangely what I found was that after a little bit, the behaviour stopped occurring and it didn't have any issues distinguishing between click and drag so I have no idea what is going on there.

Auriga2:
I believe it'd work consistently if I can use `event.isStillSincePress()`, but having trouble getting it to work with the new format. Basically when I debug, it hits line 909 and exits.
image

Looks like `getEditValueHandler` doesn't have to return an event handler in that case. I'll push a change for you to have a look.

Auriga2:
Shall we still leave the `Edit` button for a while at least?

antares1470:
> Shall we still leave the `Edit` button for a while at least, or should I remove it here?

I don't mind either way @Auriga2. On one hand, keeping the edit button allows people to slowly transition over and figure out the new way. On the other hand, if they don't see the whatsnew, they may go on blissfully unaware that the new feature exists

Similarly, removing the edit button forces users to use the new method, but maybe it makes them unsure of whether attributes are still editable if the button isn't there (?)

arcturus2:
The Edit button can only come off for the simple attributes right? Or can you do it for all objects with the double click?

I would lean towards keeping the button there for the next release and see how people feel with the new option.

Auriga2:
Can remove it for all attributes  by simply removing the button: 
image

image

Agree with you, so will keep it till next release. I'll create a ticket to remove it later.

Also noticed this doesn't work on the text area of Icon, color and blaze attributes. I am trying to fix them too.

arcturus2:
> Agree with you, so will keep it till next release. I'll create a ticket to remove it later.
> Also noticed this doesn't work on the text area of Icon, color and blaze attributes. I am trying to fix them too.

Yes thank you for both @Auriga2

Auriga2:
@arcturus2 Also note that it works for a single click (as well as double or multiple clicks.) I didn't limit it because we can still select the text using the drag event.

PR #1392, opened 2021-09-13
Data Access Pane Preferences IO Refactor

formalhaunt:
### Description of the Change

This is a follow on from the PR https://github.com/constellation-app/constellation/pull/1356.

It is now utilising the newer version of the IO code allowing us to pass it an object rather than a Jackson JSON Node. This has allowed for the clean up of a significant amount of complicated code involving several loops. It has made it easier to test and hopefully easier to understand.

I also deleted a tonne of commented out code. It hasn't been touched for years so I figure if there was an issue it would have been fixed by now.

### Why Should This Be In Core?

Improves test coverage and readability.

### Benefits

Improves test coverage and readability.

### Verification Process

Wrote unit test before and verified still worked after. One note is that in the load preferences. The `<plugin>.__is_enabled__` property was always `null` after the load. This was due to the way the JSON de-serialisation was happening. The new code will actually set that value. Looking at the save preferences code it probably wasn't an issue as that code used the actual activation flag in the object to add this property rather than looking at the properties themselves.

### Applicable Issues

N/A

PR #1399, opened 2021-09-14
:bug: Fix to data access view load templates

arcturus2:
### Description of the Change

Whilst testing https://github.com/constellation-app/constellation/pull/1392 a NPE was found which was also occurring in `master`

When a plugin has no parameters and is loaded from a template then there
is a NPE error. This has been resolved. The NPE was:

```
Parameter creation for plugin Remove Unused Attributes failed:
	at au.gov.asd.tac.constellation.views.dataaccess.panes.DataSourceTitledPane.lambda$createParameters$2(DataSourceTitledPane.java:164)
```

Used a `java.util.logger` instead of using `System.out.printf`

### Verification Process

1. In the data access view enable the "Remove unused attributes" from the Clean category
2. Options -> Save Template
3. Options -> Load Template
4. Make sure the template loads without error


antares1470 (approved):

mimosa2 (approved):
I have tested this in V2.4 and in your branch as per your test case. Although V2.4 didn't return NPE it didn't enabled the plugin which is wrong. Your branch enabled the plugin correctly and no NPE. Good work @arcturus2.

PR #1414, opened 2021-09-17
Feature/stereo

algol60:
### Description of the Change

Add anaglyphic graph display (fake 3D using colored glasses).

### Alternate Designs

This is a resurrection of code that previously existed.

Marketing. :-)
### Benefits

No functional benefit, but impressive to look at.

### Possible Drawbacks

Added code. Shaders are slightly more complex.

### Verification Process

By default, graphs still display as normal. Switching the display and back again works as expected.

### Applicable Issues

None.


arcturus2 (approved):
WOOOW! This is amazing @algol60!! Thank you very much for contributing this!

antares1470 (approved):

Issue #1407, opened 2021-09-15
Update the toolbar and welcome page icon for data access view


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

A new icon has been added on the Welcome page and the Toolbar to access the Data Access view. This is an eye which makes sense but it could be a bit off putting to some people. A specific data icon might be more meaningful.

### Steps to Reproduce

1.  Review the icon for the Data access view on the welcome page and toolbar

**Expected behaviour:** Icon reflects what the view does

**Actual behaviour:** eye could be off putting.... looks like big brother watching. 

**Reproduces how often:** 100%
### Additional Information

Thanks for considering this minor issue. I really like the shortcuts. 
![image](https://user-images.githubusercontent.com/83679004/133388712-5eee3062-4cbe-4d64-a163-2fd463804960.png)

GammaVel:
I really like the Add chart Icon which you can see on google icons here - https://fonts.google.com/icons?selected=Material+Icons+Outlined:folder_open:&icon.query=data   it shows that the graphs can be enriched using this feature.

arcturus2:
^https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Aadd_chart%3A

GammaVel:
Nice work @Atlas139mkm and @arcturus2 ! I definitely think it is better than the eye. It just occurred to me that a key might have been good too as I like to use this view to unlock my data. Maybe next time!

PR #1423, opened 2021-09-20
Changed data access view icon

Atlas139mkm:
### Description of the Change

Small and quick change.
- Changed the data access icon to the one suggested in #1407
image

Eye ball was intimidating and the chart makes more sense.

### Benefits
 
UI UX

### Verification Process

All buttons still work as intended, just changed the image

### Applicable Issues

#1407


arcturus2:
@Atlas139mkm thank you for working on this straight away and @GammaVel thank you for the suggestion and even the icon. For me though I am not convinced the bar graph icon is the right fit. To me it does not explain that the data access view is able to query data.

arcturus2 (changes requested):
I'm not sure about this icon.

Atlas139mkm:
All good, I'll see if I can find a more fitting one that portrays the message of querying data.

Atlas139mkm:
Here are a couple of options I have looked at.
1
image
2
image
3
image
4
image

I figured the magnify glass indicated querying a bit more.
Let me know your thoughts @arcturus2 @GammaVel  1,2,3 or 4?

arcturus2:
@Atlas139mkm @GammaVel I've had a bit of a rethink and this icon is warming up to me so I am OK to merge this PR in as its way better than the eye!

GammaVel:
Nice work @Atlas139mkm and @arcturus2 !  I definitely think it is better than the eye. It just occurred to me that a key might have been good too as I like to use this view to unlock my data. Maybe next time!

PR #1416, opened 2021-09-17
Histogram view added numeric label sort option

Atlas139mkm:
### Description of the Change

- Added a numeric sort option by label. Label (Number)
- Changed label option to Label (String)

Descending Number
image

Descending Number
image

Ascending Number
image

Ascending String
image

### Why Should This Be In Core?

Allows users to sort their labels by their numeric value if the labels can be converted to a number.

### Benefits

Potential use cases where the labels are numbers and the user would like to sort them in ascending or descending order.

### Verification Process

Screen shots provide various different mixes of numbers, strings, number/string combos. These show how it is working.

### Applicable Issues
#861

PR #1424, opened 2021-09-20
changed Label (String) to Label (Text)

Atlas139mkm:
### Description of the Change

Requested at the end of PR #1416 
- Changed Label (String) to Label (Text)

image

Text is more intuitive for normal users instead of String. 

### Benefits

UI

### Applicable Issues

PR
#1416


arcturus2 (approved):
Thank you @Atlas139mkm. I was going to do it but then forgot to.

PR #1429, opened 2021-09-20
Feature/stereo

algol60:
### Description of the Change

Allows for different colors for the left/right eyes of anaglyphic glasses.

### Verification Process

Checked that the normal graph drawing is not changed.


arcturus2 (approved):
Thank you for the updates @algol60

PR #1406, opened 2021-09-15
resize screen shots for welcome page

Atlas139mkm:
### Description of the Change

Please give solid feedback about it one. I found the resources of how to resize and image in java from here:
https://mkyong.com/java/how-to-resize-an-image-in-java/

I'm not 100% on what the code all does, however my understanding is it retrieves the image, resizes it and saves it under a different name. I'm not too sure on some of the in-between processes so please let me know if its good/bad or even allowed.

The changes I made ensure that no matter the size of the screenshot taken it is resized to that 145 by 145 size that welcome page displays. This means screens of all sizes can utilize the welcome page screenshots effectively.

image

### Why Should This Be In Core?

Allows small monitors to have neat images.

### Benefits

UX UI

### Possible Drawbacks

Unsure. I imagine there shouldn't be any.

### Verification Process

Tested with various monitor sizes.

### Applicable Issues
#1094

Issue #1094, opened 2021-06-07
Welcome page - Missing Images for Recent files


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Welcome page has a section for Recent Files which shows a small image of the graph and File name.  I can click on this to open the file. There is a slight issue - the image is only appearing for Files that have been previously saved and I use "Save As" to save them with a different file name. If I use " Save" or Save As" on a new graph there is no image created. There is only the file name in this section. 

There is a secondary related issue where the images are very zoomed in and appear to be a random part of the graph. It would be great if it was a full snap shot or it wasn't as zoomed in. 

### Steps to Reproduce

1.  Open a New Graph and add a node. Use Control S, File Save or Save As. 

2.  Close the graph. It won't be on the Welcome page yet. Close Constellation

3.  Reopen Constellation and check the Recent section -> the file name is shown but not the icon

4. Open the graph and make a few updates and Save it as a new graph using "File - Save As"

5. Close graph and Constellation - Check the Recent section. An image is shown for the graph

6.  Repeat these steps for an existing graph.

**Expected behaviour:**I have a small image for all of the graphs I have opened and saved in the Recent section, not just the ones that have previously been saved and saved again with "Save AS"
![image](https://user-images.githubusercontent.com/83679004/120961012-57ffaa80-c7a0-11eb-8f32-f08791b460d5.png)

The images are also not zoomed in to a small section of the graph. 

**Actual behaviour:** as per the Repro steps

**Reproduces how often:** 100%

GammaVel:
Same area as #1097

cygnus-x-1:
The images that are there are also zoomed in for some reason.

GammaVel:
Great point @cygnus-x-1 , I have updated the bug description so your comments don't get overlooked and this gets fixed :-)

Atlas139mkm:
While attempting to solve this I have discovered a few things.
[...]

PR #1430, opened 2021-09-20
Welcome View Recent Image Minor Refactor

arcturus2:
### Description of the Change

During my testing of #1406 I found some issues which I decided I wanted to have a go at fixing and propose back to @Atlas139mkm.

@Atlas139mkm has correctly addressed the issue identified by @GammaVel regarding the image resolutions but the logic has some side effects I was not happy with. The main issue was that every time you open Constellation, the `resize` method runs and resizes all the images again (and even if they are the correct size) and writes the change back to disk. This is unnecessary and a performance drain. This may have been a side effect from the last change I requested which was to override the same file.

Though this could be fixed easily by checking the size first, I have thought about the solution further and feel that the Welcome View's job shouldn't be to resize the image. In fact the image should be written at to the correct dimensions first time. I noticed that when you do a save as, the screenshot is written to disk (with the screen resolution dimensions - i.e wrong size) and then when you open Constellation again, it updates the image every time with 145 x 145 dimension.

My change in this PR moves the logic to resize to the `RecentGraphScreenshotUtilities` and it will write to disk the screenshot at the 145 x 145 dimension first time. There is also the use of `BufferedImage` so everything happens in memory and the disk is touched only at the end. This saves 2+ writes with 1 write only.

The images have also been renamed to simplify and make more consistent.

### Benefits

* Screenshot is written once at the final dimenstions and never touched again

### Possible Drawbacks

* Existing screenshots will still look broken because they are not being resized. I think this is OK because it will eventually fill up with new screenshots anyway at the proper dimensions with more use.
* It also means that the logic of the Welcome View is simplified greatly

### Verification Process

I've saved new graphs and opened old graphs. I have checked that the file size and created timestamp of the file does not change and that images are created at the correct dimensions in the `Screenshots` folder.

When images are removed from this folder, a placeholder icon appears as expected

### Applicable Issues

#1094


Atlas139mkm:
@arcturus2 I completely agree, I thought about this too when I was doing it yesterday. I was trying to think of logic to ensure it only does it if the image is not the correct size. However doing it when it takes the screen shot seems like a much better way of doing it. I'll get onto it today.

Atlas139mkm (approved):

Delphinus8821 (approved):

Issue #856, opened 2020-10-21
Find and Replace search history


antares1470:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Find and Replace plugin's "Find what" text box does not allow the user to recall previous (or subsequent) searches using ctrl+z/ctrl+y

**Expected behaviour:** ctrl-z recalls previous search and ctrl-y recalls next search
**Actual behaviour:** This functionality isn't implemented

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Atlas139mkm:
I was just looking into this one. Is the intended use for this to simply retrieve the users previous searches? or do we specifically want ctrl + z and ctrl + y? I was thinking if we just want the users previous searches, would a drop down of previous searches be easier to implement and achieve the same outcome potentially in a easier way? 

There isn't too much info on this issue so wanted to get clarification.
@arcturus2 thoughts?

arcturus2:
@Atlas139mkm sorry about the lack of info on this. The functionality would be as per other tools like Word that allow you to undo/redo text in the input text field. I don't think a drop down list is required and the added complexity that could bring. Being able to mimic what other tools do would be good for consistency at this stage.

Let me know if you need any more information.

Atlas139mkm:
@arcturus2 I'm currently looking into extending the JTextField to add some extra functionality with it to undo and redo. I'll try implementing that on a small scale first and if it works well I can work on replacing the JTextFields with the new Custom version that has the better functionality.

GammaVel:
This is handy functionality and I have confirmed it is working on the basic find, find and replace and the advanced find. Some other feedback has gone onto the Find ticket #55 .

PR #1437, opened 2021-09-22
DRAFT undo redo on JTextFields

Atlas139mkm:
### Description of the Change

I need some feedback on the implementation of this.

functionality:
- What functionality do you want?
- Ctrl + Z currently works but it works on every character or edit. For example:
    - hello | highlight hello | delete | ctrl + z | = hello 
- If I type the word "hello" Ctrl + z will remove o, because it is the last edit. 
    - hello | highlight hello | delete | ctrl + z | = hello
- If I highlight the word "hello" delete it then Ctrl + z, it will bring back the whole word as this was the last edit.
so my question is are you okay with ctrl z only bringing back the last edit? or should when I type "hello" should Ctrl + z remove the whole word?

So should:  hello | ctrl +z | 
equal
- hello
or 
-

If you would like the whole word gone, this would require me to do extra implantation by extending the UndoManager class. Which probably can be done, however will require more effort. But since this is going to ideally be used application wide for swing textFields maybe you want it working closer to how javafx and other applications do it.

Class locations:
currently I'm just putting the classes in the find view as this is where the original issue was directed at. However were would prefer these classes end up within constellation?

### Applicable Issues
PR #856

### DIFF
--- a/CoreFindView/src/au/gov/asd/tac/constellation/views/find/gui/BasicFindPanel.form
+++ b/CoreFindView/src/au/gov/asd/tac/constellation/views/find/gui/BasicFindPanel.form
@@ -2,13 +2,13 @@
 
 <Form version="1.8" maxVersion="1.8" type="org.netbeans.modules.form.forminfo.JPanelFormInfo">
   <NonVisualComponents>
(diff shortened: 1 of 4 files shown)


antares1470:
I believe this is superceded by #1489

PR #1438, opened 2021-09-22
:art: Allow programmatic setting of verbose mode for ImmutableObjectCache

groombridge34a:
### Description of the Change

`ImmutableObjectCache` has two paths of logic. One of these paths inspects objects added to the cache and keeps statistics on how the cache reacts to each object, which can be retrieved using the `toString` method. This is called `verbose` mode. The other path doesn't keep statistics and is consequently much faster.

`verbose` mode is not enabled by default.

Previously the only way to enable `verbose` mode was to make a code change to `ImmutableObjectCache` - if a developer changed the `VERBOSE` variable to `true` the cache would run in `verbose` mode. This made `verbose` mode hard to test. The improvement made was to add a setter allowing programmatic setting of the `verbose` flag. This retains the functionality of the original code as the developer can change the default of `verbose` from `false` to `true` before running their developer instance, but also allows classes that use `ImmutableObjectCache` to choose if they would like `verbose` mode or not. Which also facilitates unit testing.

### Alternate Designs

`ImmutableObjectCache` could log diagnostic information on each cache addition. This would be significantly slower than the existing non-verbose mode.

`verbose` could be set in the constructor of `ImmutableObjectCache`. But this would lose the original behaviour of allowing the developer to set all `ImmutableObjectCache` instances to `verbose` mode.

Update to an existing core class.

### Benefits

This change allows classes that use `ImmutableObjectCache` to choose if they would like `verbose` mode or not. Which also facilitates unit testing.

### Possible Drawbacks

None.

### Verification Process

Added comprehensive unit testing.

### Applicable Issues

#413

### DIFF
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/ImmutableObjectCache.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/datastructure/ImmutableObjectCache.java
@@ -30,15 +30,23 @@ import java.util.Map.Entry;
  */
 public class ImmutableObjectCache {
 
-    private static final boolean VERBOSE = false;
+    private boolean verbose = false;
 
-    private final Map<Class<?>, int[]> results;
+    private Map<Class<?>, int[]> results = null;
     private long savedStringBytes = 0;
 
     private final Map<Object, Object> cache = new HashMap<>();
-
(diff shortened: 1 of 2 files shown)


arcturus2 (approved):
This is great thank you @groombridge34a

PR #1459, opened 2021-09-29
Show exception to user in executable version v2 + Exception Refactor

arcturus2:
### Description of the Change

This is a follow up to the PR https://github.com/constellation-app/constellation/pull/1134 which I've closed and started a fresh one because that was getting too busy.

I have worked on the same branch @aldebaran30701 started (thank you) and made further changes.

To summarise the main problem with #50 was that in the executable version exceptions were not presented the same way when run from within the IDE. The executable version would indicate there was an error with a little red icon down the button right corner but it was not obvious.

This change forces exceptions to be shown to the user and the main solution here is the addition of `netbeans.exception.report.min.level=900` and `netbeans.exception.alert.min.level=900` with both set to `900` to make sure all error above the `WARNING` threshold will present a dialog box.

There have been some ongoing issues which @aldebaran30701 has addressed and I have further worked on. These are mainly:
* Inconsistencies with exception handling with the `SimpleReadPlugin`, `SimpleEditPlugin`, `SimpleQueryPlugin` and the `DefaultPluginEnvironment` whereby sometimes you get an exception, sometimes 2 (one from the plugin and another from NB) and on rare occasions (unproven) but no exception.
* Inconsistencies with logging to file
* Decided not to wrap exceptions around `RuntimeException()`

The logic now is that after a plugin runs, when there is any exception, it is first "seen" (and not caught) by the Simple* plugin so the exception bubbles up to the `DefaultPluginEnvironment` which runs every plugin to deal with it. The `DefaultPluginEnvironment` handles all cases and notifies the user accordingly + also consistently logging it. The plugin environment treats `InterruptedException` and `PluginException` with a bit more attention and then any other type of error is given to NetBeans to deal with. This is all now consistantly handled by the `reportException` method @aldebaran30701 originally started.

NB reports `FATAL` and `ERROR` with a dialog shown to the user, `WARNING` is a bubble and `INFO` is the status notification at the bottom left. All of this is documented in the `DefaultPluginEnvironment` and `DefaultPluginInteraction`.

### Alternate Designs

Notification bubble was tested, but a dialog was seemingly the best option as it stops the user continuing without closing the dialog. Though notification bubbles are cool, for serious errors we prefer to force the user to deal with it straight away.

### Possible Drawbacks

This change could have simply included the 2 arguements and largely achieved the requirements of #50 but a refactor was also done to make everything more consistent based on changes made through the years.

As a result this is a fundamental change and something could have broken as a result so a good amount of testing is required. Any change to the `Simple*` classes and `DefaultPluginEnvironment` which every plugin uses needs a detailed amount of testing.

I'm pretty confident with the changes but just wanted to highlight the risk of making changes to these fundamental classes.

### Verification Process

Tested plugins like those available from the Experimental -> Developer menu to test `PluginException`'s and `RuntimeException`'s. Also tested errors thrown from the Data Access View like when parameters are missing etc.

Testing was first done within the IDE but then the distribution version was tested to make sure it behaved the same.

Verified there was dialog messages shown when errors occurred.

Verified that only 1 dialog box was shown to the user.

### Applicable Issues

#50


arcturus2:
I'm keeping this as draft PR as I want to add some unit tests.

arcturus2:
Alright @formalhaunt @antares1470 @aldebaran30701 so I did try to refactor the duplicate code that exists with `executeEditPluginNow` and `executeReadPluginNow`but because one requires a `GraphWriteMethods` and `GraphReadMethods` respectively I tried using the `GraphReadMethods` for both but it then **correctly** caused plugins that wanted write locks to fail with an exception.

I tried refactoring internal parts of the methods but that also was piecemeal and did not really help. I also tried to pass a generic `Class` object and use the `.cast()` method but that was also not working well so I've given up.

arcturus2:
Thanks @formalhaunt for the reply and providing those extra tips. Nearly everything checks out and the part that still is the issue is what I mentioned earlier which is the `plugin.run()` call:

```
plugin.run(manager.getGraph(), interaction, parameters);
```

We can't use a `manager.getGraph()` because it returns a `Graph` which still needs to be casted to a `GraphReadMethods` and `GraphWriteMethods` for the `executeReadPluginNow` and `executeEditPluginNow` respectively. 

We can't use generic types either like `<T>` as the same `T` needs to support different types for the same instance of `DefaultPluginEnvironment`.

Screenshot from 2021-11-18 13-22-49


image

arcturus2:
@formalhaunt @antares1470 ready for this to go into master if all ok as is?

PR #1465, opened 2021-10-01
Welcome Page Recent Graphs Exception

antares1470:
### Description of the Change

The issue described here #1284 was not as fixed as we first thought as the exception came up again. Adding some logs to what was being loaded suggested that the open recent menu item used the '/' path separator whereas the welcome page used '\'. I reckon there is a possibility that this is the cause of the issue (though why it only happens sometimes and not always baffles me)

I've made an adjustment to the welcome page one to replace the path separator to mimic what is happening in the open recent menu item. This should hopefully fix the issue once and for all (fingers-crossed)

Bug fix

### Benefits

Bug fix

### Applicable Issues

#1284

### DIFF
--- a/CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/open/RecentFilesWelcomePage.java
+++ b/CoreGraphFile/src/au/gov/asd/tac/constellation/graph/file/open/RecentFilesWelcomePage.java
@@ -66,7 +66,8 @@ public class RecentFilesWelcomePage {
             }
         }
         if (index != -1) {
-            final String path = files.get(index).getPath();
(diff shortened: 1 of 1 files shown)


arcturus2 (approved):

antares1470:
So I was finally able to get a reproducer for this and it appears it is still occurring so I will do some more digging on this one

antares1470:
I've now also included a fix for the open recent files duplication. As I was looking at this issue I realised that the code I wanted to use was in this PR so figured I would simply include the fixes as a part of this PR rather than trying to handle it in a separate PR

Issue #841, opened 2020-09-08
Investigate long term solution for Invisible text appearing in Schema View


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Following on from the work done in https://github.com/constellation-app/constellation/issues/585, a work around solution was done because all options where exhausted at the time. This task will be to potentially undo the change in PR https://github.com/constellation-app/constellation/pull/836 and identify the root cause.

Auriga2:
Looks good in master-2021-10-11

PR #1466, opened 2021-10-01
Schema View Filter css

antares1470:
### Description of the Change

Have moved away from the temp solution for the schema view css  my making the needed minor adjustments to more align it with other filter search bars in other parts of the application while also solving the issue that the temp solution originally aimed to resolve.

Better style alignment across the application

### Benefits

More familiar style

### Applicable Issues

#841

### DIFF
--- a/CoreSchemaView/src/au/gov/asd/tac/constellation/views/schemaview/resources/schema-view.css
+++ b/CoreSchemaView/src/au/gov/asd/tac/constellation/views/schemaview/resources/schema-view.css
@@ -3,7 +3,7 @@
     -fx-background-color: rgb(30,144,255); -fx-padding: 2 5 2 5;
 }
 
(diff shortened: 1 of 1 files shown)


arcturus2 (approved):

Auriga2 (changes requested):
Changing the  `-fx-text-inner-color` hasn't fixed the issue when we handled it last time, as per the notes in https://github.com/constellation-app/constellation/issues/585. So I removed both these lines and it still works. Probably this was fixed by some other changes in between?

aldebaran30701 (approved):
Static code looks good, have not tested in the app.
Not sure if it is worth digging to find out what changed in the app that made this work as intended. just the fact that it works should be fine.

Auriga2 (approved):

Issue #633, opened 2020-07-07
Mark other parameters  as `*required` similar to the `CONSTELLATION secret`


Auriga2:
In the Rest API there are no checks for missing parameters. Instead they should also be marked as `*required` similar to the `CONSTELLATION secret`
![image](https://user-images.githubusercontent.com/64826928/86327634-c54ab780-bc82-11ea-8cc6-df3f84fdb7bf.png)

Auriga2:
arcturus2 : In the `SwaggerServlet` I saw

`param.put("required", false); // TODO Hard-code this until PluginParameters grows a required field.`
so once we implement the required property for plugins we should be able to get this automatically in Swagger.

Auriga2:
`/v2/service/set_graph_values` end point doesn't seem to work even with a valid graph ID. Looks like it requires a new body parameter similar to the `/v2/service/add_recordstore`

PR #1467, opened 2021-10-01
REST API - Mark required parameters and fix the `set_graph_values` end point

Auriga2:
### Description of the Change
Following improvements have been made in the REST API :

- Marked the required parameters as `*Required`
- Other parameters are mentioned as `Optional` or default values are mentioned in the description.
- Fixed the `set_graph_values` end point

Better UX

### Benefits
Better UX

### Verification Process

1. Open Swagger (Run constellation and Tools->Start Rest Server then Display Rest Server Documentation)
2. Verify `set_graph_values` end point can now add graph attributes
3. Verify parameters in all the end points are marked as `Required` or mentioned as optional or the default valued is specified.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/633

### DIFF
--- a/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/PluginParameter.java
+++ b/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/PluginParameter.java
@@ -74,6 +74,7 @@ public class PluginParameter<V extends ParameterValue> {
     private String helpID;
     private boolean isSuppressed = false;
     private String requestBodyExample;
+    private boolean isRequired = false;
 
     private final List<ParameterChange> suppressedEvents = new ArrayList<>();
 
@@ -828,6 +829,7 @@ public class PluginParameter<V extends ParameterValue> {
     public final String getRequestBodyExampleJson() {
         return requestBodyExample;
     }
+
     /**
      * Set the swagger Request Body Example value.
      *
@@ -837,5 +839,25 @@ public class PluginParameter<V extends ParameterValue> {
         if (!Objects.equals(requestBodyExample, this.requestBodyExample)) {
             this.requestBodyExample = requestBodyExample;
         }
+    }
+
+    /**
+     * Is the parameter required? Required parameters are marked as required in
+     * swagger.
+     *
+     * @return True if the parameter is required, false otherwise.
+     */
+    public boolean isRequired() {
+        return isRequired;
+    }
+
+    /**
+     * Set whether the parameter is required. Required parameters are marked as
+     * required in swagger.
+     *
+     * @param isRequired A boolean indicating whether the parameter is required.
+     */
+    public void setRequired(boolean isRequired) {
+        this.isRequired = isRequired;
     }
 }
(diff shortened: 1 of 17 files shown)


antares1470 (changes requested):
Looks good. Just a small thing but are you able to add a changelog entry mentioning that a parameter being required is now configurable

antares1470 (approved):

arcturus2 (changes requested):
So I like this change but I think the `isRequired` needs to be updated for all plugins and similar to the Rest API with the asterisk, it should be shown show for any popup plugin and in the Data Access View.

Ideally I feel it should be part of this change but I don't mind if we do it in a separate followup PR but it should be done sooner rather than later.

Thoughts?

Auriga2:
Yeah that would be a nice improvement, but this was mainly avoiding a lot of Illegal argument and null pointer exceptions in the Swagger. I think a separate issue would suit better.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Constellation Changes
 
+## Changes in October 2021
+
+-   Added `isRequired` in `PluginParameter` with a getter and a setter, which 
+    can be used to configure the required plugin parameters to mark as `*required`
+	in the swagger.
```
aldebaran30701, line 7: How does the below sound? I felt API needed to be mentioned, and maybe that swagger was an interface.

Added `isRequired` in `PluginParameter` with a getter and a setter. This adds a `*required` annotation to the REST API parameter within the Swagger interface.

aldebaran30701 (commented):
Static code analysis looks good to me - Just one comment. 
Do you think the behaviour of the end points deserved some test cases? 
ie. does the API enforce the required tag that you can test the response for?

CHANGELOG.md
Auriga2: How about this then, because the changes are only visible in the swagger currently:
"Added `isRequired` in `PluginParameter` with a getter and a setter, which can be used to configure the required plugin parameters to mark as `*required`	in the swagger interface of the REST API."?

There are no tests added in this area yet, and from what I can see, we can add tests to assert the endpoints, but it could be impossible or tricky to verify this specific change, and I don't see any value there anyway, as it's just enforcing the swagger users to enter the required parameters before executing.

Auriga2:
`So I like this change but I think the isRequired needs to be updated for all plugins and similar to the Rest API with the asterisk, it should be shown show for any popup plugin and in the Data Access View.`

@arcturus2 On a second thought I guess we already handle this situation better, dynamically in the UI. It disables non-required fields and pops up custom error messages when required fields are missing. (E.g. In the DAV)

The only improvement I can think of is to add a check for required parameters before running a plugin and then throw a better error message. It'll make more sense to the user when the plugins are run through `run_plugins` or `run_plugins` with missing parameters in the swagger or the Notebook. Some plugins already work well, such as the ones in the DAV. So probably we need to fix the rest. (HISTOGRAMFILTERONSELECTION, EXPANDCOMPOSITE etc.)

image
image

Issue #1468, opened 2021-10-01
Test Fail logs in CI


aldebaran30701:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
When a PR is running on the CI to execute all tests, the test runner will generate a report in the logs.
The issue is that it does not show exactly the file or test method that failed. This can cause long delays in finding the source of the test failure.

Adding possible ant tags or testng tags may assist with this - although it appears that it is already logged at a verbose level using testNG.
https://testng.org/doc/ant.html
https://javadoc.io/static/org.testng/testng/6.13/org/testng/reporters/VerboseReporter.html

As an example, it is hard to tell which test actually failed from https://github.com/constellation-app/constellation/pull/1462/checks?sha=d484cec72775d68515ffda11f851a412eba09e98

Ideally, it would print the name of the method which failed, rather than the whole module of tests which had a failure.
\[Description of the issue\]

### Steps to Reproduce

1.  \[First Step\]

2.  \[Second Step\]

3.  \[and so on…\]

**Expected behaviour:** \[What you expect to happen\]

**Actual behaviour:** \[What actually happens\]

**Reproduces how often:** \[What percentage of the time does it reproduce?\]

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

arcturus2:
Thanks @aldebaran30701 for the feedback. I think I can fix this by adding the `-verbose` flag to the `ant` command. It will take longer to run and more output but did you want me to try that?

arcturus2:
@aldebaran30701 yep loving the idea.

>Alternately this could be a parameter parsed in to the file.
I would suggest modifying the `run-tests.sh` script to pass this as a parameter.

Regarding the label I do also have a suggestion being just calling it what it is so "verbose logging" ?

PR #1470, opened 2021-10-01
:white_check_mark: Verbose logging during unit testing

arcturus2:
### Description of the Change

Verbose logging to help with finding out when unit tests fail.

### Possible Drawbacks

Unit tests can take longer to run and there is a lot more logs to go through.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/1468

### DIFF
--- a/.githubutilities/run-tests.sh
+++ b/.githubutilities/run-tests.sh
@@ -14,7 +14,7 @@ ant \
 
 title "Run Core Unit Testing"
 
-ant \
+ant -verbose \
   -Dnbplatform.active.dir="${NETBEANS_HOME}" \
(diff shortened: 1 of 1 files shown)


aldebaran30701 (approved):
I'd say go for it, it can always be reverted if it is causing issues.

antares1470 (approved):
agreeing with @aldebaran30701. Should definitely keep an eye on this and its impact

arcturus2:
@antares1470 @aldebaran30701 so I downloaded the logs to compare the site and the unit test run has gone from 2.1MB to 70.2 MB. The issue is that the browser has a lot more work to do to load the logs and searching will be a bit slower too.

Execution time increased by 1 min so not that bad.

I feel we should revert it.

PR #1471, opened 2021-10-02
TableView Export Bugs

formalhaunt:
### Description of the Change

Created a core utilities class for the opening of file chooser dialogs. This should be used in any location where a file chooser needs to be opened. It is protected against being opened on either a UI thread or non UI thread. This new implementation has also been used in table view to fix a MacOS issue where the file choose was not opening.

This implementation used the OpenIDE `FileChooserBuilder` which unless you have the system property `nb.native.filechooser` set to true will use the Swing version rather than the AWT version which is in line with the decisions made about this previously.

Fixed a `NullPointerException` in table view export where if the user fails to select a file by hitting cancel or through some other means a `NullPointerException` would occur.

Fixes two bugs and provides common code to reduce duplication across the code base.

### Benefits

Fixes bugs and reduces the chances of errors being made when the file chooser dialogs are opened.

### Possible Drawbacks

N/A

### Verification Process

Both fixes have been fully unit tested.


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/filechooser/FileChooser.java
```diff
@@ -0,0 +1,126 @@
...
+        final ShowFileChooserDialog showDialog = new ShowFileChooserDialog(
+                fileChooserBuilder, fileDialogMode);
+        
+        try {
+            // Make a request to open the file chooser dialog on the UI thread
+            EventQueue.invokeAndWait(showDialog);
+        } catch (InterruptedException ex) {
+            LOGGER.log(Level.WARNING, "Thread displaying the file chooser was interrupted.", ex);
+            Thread.currentThread().interrupt();
+        } catch (InvocationTargetException ex) {
```
antares1470, line 119: exceptions final

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/filechooser/FileChooser.java
```diff
@@ -0,0 +1,126 @@
...
+            LOGGER.log(Level.SEVERE, "Error occured during selection in file chooser.", ex);
+            Exceptions.printStackTrace(ex);
```
antares1470, line 121: can you remove this stack trace print?

antares1470 (changes requested):
Are you able to add an entry to changelog as well with this change to the file chooser?

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/filechooser/FileChooser.java
formalhaunt: Done

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/filechooser/FileChooser.java
formalhaunt: Done

arcturus2 (changes requested):
This is very impressive @formalhaunt. I tested it and I only had a couple of minor spelling errors I picked up in the review.

Similar to @antares1470 I also got the Data Access View open in the location where the graphs open. This has to do with how the underlying NetBeans Platform framework maintains state of the views. The properties file in the cache area has a reference to the full class path of the top component. Given that it's moved packages, there is an exception thrown at load time and the framework is unable to handle the exception neatly and I suspect it results in the `explorer` position being ignored.

The only way I found I could fix this was to clear the user cache. We might need to include a simple script with the installer version to clear users cache if they come across this issue. We can add something to a "common issues" section on the website. I guess we could also try to automate it and clear the cache without/without the user knowing and set a flag file somewhere so we know it was done for this version.

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/filechooser/FileChooser.java
antares1470: I believe the throwable parameter in the logger will do this (from my recollection)

CoreTableView/test/unit/src/au/gov/asd/tac/constellation/views/tableview2/components/ExportMenuNGTest.java
```diff
@@ -78,13 +90,18 @@ public ExportMenuNGTest() {
...
+        }
     }
 
     @AfterClass
     public static void tearDownClass() throws Exception {
-        FxToolkit.hideStage();
+        try {
+            FxToolkit.cleanupStages();
+        } catch (TimeoutException ex) {
+            LOGGER.log(Level.WARNING, "FxToolkit timedout trying to cleanup stages", ex);
```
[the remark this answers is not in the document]
formalhaunt: Fixed

aldebaran30701 (commented):
Noted that this is only a static code analysis, just one comment. Approved after the change is made

antares1470 (approved):
Good to go once @aldebaran30701's comment has been addressed

Issue #1278, opened 2021-08-24
Average Degree Analytic appears to ignore parameters


antares1470:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Average Degree analytic has 3 booleans for determining whether to include incoming, outgoing, or undirected transactions in the calculation. Problem is that it doesn't seem to matter which ones you have or haven't ticked as it always gives the score it would have if all those options were ticked.

### Steps to Reproduce

1. Create a graph with some number of nodes and transactions (at least 1 pair of connected nodes)
2.  Open Analytic View and select the Average Degree Analytic 
3. Untick all the parameter options
4. Run the analytic

**Expected behaviour:** Score should be 0

**Actual behaviour:** Score is > 0 (more specifically, the score if all of the parameters are ticked)

**Reproduces how often:** 100%

### Additional Information

Can confirm this is happening in v2.3 as well so I think it has been a lingering issue. Unsure whether this is occurring in other analytics but a test could confirm this

antares1470:
I've observed a similar issue for Betweeness Centrality analytic. Once again, those 3 specific boolean parameters (include incoming, outgoing, undirected) appear to be ignored. I'm going to try this for a few more analytics and see what happens

antares1470:
@arcturus2 so looking at the code for Average Degree, the calculation is essentially the sum of degrees / number of nodes. The number of nodes though doesn't represent the total number of nodes but rather the number of verticnodeses whose degrees are greater than 0 + their neighbours. This means that if none of the nodes have a degree greater than 0 (e.g. you untick incoming, outgoing and undirected), you'll be diving by 0 (which you are in your case), hence the answer is NaN (Not a Number)

PR #1472, opened 2021-10-05
Analytic transaction parameters

antares1470:
### Description of the Change

A few of the analytics that had incoming,outgoing,undirected parameters were producing very odd results when one of the normalised options were selected. I didn't dig enough to determine where exactly the numbers were coming from in those instances but I was able to determine that the analytic plugins ends early if nether incoming, nor outgoing was selected for the analytics in question. Slightly altering the logic for determining the attribute to be used allowed the plugin to continue and consequently produce the expected results based on the calculations that were made.

Note there are probably still some issues with the results in some cases but these issues are now solely with the algorithm itself and I believe are outside the scope of this work.

While I was at it, I also observed that none of the *Action classes for the analytics were being used so I removed them to ease maintenance.

Fixes an issue with some of the analytics

### Benefits

Results shown better reflect parameter configuration

### Possible Drawbacks

Algorithms still have issues I guess.

### Applicable Issues

#1278


aldebaran30701 (approved):
Approving based on static code analysis, looks good to me and the logic seems to be changed to reflect the root of the issue. (which is good)

Issue #1318, opened 2021-08-30
File Import - not able to import all JSON files (related to JSON lists)


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Constellation is throwing an error when trying to import JSON lists, which seems to be a popular JSON format. This can be seen when importing the JSON files available from this data science website - try importing these: 
https://www.kaggle.com/cmenca/new-york-times-hardcover-fiction-best-sellers
https://www.kaggle.com/aritzg/donostia-san-sebastian-temp-rain-bike-demand

### Steps to Reproduce

1.  Import JSON files from the above website
2.  An error message is displayed warning that they are not valid file types - See additional information
3.  Observe there is also a run time exception and error in the logs

**Expected behaviour:** JSON file types can be imported

**Actual behaviour:** Error message

**Reproduces how often:** 100% for these file types

### Additional Information

![image](https://user-images.githubusercontent.com/83679004/131287169-a15563b6-66a3-49d4-b230-052bd5be4544.png).

Guilty-Spark-343:
Is this specific to newline delimited JSON (ndJSON) files?  The new york book lists appear to be in this format but the bike demand datasets are nested dicts without lists which would be a really hard problem to work out which dicts would be used (especially with different sizes) and all of the various edge cases.   I think ndJSON would be a nice addition, I'm keen to make the change if I've interpreted the request correctly? 

New York Book list  / ndJSON example:
![image](https://user-images.githubusercontent.com/63685177/136151575-0b3025be-ee3b-4218-9295-23117689ce67.png)

PR #1478, opened 2021-10-06
Added the processing of newline delimited JSON (ndJSON) to JSONImport…

Guilty-Spark-343:
### Description of the Change
Added the ability to parse newline delimited JSON files where multiple valid JSON objects are stored within one file, used heavily in streaming data.  Previously the JSON parser read the first entry of the file then stopped, the new implementation is a small change that will read all of the objects and add them to a valid JSON object in the format {"results": [<each ndJSON object>]}.  Normal JSON files will be unchanged.  This PR will address the first part of ticket #1318.

Example ndJSON:
image

Example ndJSON after processing (with additional formatting):
image

### Alternate Designs

None considered.

ndJSON is a rather common format and the change to allow parsing will add minimal overhead and complexity to core.  

### Benefits

ndJSON is now processed seamlessly with the Json parser.

### Possible Drawbacks

None

### Verification Process
Steps to verify:
- Open the file delimiter pane - File > Import > From File...
- Download a ndJSON dataset e.g. https://www.kaggle.com/cmenca/new-york-times-hardcover-fiction-best-sellers
- Add the file to the file importer
- You will now see the data in the preview bar
The resulting JSON will be processed like a normal JSON file.
image

### Applicable Issues

#1318

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/parser/JSONImportFileParser.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/parser/JSONImportFileParser.java
@@ -404,14 +404,18 @@ public class JSONImportFileParser extends ImportFileParser {
             }
             // Maps newline delimited JSON to valid JSON in the format 
             // {"results": [<ndjson>]}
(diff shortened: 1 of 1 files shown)


Guilty-Spark-343:
Thanks for the review :), I made the changes and pushed them.

Issue #991, opened 2021-03-16
Select Top N Multi choice parameter options updated incorrectly


antares1470:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

In the Select Top N plugin, the Type Category parameter updates the options to pick from in the Specific Types parameter and the number of possible options in this parameter differs for each type category.

When you change the type category selected, if the number of specific types to pick from is less than the number of previously selected ones (for the previous type category), these extras will be replaced with null as having been selected (not available in the drop-down list). The number of nulls 'selected' equates to the difference between the number of choices previously selected and the number of available options for the new type category (as all options are selected by default).

### Steps to Reproduce

1. Open up DAV and expand Select Top N plugin
2. Set Mode to Node
3. Set Type Category to Network Identifier and see specific types selected
4. Set Type Category to Location and see specific types selected
5. Set Type Category to Event and see 3 nulls in specific types selected
6. Set Type Category to Network Identifier again and see 2 nulls in specific types selected

**Expected behaviour:** No nulls in list of selected multi-choice options

**Actual behaviour:** null values appearing when updating multi choice parameter options

**Reproduces how often:** 100%

### Additional Information

Not sure whether this behaviour is specific to Select Top N or whether it could be a problem for any plugin with a similar process for updating parameters

mimosa2:
I have tested this and it is working fine.

PR #1498, opened 2021-10-15
 Multi choice parameter options update correctly without null additions.

Auriga2:
### Prerequisites

### Description of the Change
Removed unnecessary null additions in the preview text when the items list is reduced in a MultiChoiceComboBox.
E.g. In the Select Top N plugin, the Type Category parameter updates the options to pick from in the Specific Types parameter and the number of possible options in this parameter differs for each type category.

When you change the type category selected, if the number of specific types to pick from is less than the number of previously selected ones (for the previous type category), these extras will be replaced with null as having been selected (not available in the drop-down list). The number of nulls 'selected' equates to the difference between the number of choices previously selected and the number of available options for the new type category (as all options are selected by default).
image

### Alternate Designs
Perhaps a better solution in the controlsfx library, to not rely on the items count when clearing checks, and even better, clear the checks when clearing the options.

Better UX

### Benefits
Better UX

### Verification Process

1. Open up DAV and expand Select Top N plugin
2. Set Mode to Node
3. Set Type Category to Network Identifier and confirm 2 specific types are selected and they populate comma separated in the preview.
4. Set Type Category to Location and confirm 4 specific types are selected  and they populate comma separated in the preview.
5. Set Type Category to Event and confirm only one type type is selected,  and they populate comma separated in the preview with no 3 nulls 
6. Set Type Category to Network Identifier again and confirm 2 specific types are selected,  and they populate comma separated in the preview without any nulls.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/991

### DIFF
--- a/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/MultiChoiceInputPane.java
+++ b/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/MultiChoiceInputPane.java
@@ -97,7 +97,7 @@ public class MultiChoiceInputPane extends HBox {
                 switch (change) {
                     case VALUE:
                         isAdjusting = true;
-                        field.getCheckModel().clearChecks();
+                        field.getCheckModel().clearChecks(); //The order matters here- this should be called before clearing the options.
                         options.clear();
                         options.addAll(MultiChoiceParameterType.getOptionsData(mcPluginParameter));
                         @SuppressWarnings("unchecked") //checkedItems will be list of parameter values


Auriga2:
The issue could be simply fixed in CheckComboBoxSkin:
image

But the real issue is it not refreshing the selectedItems.size(). I tried below, but they didn't work.

   checkComboBox.getCheckModel().clearChecks();
   checkComboBox.getItems().clear();

When I debugged with the controlsfx code, the CheckComboBox.clearChecks doesn't clear checkedIndices because getItemCount is always zero.
image

 That was because we cleared options before clearing checks. Adding them in opposite order fixed the issue!
image

arcturus2:
@Auriga2 @aldebaran30701 totally agree. I was unable to find this when I had a go at it.

Issue #1426, opened 2021-09-20
Longer start up times if Import from File &/or Database is open and Import from file is taking longer to load


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Constellation is taking longer to load if the Import from File or Database is open from the previous session. It adds approx. 10 seconds to the start up time. Instead of 8 seconds on my Windows device it takes around 18/19. 

Import from File is slow to load for the first time (if it wasn't open on launch).  It takes around 10 seconds. There is no slowness when launching the Import from Database however it does appear to impact the start up time if it was already open. 

### Steps to Reproduce

1.  Set up . Make sure you don't have any views or import screens open in Constellation. Close and reopen. Time how long it takes for Constellation to launch.  

2.  Next open the Import from File screen and time how long this takes to display fully. Keep open and close Constellation

3.  Relaunch Constellation and repeat timing > Observe noticeable delay 

4.  Repeat with the Import from Database, see how long it takes to load, and how it impacts the start up time of Constellation. 

**Expected behaviour:** Having either of the import screens open does not impact the Constellation start up time

**Actual behaviour:** Noticeable delay

**Reproduces how often:** 100%, other views seem to also delay the start up time but not by as much. 

### Additional Information

Startup time nothing open: 8 sec
Startup time with Import from file open previously: 19 sec
 Startup time with Import from DB open previously: 19 sec

Open time for Import from file for the first time in a new session (not open before): 12 seconds
Version 2.4 start up time - 4 seconds

serpens24:
Performing a few timings to get a feel for what is occuring.

Installed Constellation 2.4.0 from https://github.com/constellation-app/constellation/releases/tag/v2.4.0
- Start Consty:                                                                     8.66 seconds
- Selected File->Import->From File...                                 3.11 seconds
Ensure the Delimited File... form is docked into the main Constellation application and shutdown
- Re-Start Consty:                                                               10.61 seconds

[...]

PR #1502, opened 2021-10-18
Importer improvements

antares1470:
### Prerequisites

### Description of the Change

Changes made recently to allow the filter in the file and db importers to work from the get go meant that the importers were now taking a lot longer to load (this includes application load time if an importer was open at the end of a previous session). I've reverted that change so that the speed of loading is now back.

This means the filter is no longer available from the get go so I've added in some changes to help with that. Firstly changed the prompt text in the filter box to indicate it is unavailable and then switches to the current prompt text once it becomes available. I then made changes so that the the process for loading the filter happens from app startup rather than when an importer is opened.

In this case speed difference is significant enough that it is preferred to the filter being available from the get go. UX improvements around that then are more helpful for the user

### Benefits

Importers don't take forever to load, users more informed about filter availability

### Possible Drawbacks

The obvious one is the filter isn't available from the get go. Hopefully the UX improvements in here though help the user understand better what is going on.

### Applicable Issues

#1426

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/FilterStartUpLoad.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/FilterStartUpLoad.java
@@ -29,7 +29,7 @@ public class FilterStartUpLoad implements Runnable {
     @Override
     public void run() {
         //pointless statement to trigger RunPane proesses
-        final Future<RowFilter> f = RunPane.FILTER_LOAD;
+        final Future<Void> f = RunPane.FILTER_LOAD;
     }
     
 }
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RunPane.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/RunPane.java
@@ -31,11 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.event.ActionEvent;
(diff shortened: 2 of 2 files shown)


antares1470:
@Auriga2 did you get that with the most recent version of the branch? The exception you saw was one I saw early on but haven't encountered since this commit I made: https://github.com/constellation-app/constellation/pull/1502/commits/2344134f31b7c1e2cdc195e7813af27c78f470f0

Auriga2:
Ah yes, I checked out the branch today but somehow didn't have the latest version!

Auriga2:
Not sure whether it's related to your changes, but I get this exception when importing to a `New Bare Graph`:
image

```
java.lang.ArrayIndexOutOfBoundsException
	at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.executePluginNow(DefaultPluginEnvironment.java:237)
	at au.gov.asd.tac.constellation.plugins.PluginExecutor$PluginEntry.executeNow(PluginExecutor.java:406)
	at au.gov.asd.tac.constellation.plugins.PluginExecutor$1.execute(PluginExecutor.java:256)
	at au.gov.asd.tac.constellation.plugins.templates.SimplePlugin.run(SimplePlugin.java:68)
[catch] at au.gov.asd.tac.constellation.graph.node.plugins.DefaultPluginEnvironment.lambda$executePluginLater$0(DefaultPluginEnvironment.java:130)
```

antares1470:
> Should we log a new ticket to handle that?

@Auriga2 Possibly could but I personally don't think it will be worthwhile at this stage. I think the issue stems more from the Bare Graph itself rather than how it is handled. While a Bare graph in theory sounds good, you realise that there are some attributes that it wouldn't make sense to go without anyway. I think then it is probably more a review of the bare graph that needs to be done rather instead of handling it properly, but I'm not sure that is worth a ticket at present

antares1470:
> Should we log a new ticket to handle that?

@Auriga2 Possibly could but I personally don't think it will be worthwhile at this stage. I think the issue stems more from the Bare Graph itself rather than how it is handled. While a Bare graph in theory sounds good, you realise that there are some attributes that it wouldn't make sense to go without anyway. I think then it is probably more a review of the bare graph that needs to be done rather instead of handling it properly, but I'm not sure that is worth a ticket at present

antares1470:
> Should we log a new ticket to handle that?

@Auriga2 Possibly could but I personally don't think it will be worthwhile at this stage. I think the issue stems more from the Bare Graph itself rather than how it is handled. While a Bare graph in theory sounds good, you realise that there are some attributes that it wouldn't make sense to go without anyway. I think then it is probably more a review of the bare graph that needs to be done rather instead of handling it properly, but I'm not sure that is worth a ticket at present

Issue #1261, opened 2021-08-23
Right click menu for buttons down the bottom right


CrucisGamma:
Would be nice to have a right click menu a bit like the following for the Nodes, Links, Edges and Transactions buttons on the bottom right of the screen.  Something like 'Select all' or 'Unselect all' or 'Zoom to Selection' or similar (i.e. no new functionality, just adding a right click option to them with existing operations).

![right_click_menu](https://user-images.githubusercontent.com/62917181/130386138-c7b1c965-c82b-4ce3-b603-02ae132766c5.png)

sol695510:
Tested, changes appear as intended.

PR #1508, opened 2021-10-20
Add icons to right click popup context menu

CrucisGamma:
### Prerequisites

### Description of the Change

Adds in the Selection menu (with icons) on right clicking the Graph menu and adds in the Reset menu (with icons) as well
image

### Alternate Designs

Nil.  The ordering of the selection and reset view sub menus are the same as the default menu options

Already there

### Benefits

Enhances UX for the user

### Possible Drawbacks

Nil

### Verification Process

Ran across multiple graphs with no issues or bugs noticed.

### Applicable Issues

#1261 I see this as addressing the actual problem that that issue was looking at.

### DIFF
--- a/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/zoom/ResetViewContextMenu.java
+++ b/CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/plugins/zoom/ResetViewContextMenu.java
@@ -22,9 +22,12 @@ import au.gov.asd.tac.constellation.graph.interaction.InteractiveGraphPluginRegi
 import au.gov.asd.tac.constellation.graph.visual.contextmenu.ContextMenuProvider;
 import au.gov.asd.tac.constellation.plugins.PluginExecution;
 import au.gov.asd.tac.constellation.utilities.graphics.Vector3f;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import javax.swing.ImageIcon;
+import org.openide.util.ImageUtilities;
 import org.openide.util.lookup.ServiceProvider;
 
 /**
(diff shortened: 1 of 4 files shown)


CrucisGamma:
> If you could add a whatsnew entry for this, that would be great. Some unit tests for the functions you added would be nice as well

The context menus aren't new though, so not sure if they warrant having a what's "new" entry?  Will get right onto adding unit tests for the two functions that return a list of icons.

antares1470:
> The context menus aren't new though, so not sure if they warrant having a what's "new" entry? Will get right onto adding unit tests for the two functions that return a list of icons.

The context menus aren't new that's true, but your additions to said context menus are. The purpose of the whats new is to inform users of changes made that will be of interest to them and I think they would be interested in this

antares1470:
> Hi @antares1470 I don't believe it's necessary to do that (just following the checklist here). Users won't notice that it's changed as it's simply the way that it should have been done earlier and, as I've mentioned before, the context menus aren't new.
> 
>     * [ ]  Update [what's new] if you want to inform users of your change. << My reading on this is that 'you' is the creator of the pull request
> 
> 
> If we change the checklist to make it mandatory, or to add conditions on when it is mandatory, will be happy to update the what's new.

@CrucisGamma sorry for not getting back sooner. I still think it would be good to add a whatsnew entry for reasons you actually outlined yourself. 

"Users won't notice that it's changed as it's simply the way that it should have been done earlier and, as I've mentioned before, the context menus aren't new." -> let me break this down from my perspective
- If you don't inform users of a change, they won't notice a change unless it impacts their current workflows (in this case, people who already use the context menus regularly). The people who don't currently use the context menus regularly might want to know so that they can start using it and the people who do regularly use it might want to know that changes they see are intentional (and so they shouldn't freak out)
- Just because it's the way it should have been done earlier doesn't mean we shouldn't inform users that a change has been made. Whether or not it's a change that should have been done earlier, it's a change nonetheless and so we should let users know if it is something of interest to them. As an aside, I would also be careful about assuming what is the way something "should've been done". I would guess not everyone will use Constellation the same way and so different users will have different ideas for how something "should be done"
- The context menus aren't new that is correct. I'm not asking for a whatsnew entry for the context menus themselves though but merely what you've added to them

The idea of the wording in the checklist was really about distinguishing between things users care about (e.g. new features, improvements) and things developers care about (e.g. unit tests, API changes), as appose to "deciding whether or not I want to tell users about this thing". Perhaps the wording could change from "want to" to "should" but wouldn't need much more than that.

CrucisGamma:
Happy for this PR to be closed with no further consideration given to it.  
I don't believe I will be able to contribute on Constellation in the foreseeable future given the changes to WFH arrangements.

Issue #1488, opened 2021-10-13
Exception while loading Test Parameters preferences 


mimosa2:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Exception while loading Test Parameters preferences 

### Steps to Reproduce

1.  Open data access view and select few parameters from Test parameters plugin

2.  Select Workflow Options -> Save Templates and save the file

3.  Right click and  Select Deactivate all plugins to clear the selections. 

4.  Select Workflow Options -> Load Templates and open the file in step 2

### Behaviour

**Expected behaviour:**  The selected parameters should be loaded w/o exception

**Actual behaviour:**  The exception thrown. But the parameters are loaded except Graph Element Type.

This is from current master
![Testparameters_error2](https://user-images.githubusercontent.com/54261086/137048100-00af4b62-a904-4450-8a1b-649833b549e7.PNG)
This is from Version 2.4
![Testparameters_error](https://user-images.githubusercontent.com/54261086/137048221-e210f1ee-204f-4e4e-9b25-0bfc1c466701.PNG)

**Reproduces how often:** 100%

### Additional Information
![Testparameters_error](https://user-images.githubusercontent.com/54261086/137049330-cda530ed-b012-4bc2-aaca-084368eddd76.PNG)
![Testparameters_error2](https://user-images.githubusercontent.com/54261086/137049373-2e556b4e-be2b-4b84-a868-9926c8fb9289.PNG)

sol695510:
Tested by following the steps to reproduce and received the expected behaviour, with no exceptions, with 2 different selections for templates.

PR #1510, opened 2021-10-20
Fix the NoSuchMethodException when loading certain Templates in the DAV

Auriga2:
### Prerequisites

### Description of the Change
Fixed the `NoSuchMethodException` when loading a Template with any `Graph element type` in the `Test Parameters` plugin in the Data Access View.

Bug Fix

### Benefits
No exceptions are thrown

### Verification Process
Open data access view and select few parameters from Test parameters plugin including `Graph Element Type`

Select Workflow Options -> Save Templates and save the file

Right click and Select Deactivate all plugins to clear the selections.

Select Workflow Options -> Load Templates and open the file in step 2 and verify no exceptions are thrown, and all the fields are populated including `Graph Element Type`

### Applicable Issues

https://github.com/constellation-app/constellation/issues/1488


Auriga2:
None of the constructors available are matching because the GraphElementTypeParameterValue is an inner class and the outer class TestParametersPlugin is also included as a constructor parameter by default in each constructor.
image

Auriga2:
@antares1470  Fixed the test parameters plugin by making the `GraphElementTypeParameterValue` a static nested class.
Also added a comment in SingleChoiceParameterValue to avoid future issues. We might get the same issue with MultiChoiceParameterValues. I'll add a comment there as well if you think is required.

antares1470:
> @antares1470 Fixed the test parameters plugin by making the `GraphElementTypeParameterValue` a static nested class. Also added a comment in SingleChoiceParameterValue to avoid future issues. We might get the same issue with MultiChoiceParameterValues. I'll add a comment there as well if you think is required.

@Auriga2 I wouldn't say it is "required" might if you reckon there will be a similar issue for multi choice parameters then I think it would be nice to include a similar comment there as well

antares1470 (approved):
You could add to the multi choice parameters as you mentioned but looks good otherwise

antares1470 (changes requested):
Sorry I overlooked something, it appears there are some unit tests failing. If you're able to check those out, that would be great

antares1470 (approved):

antares1470 (approved):

arcturus2 (approved):

Issue #1340, opened 2021-09-02
Select Backbone also selecting nodes only connected by a loop


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Select Backbone should only select nodes with more than 1 neighbour and transactions that connect 2 backbone nodes. This should exclude nodes only connected by loops. At present though, it appears that nodes with loops (and the loops themselves) are being selected, even though they wouldn't form part of the backbone otherwise. This should be fixed.

### Steps to Reproduce

1.  Construct the graph below but ensure everything is deselected
2.  Select 'Select Backbone'

**Expected behaviour:** Vertex \#0 is selected

**Actual behaviour:** Vertex \#0, Vertex \#2, and Vertex\#3 are selected

**Reproduces how often:** 100%

### Additional Information

This is the graph I used to test it (showing the result after running Select Backbone with nothing selected prior):

![SelectBackboneTest](https://user-images.githubusercontent.com/62311243/131787906-c575d327-46ef-48c2-beb8-06632a83b5e0.png)

Atlas139mkm:
Just gave this a test and looks like its working as intended. Just vertex0 is being selected.

PR #1514, opened 2021-10-20
fix backbone plugin treating self-loops as links

Guilty-Spark-343:
### Prerequisites

### Description of the Change

Remove self-loops when calculating the backbone.  In the current implementation nodes that have self nodes are included in the backbone without any external links as observed in #1340.  The new implementation simply checks that their neighbors aren't self-loops.  

Before:
image

After:
image

### Alternate Designs

None considered.

This behavior is more in line with user expectations and doesn't allow singletons in the backbone.

### Benefits

It is a more accurate representation of the backbone.  

### Possible Drawbacks

The additional check to identify the neighbors aren't self-loops has some overhead.  The change introduced an average of 6.4% delay to computing the backbone when computing on a small world generated graph with the following:
Nodes: 10000
Nearest Neighbors to attach: 2
Rewiring probability: 0.5
Link: 14141
Edges: 38328
Transactions: 369411

### Verification Process

- Create a new graph
- Add some nodes that have self loops as well as nodes that have more than 1 link (not including self-loops)
- Run the backbone plugin
- The nodes that have more than 1 link should be selected and the nodes that just have self loops shouldn't be selected on the graph.

### Applicable Issues

#1340


Atlas139mkm:
Just gave this a test and looks like its working as intended. Just vertex0 is being selected.

Issue #882, opened 2020-10-29
Change calls to getActiveGraph within plugins


aldebaran30701:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
A need has arisen from multiple bugs and undefined behaviour from using getActiveGraph within the executed plugins. This is to be solved by actually running the plugin on only one graph. There is a structure available that you can get the graph you want via the graphmanager.getdefault() manager. This will require the graph id to fetch the graph object.

Fixing this fundamental issue will alleviate the below issue and also ones relating to long data access view plugins running on different graphs. 

Below is the original issue.
------------------------------------------------------------------------------------------------
If you generate a graph while also waiting for another, it will render both onto one graph.
I'd say due to how hard it is to reproduce during normal operation that it is not a high priority, but has the potential to occur possibly even with importing data into a graph.

I followed the steps below quickly.
open new graph -> generate structured graph builder (1000) backbone size
open new graph -> generate sphere graph (10000)
open new graph -> generate sphere graph (1000)
While it was all trying to load, I switched between graphs and while it happened to try and load the sphere, I was on the structured graph. It loaded those elements onto that graph instead of the sphere graph.
See the below gif for the behavior

**Expected behaviour:** Sphere graph to display on graph 2

**Actual behaviour:** Sphere graph generates all on 0,,0,0 in graph 1.

**Reproduces how often:** 100%
### Additional Information
N/A

antares1470:
My gut would say (without looking at the code) that there may be a call to `getActiveGraph()` that might be resulting in this behaviour

aldebaran30701:
Tested, seems to be no issues from the #1524 PR. 
This ticket could include a closer look at things like the chaining of plugins. Where the `.getActiveGraph()` may be used for things like running arrangements after plugins are executed.

PR #1524, opened 2021-10-27
Fixing active graph calls

antares1470:
### Prerequisites

### Description of the Change

Some of issues have been identified previously where unexpected behaviour could occur if you have multiple graphs open and switch graphs before a process ends and this is partially due to functions applied to the active graph. This PR starts to take a look at some of the `GraphManager.getDefault().getActiveGraph()` calls scattered throughout the code and see if there is a way to better consolidate them to help better ensure that plugins and other behaviours are applied to the intended graph. 

Firstly I consolidated some calls where a function called `getActiveGraph()` more than once into one call stored in a variable. In these instances, it would seem it was intended that they were all run on the same graph (being the active graph at the time) and so this minimises the chances of the graph changing mid-function.

Secondly I removed some calls that were not needed at all. In these instances, there were graph parameters that already were (or could be) being passed in which were really the intended targets (and would generally be the active graph in normal use anyway).

In total, the number of instances of `GraphManager.getDefault().getActiveGraph()` in Core goes down from 125 to 103.

Starts to address the issue of ensuring behaviours are always applied to the intended graph.

### Benefits

More likely that plugin behaviours only applied to intended graph

### Possible Drawbacks

There might be some, can't think of any though

### Verification Process

Ran affected plugins before and after changes to ensure that they still worked as intended

### Applicable Issues

#882


antares1470:
@aldebaran30701 as requested I've added some unit tests. Rather than doing the whole classes though, I've only focused on adding tests for the major deltas of the PR

Issue #1518, opened 2021-10-22
Enhanced functionality in the New Attribute dialog box


CrucisGamma:
As a user, I would like to be able to press enter in either the Label or in the Description and have 'Enter' be the result.  However, if there is no characters in the Label edit text box, then OK should be inactivated.

As a user, if I type in 'i' if I've tabbed to the Type dropdown, it should display the first of the 'i' types (aka atm that's 'icon').  It correctly scrolls up and down if I then use the arrow keys.  

There's also behaviour in other products that if you type in say 'i' then 'n' then 't' in will bring up the alphabetical first of types starting with 'int' (atm this is 'integer') but if you leave a pause then it would go to the first of 'i' types, then the alphabetical first of the 'n' types and then the first of the 't' types. 

![image](https://user-images.githubusercontent.com/62917181/138393395-01832a9d-01b4-4716-a76e-56d65c585e40.png)

capricornunicorn123:
The type-ahead / autofill functionality has been addressed in the #2025

PR #1532, opened 2021-10-29
Changes to enhance UX for OK button in user facing dialog boxes

CrucisGamma:
### Prerequisites

### Description of the Change

Allows the user to control whether or not the OK button is set to 'setDefaultButton' for dialog boxes.  This will allow the user to press enter on relevant dialogs to submit whatever it is they've typed in.  It's controlled through a new Constellation preferences option.

### Alternate Designs

The existing option of not having a default button is super slow as you have to navigate manually to it with the mouse button each time.  Having the default button always on was a change that may not go over well with existing users of constellation.  Having the default for the new option be set to 'False' makes this an opt in feature so only users who want it on will actually have it enabled.

Adds to the quickness of interacting with, for example, adding in new attributes for importing data (trying to do it the existing way with say 20 fields is super slow).

### Benefits

Enhanced UX for the user with complete control given to the user.

### Possible Drawbacks

Nil.

### Verification Process

Reviewing all the code for any possible OK button, navigating directly to having those come up in the Constellation client. 

### Applicable Issues

#1518

### DIFF
--- a/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/ConversationBox.java
+++ b/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/ConversationBox.java
@@ -123,13 +123,17 @@ public final class ConversationBox extends StackPane {
     private volatile boolean isAdjustingSenderLabels;
 
     private int foundCount;
-    private static final String FOUND_TEXT = "Found: ";
+    private int searchCount;
+    
+    private static final String FOUND_TEXT = "Showing ";
     private static final String FOUND_PASS_COLOUR = "-fx-text-fill: yellow;";
     private static final String FOUND_FAIL_COLOUR = "-fx-text-fill: red;";
     private final Label foundLabel = new Label();
 
     private final TextField searchTextField = new TextField();
-    private final VBox searchVBox = new VBox();
+    private final Button nextButton = new Button("Next");
+    private final Button prevButton = new Button("Previous");
+    private final HBox searchHBox = new HBox();
 
     /**
      * Create a ConversationBox with the given Conversation.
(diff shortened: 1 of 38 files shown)


antares1470 (changes requested):
I had to have a hard think about this. My gut said that this seems like a really odd thing to have as a preference that users can set so I don't think this is the way to go.

What I reckon may be better (and still achieve what you're after) is simply to make a call on which dialogs we should or shouldn't have the ok button highlighted by default (we have similar options for the swing dialogs). As a rule of thumb, I would suggest any dialog where you might want to use the enter button for any other reason besides submitting (e.g. newline character) would be one we wouldn't highlight ok by default for and highlight by default otherwise

arcturus2:
@antares1470 so I am guessing that @CrucisGamma made it a preference to cater for users that didn't want this feature but I agree with you that making a sensible jugdement call for which ones should be default OK on enter will be fine and a natural experience.

CrucisGamma:
There was the specific point made in discussions at the time that "some users would not like to see them changed".  I've not seen the outcomes of any survey done in this matter but took it at face value that that was the case.  
Happy for this PR to be closed with no further consideration given to it.  
I don't believe I will be able to contribute on Constellation in the foreseeable future given the changes to WFH arrangements.
Noting, however, that with adding in 20 extra columns for a single input in JSON took around 3 minutes with the change in here and around double that without (and with quite a bit more frustration).

aldebaran30701:
Just my 2c, this seems like a neat idea. I like the idea of no change for current users but the power users will be able to just hit enter through the dialogs they are familiar with. If this was added, the other review comments could be addressed by actually setting sensible defaults to dialogs. The power user preference would override this default and always default to OK.

PR #1536, opened 2021-11-02
DAV template structure update

antares1470:
### Prerequisites

### Description of the Change

There was an issue with current structure of the Data Access View templates, specifically with the "plugins" section. When a template is created, all the plugin parameters and their values are added to the same map (which in turn populates the plugins section). The problem here though is that if multiple plugins being added to a template use the same parameter (different instantiations but still have the same id), those different plugins will override the parameter value stored in the map. This results in those plugins then having the same value for that parameter when the template is loaded (especially frustrating for multi-choice parameter which would override not just the choices but the options to pick from as well).

The change in this PR alters the structure of the plugins section by adding sub-sections for each plugin. This way, the parameters being stored are added with respect to the plugin and there is no inherent assumption that a parameter being added is only applicable to one plugin.

### Alternate Designs

I looked at options to keep the existing structure alongside this new structure (for reasons explained in the drawbacks section) but this was going to make the way more complicated and confusing then it needed to be.

This future-proofs the data access view template should parameters be re-used amongst multiple data access view plugins.

### Benefits

Future-proofing, less assumptions being made about how parameters are used

### Possible Drawbacks

The major drawback with this change is that this will mean any existing DAV templates will no longer be usable and will need to be re-created. This might be a little annoying for any existing users of those templates but as explained in the alternate designs, any efforts to try and support both structures was going to make code unnecessarily complicated. Coming up with a template structure that complimented the existing structure while also solving the issue outlined in the description also proved futile.

### Verification Process

1) Open Constellation in master
2) Open Data Access View and edit some parameter values for both the Test Parameters and Test Parameters (Batched) plugin (ensure at least one parameter has a different value for each plugin)
3) Ensure both plugins are enabled and save the template
4) Close the tab/step and then open a new one (or some other way to reset/ensure you can see the template applied)
5) Load the template you just created
6) Observe the values applied to both are the same
7) Run steps 2-5 in this branch and verify the expected values are loaded for each plugin


Delphinus8821:
Followed the steps provided for testing. Each time I tried to load the template I got this error. 

DAVTemplateException

antares1470:
> 
> 
> Followed the steps provided for testing. Each time I tried to load the template I got this error.
> 
> DAVTemplateException
> 
> Noted this should be fixed with #1510

Yes as you identify, that is a separate issue that is being fixed. That issue aside though, hopefully it works.

Delphinus8821 (approved):
Aside from the exception mentioned prior, seems to work well.

aldebaran30701 (approved):

PR #1541, opened 2021-11-04
Fix to Analytic View sort

mimosa2:
### Prerequisites

### Description of the Change

After the recent Analytic View refactor the default sort was disabled and the results were displayed in random order. This fix brings back the default descending order of the results. 

Core functionality

### Benefits

The descending order is more intuitive to analyze the data. The random ordering was misleading   

### Possible Drawbacks

Possible performance issues tested, but there is no change in performance.  

### Verification Process

1. Open the Analytic View and open a graph.
2. Select any Analytic Question e.g. 'Connects The Network Best' -> 'Betweenness Centrality Alalytic' and Run the plugin.
3. Notice the results are sorted in descending order.

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/results/AnalyticResult.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/results/AnalyticResult.java
@@ -20,6 +20,8 @@ import au.gov.asd.tac.constellation.views.analyticview.AnalyticViewTopComponent.
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/results/AnalyticResult.java
```diff
@@ -78,6 +79,35 @@ public final int size() {
     }
 
     public final void sort() {
+        result = sortMapByValuesWithDuplicates(result);
+    }
+
+    private LinkedHashMap sortMapByValuesWithDuplicates(Map<IdentificationData, D> passedMap) {
```
antares1470, line 85: I'm wondering whether of all this can be a void function which is simply applied directly to the `result` map? That would also allow `result` to remain final. That would probably mean the following changes to your function to allow that to happen:
- replace calls of `passedMap` with `result`
- `sortedMap` declaration just becomes `result.clear()`
- adding to `sortedMap` would then instead add to `result`
- your key set map should then become an entry set map to allow the comparison to occur

antares1470 (changes requested):
If you're able to add finals, as well as add paramatised types to the collections (rather than use the raw type), that would be great. I think a unit test for the sort function would also be good

Delphinus8821 (approved):

arcturus2 (approved):

antares1470 (approved):

PR #1558, opened 2021-11-16
Unit tests for JMultiChoiceComboBoxMenu class

Auriga2:
### Prerequisites

### Description of the Change
Unit tests for `JMultiChoiceComboBoxMenu` class

### Verification Process
All tests pass

### Applicable Issues
#413

### DIFF
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -19,6 +19,7 @@ import java.awt.Point;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.net.MalformedURLException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -229,6 +229,11 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
         button.setToolTipText(text);
     }
 
+    @Override
+    public final String getToolTipText() {
+        return button.getToolTipText();
+    }
+
     public static class OpenAction implements ActionListener {
 
         private final JPopupMenu menu;
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -224,6 +224,10 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
         });
     }
 
+    public Set<ListSelectionListener> getListeners() {
+        return listeners;
+    }
+
     @Override
     public void setToolTipText(final String text) {
         button.setToolTipText(text);
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -235,7 +235,7 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
     }
 
     protected Set<ListSelectionListener> getListeners() {
-        return listeners;
+        return Collections.unmodifiableSet(listeners);
     }
 
     protected JButton getButton() {
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -89,16 +89,7 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
             this.button = new JButton(text, arrow);
         }
         button.setHorizontalTextPosition(SwingConstants.LEFT);
-        button.addActionListener(event -> {
-            if (!menu.isVisible()) {
-                final Point p = button.getLocationOnScreen();
-                menu.setInvoker(button);
-                menu.setLocation((int) p.getX(), (int) p.getY() + button.getHeight());
-                menu.setVisible(true);
-            } else {
-                menu.setVisible(false);
-            }
-        });
+        addButtonActionListener(button, menu);
         add(button);
 
         selectedItems = new HashSet<>();
@@ -123,6 +114,19 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
             });
             menuItems.put(menuItem, item);
             menu.add(menuItem);
+        });
+    }
+
+    protected final void addButtonActionListener(final JButton button, final JPopupMenu menu) {
+        button.addActionListener(event -> {
+            if (!menu.isVisible()) {
+                final Point p = button.getLocationOnScreen();
+                menu.setInvoker(button);
+                menu.setLocation((int) p.getX(), (int) p.getY() + button.getHeight());
+                menu.setVisible(true);
+            } else {
+                menu.setVisible(false);
+            }
         });
     }
 
@@ -241,10 +245,6 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
 
     protected JButton getButton() {
         return button;
-    }
-
-    protected void setButton(JButton button) {
-        this.button = button;
     }
 
     protected JPopupMenu getMenu() {
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -50,7 +50,7 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
 
     private static final Logger LOGGER = Logger.getLogger(JMultiChoiceComboBoxMenu.class.getName());
 
-    private final JButton button;
+    private JButton button;
     private final JPopupMenu menu;
     private final Map<JMenuItem, E> menuItems;
     private final Set<E> selectedItems;
@@ -224,10 +224,6 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
         });
     }
 
-    public Set<ListSelectionListener> getListeners() {
-        return listeners;
-    }
-
     @Override
     public void setToolTipText(final String text) {
         button.setToolTipText(text);
@@ -236,6 +232,22 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
     @Override
     public final String getToolTipText() {
         return button.getToolTipText();
+    }
+
+    protected Set<ListSelectionListener> getListeners() {
+        return listeners;
+    }
+
+    protected JButton getButton() {
+        return button;
+    }
+
+    protected void setButton(JButton button) {
+        this.button = button;
+    }
+
+    protected JPopupMenu getMenu() {
+        return menu;
     }
 
     public static class OpenAction implements ActionListener {
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
@@ -98,25 +98,26 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
         items.forEach(item -> {
             final JMenuItem menuItem = new JCheckBoxMenuItem(item.toString());
             menuItem.addActionListener(new OpenAction(menu, button));
-            menuItem.addActionListener(event -> {
-                final JMenuItem changedMenuItem = (JMenuItem) event.getSource();
-                final E changedItem = menuItems.get(changedMenuItem);
-                if (selectedItems.contains(changedItem)) {
-                    removeSelectedItem(changedItem);
-                } else {
-                    addSelectedItem(changedItem);
-                }
-                listeners.forEach(listener -> {
-                    final int menuIndex = menu.getComponentIndex(changedMenuItem);
-                    final ListSelectionEvent selectionEvent = new ListSelectionEvent(changedItem, menuIndex, menuIndex, false);
-                    listener.valueChanged(selectionEvent);
-                });
-            });
+            addMenuItemActionListener(menuItem);
             menuItems.put(menuItem, item);
             menu.add(menuItem);
         });
     }
 
+    protected final void addMenuItemActionListener(final JMenuItem menuItem) {
+        menuItem.addActionListener(event -> {
+            final JMenuItem changedMenuItem = (JMenuItem) event.getSource();
+            final E changedItem = menuItems.get(changedMenuItem);
+            if (selectedItems.contains(changedItem)) {
+                removeSelectedItem(changedItem);
+            } else {
+                addSelectedItem(changedItem);
+            }
+            final int menuIndex = menu.getComponentIndex(changedMenuItem);
+            final ListSelectionEvent selectionEvent = new ListSelectionEvent(changedItem, menuIndex, menuIndex, false);
+            valueChanged(selectionEvent);
+        });
+    }
     protected final void addButtonActionListener(final JButton button, final JPopupMenu menu) {
         button.addActionListener(event -> {
             if (!menu.isVisible()) {
@@ -148,6 +149,10 @@ public class JMultiChoiceComboBoxMenu<E> extends JComponent implements ListSelec
 
     public final Set<E> getItems() {
         return new HashSet<>(menuItems.values());
+    }
+
+    protected Map<JMenuItem, E> getMenuItems() {
+        return menuItems;
     }
 
     public final Set<E> getSelectedItems() {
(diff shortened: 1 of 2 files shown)


CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+
+    /**
+     * Test of setSelectedItem method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testSetSelectedItem() {
+        System.out.println("setSelectedItem");
+        String item = "Item 3";
+        JMultiChoiceComboBoxMenu instance = mock(JMultiChoiceComboBoxMenu.class);
+        instance.setSelectedItem(item);
```
aldebaran30701, line 215: Can you get the selection model and then check the selected item is the same?

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+     * Test of setSelectedItems method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testSetSelectedItems() {
+        System.out.println("setSelectedItems");
+        String s1 = "Item 1";
+        String s2 = "Item 3";
+        String s3 = "Item 4";
+        JMultiChoiceComboBoxMenu instance = mock(JMultiChoiceComboBoxMenu.class);
+        instance.setSelectedItems(s1, s2, s3);
```
aldebaran30701, line 229: same here

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Test of addSelectionListener method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testAddSelectionListener() {
+        System.out.println("addSelectionListener");
+        ListSelectionListener listSelectionListenerMock = mock(ListSelectionListener.class);
+        JMultiChoiceComboBoxMenu instance = mock(JMultiChoiceComboBoxMenu.class);
+        instance.addSelectionListener(listSelectionListenerMock);
+        verify(instance, times(1)).addSelectionListener(listSelectionListenerMock);
```
aldebaran30701, line 341: can you retrieve the selection listener to check it at all?
It seems like the test really verifies that the above call was executed and would probably always pass

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Test of valueChanged method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testValueChanged() {
+        System.out.println("valueChanged");
+        JMultiChoiceComboBoxMenu instance = mock(JMultiChoiceComboBoxMenu.class);
+        ListSelectionEvent mockEvent = mock(ListSelectionEvent.class);
+        instance.valueChanged(mockEvent);
+        verify(instance, times(1)).valueChanged(mockEvent);
```
aldebaran30701, line 353: possibly same as above here

aldebaran30701 (changes requested):
Just a few comments as below

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    @Test
+    public void testAddSelectedItems() {
+        System.out.println("addSelectedItems");
+        String s1 = "Item 1";
+        String s2 = "Item 3";
+        String s3 = "Item 4";
+        String[] selectedItems = {s1, s2, s3};
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+        instance.addSelectedItems(s1, s2, s3);
+        Set resultSet = instance.getSelectedItems();
```
antares1470, line 259: Are you able to add the parameterized type for the set rather than using the raw type?

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Test of getSelectedItems method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testGetSelectedItems() {
+        System.out.println("getSelectedItems");
+        String item = "Item 3";
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+        instance.setSelectedItem(item);
+        Set resultSet = instance.getSelectedItems();
```
antares1470, line 202: same here

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * Test of getItems method, of class JMultiChoiceComboBoxMenu.
+     */
+    @Test
+    public void testGetItems() {
+        System.out.println("getItems");
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+        Set result = instance.getItems();
```
antares1470, line 189: and here

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+    public void testRemoveSelectedItem() {
+        System.out.println("removeSelectedItem");
+        String s1 = "Item 1";
+        String s2 = "Item 3";
+        String s3 = "Item 4";
+        String[] selectedItems = {s1, s2, s3};
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+
+        instance.addSelectedItems(s1, s2, s3);
+        Set itemsbeforeRemove = instance.getSelectedItems();
```
antares1470, line 277: ditto

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+        assertEquals(itemsbeforeRemove.size(), 3);
+        assertEqualsNoOrder(itemsbeforeRemove.toArray(), selectedItems);
+
+        instance.removeSelectedItem(s2);
+        Set resultSet = instance.getSelectedItems();
```
antares1470, line 282: ditto

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+        String[] selectedItems = {s1, s2, s3};
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+
+        instance.addSelectedItems(s1, s2, s3);
+        Set itemsbeforeRemove = instance.getSelectedItems();
+        assertEquals(itemsbeforeRemove.size(), 3);
+        assertEqualsNoOrder(itemsbeforeRemove.toArray(), selectedItems);
+
+        instance.removeSelectedItems(s2, s3);
+        Set resultSet = instance.getSelectedItems();
```
antares1470, line 305: ditto

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -0,0 +1,369 @@
...
+        String[] selectedItems = {s1, s2, s3};
+        JMultiChoiceComboBoxMenu instance = new JMultiChoiceComboBoxMenu("Text", items);
+
+        instance.addSelectedItems(s1, s2, s3);
+        Set itemsbeforeRemove = instance.getSelectedItems();
+        assertEquals(itemsbeforeRemove.size(), 3);
+        assertEqualsNoOrder(itemsbeforeRemove.toArray(), selectedItems);
+
+        instance.clearSelection();
+        Set resultSet = instance.getSelectedItems();
```
antares1470, line 328: ditto

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
antares1470: Agree with @aldebaran30701. It seems in the current state this is simply testing the function executed rather than checking it executed as intended

antares1470 (changes requested):

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: I guess Verify is still a valid way of testing a void method. All I can think of is changing the modifier of `Set<ListSelectionListener> listeners` to protected or adding a getter. Not sure it's worth the effort because this method just adds the listener in the collection.

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: So how would you verify the value is actually changed here?

aldebaran30701: I'd assign some listener mocks to the instance's `listeners` then call `valueChanged()`. Then verify that the method iterated over all listeners and called `valueChanged()` on them.

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
aldebaran30701: Usually if it's one line I wouldn't say to worry, but considering the test case is already set up, it wouldn't hurt to change it to protected given it's useful to test

antares1470: @Auriga2 verify can be used to check calls made within a function you're testing but shouldn't really be used for the function itself. Otherwise it doesn't actually matter what the function implementation is (e.g. a blank implementation would pass this test, even though that's definitely a bug). 

If you were to check this, I would add a getter rather than change the modifier in this instance. If you feel the method is too trivial to check though, it may be worth omitting this unit test all together. Worth noting though that you don't appear to use the function for any of the other unit tests so I think it may still be worth having the basic test in this instance

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: Because this is already used in other methods, sonar now complaints of a 1.3% Duplication. Shall I remove this method?

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: Because this is already used in another method, sonar now complaints of a 1.3% Duplication. Shall I remove this method?

Auriga2: Above comment not relevant here, so I'll delete

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: Thanks Aldebaran. Fixed now.

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: Thanks, it was my misunderstanding. Fixed now.

aldebaran30701 (approved):

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
```diff
@@ -224,6 +224,10 @@ public final void valueChanged(final ListSelectionEvent event) {
         });
     }
 
+    public Set<ListSelectionListener> getListeners() {
```
antares1470, line 227: are you able to make this protected (since this was only added for unit test purposes)

antares1470 (changes requested):

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
Auriga2: Fixed

Auriga2:
Also added 2 new tests to cover missing parts. 

I tried but couldn't find a way to test the `menuItem AddActionListener` because of the 2 ActionListeners and not sure how to mock the `OpenAction` in `menuItem.addActionListener(new OpenAction(menu, button));`

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
```diff
@@ -238,6 +234,22 @@ public final String getToolTipText() {
         return button.getToolTipText();
     }
 
+    protected Set<ListSelectionListener> getListeners() {
+        return listeners;
```
antares1470, line 238: small but can you make this return `Collections.unmodifiableSet(listeners)`

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenu.java
```diff
@@ -238,6 +234,22 @@ public final String getToolTipText() {
...
+    }
+
+    protected JButton getButton() {
+        return button;
+    }
+
+    protected void setButton(JButton button) {
+        this.button = button;
+    }
```
antares1470, line 247: is there a way your unit tests can be written such that this setter isn't needed? I tend to be ok with getters being added for unit testing but a little more reluctant to add setters if unit testing is their only use. Especially seeing as in this instance you also had to make `button` non-final in the process

Auriga2: I couldn't find a way around this. The issue is because this code is in the constructor and the mock on the `JButton` doesn't apply unless you set it again. Do you have other ideas?

antares1470: given the button is created within the constructor, could you just use the getter to get the button and then call something like: `button.getActionListeners()[0].actionPerformed(null);`? (the parameter for actionPerformed isn't used in this case so you can pass null ok)

Auriga2: The problem is `button.getLocationOnScreen()` which  relies on it being shown on the screen unless we mock. This is the error:
image

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
```diff
@@ -181,6 +185,72 @@ public void testaddButtonActionListener() {
...
+        assertEquals(resultSet.size(), 2);
+    }
+
+    public <K, V> K getKey(Map<K, V> map, V value) {
+        for (Entry<K, V> entry : map.entrySet()) {
+            if (entry.getValue().equals(value)) {
+                return entry.getKey();
+            }
+        }
+        return null;
```
antares1470, line 252: I would make this private since this won't be used outside the test class. 

Probably worth noting though that while it may be ok for the test class, this function does seem to somewhat assume that values are unique (which isn't the case with maps) which means in the general case it could return a different result each time depending on how the map is iterated through

antares1470 (changes requested):

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/gui/JMultiChoiceComboBoxMenuNGTest.java
Auriga2: Yes agree, added the support of multiple values for clarity, although the test cases this function is used only contain unique values.

antares1470 (approved):

PR #1560, opened 2021-11-17
Unit tests for IconData and UriIconData

sol695510:
### Prerequisites

### Description of the Change
Added unit tests for Icon Data and UriIconData. A setData() method was created in IconData exclusively for testing, and the equals() method in IconData was also refactored.

### Alternate Designs
N/A

N/A

### Benefits
Testing.

### Possible Drawbacks
N/A

### Verification Process
Used print statements to verify all logic paths in code were accounted for.

### Applicable Issues
#413


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/icon/IconData.java
```diff
@@ -179,17 +188,15 @@ public int hashCode() {
...
-        } else if (obj == null) {
+        if (obj == null) {
             return false;
-        } else if (getClass() != obj.getClass()) {
+        }
+
+        if (!(obj instanceof IconData)) {
             return false;
-        } else {
-            // Do nothing
```
antares1470, line 189: I'm a little confused by this change. Is this meant to be equivalent?

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,465 @@
...
+        verify(instance2, times(1)).createData(Mockito.eq(size2), Mockito.eq(color2));
+        assertEquals(result2, expResult2);
+
+        // When size is the default (256).
+        final int size3 = 256;
+        final Color color3 = null;
+
+        // When data is null.
+        final IconData instance3 = spy(new IconDataImpl());
+        instance3.setData(null);
```
antares1470, line 117: I think you should only need one of these. That is either creating a new instance every time, or adding the setter to be able to set back to null. Personally I would prefer the former (since I'm guessing there is a reason why there isn't a setter in place judging by the code) but even then I think if you find yourself needing to create more than one instance in a unit test then I would possibly look at splitting the tests up

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,465 @@
...
+
+        final byte[] expResult3 = new byte[0];
+        final byte[] result3 = instance3.getData(size3, color3);
+
+        verify(instance3, times(1)).createData(Mockito.eq(size3), Mockito.eq(color3));
+        assertEquals(result3, expResult3);
+
+        // When data is not null.
+        final IconData instance4 = spy(new IconDataImpl());
+        instance4.setData(new byte[1]);
```
antares1470, line 127: I suspect you probably don't need to create a new instance for this (as any of your other ones will already have data not null)

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/UriIconDataNGTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        uriStaticMock.when(() -> URI.create(Mockito.any(String.class))).thenReturn(uriMock);
+
+        // If the URI is absolute and has a scheme component.
+        when(uriMock.isAbsolute()).thenReturn(true);
+
+        final UriIconData expResult = new UriIconData("");
+        final UriIconData result = new UriIconData("");
+
+        assertEquals(result, expResult);
```
antares1470, line 106: I'm a little confused as to what is being tested here

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/UriIconDataNGTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        // If the URI is not absolute and does not have a scheme component.
+        when(uriMock.isAbsolute()).thenReturn(false);
+
+        try {
+            final UriIconData instance = new UriIconData("");
+            fail("AssertionError was not thrown.");
+        } catch (final AssertionError er) {
+        }
+    }
```
antares1470, line 116: I would separate this into its own unit test (and then have the expected exception as AssertionError, rather than catching it)

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/UriIconDataNGTest.java
```diff
@@ -0,0 +1,224 @@
...
+        assertEquals(result, expResult);
+
+        // If the URI is not absolute and does not have a scheme component.
+        when(uriMock.isAbsolute()).thenReturn(false);
+
+        try {
+            final UriIconData instance = new UriIconData(uriMock);
+            fail("IllegalArgumentException was not thrown.");
+        } catch (final IllegalArgumentException ex) {
+        }
```
antares1470, line 142: Similar

antares1470 (changes requested):

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/icon/IconData.java
sol695510: Its not meant to be equivalent. It first checks if the object to compare is null (yes then false), if its an instance of IconData (no then false), then finally if it wasn't null and is an instance of IconData, it then checks whether the "data" arrays match and returns true/false.

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/UriIconDataNGTest.java
sol695510: It has been removed as the tests just test the case when the exception is thrown on the constructor.

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/icon/IconData.java
antares1470: @sol695510 I think I would keep it as it was (though some of the if conditions that only returned false could probably be merged together). I think `getClass() != obj.getClass()` is a slightly stronger condition compared to `!(obj instanceof IconData)` (the latter only implies the supertype is the same, not the actual type) and you probably don't want to spend time doing extra checks if the 2 objects are the same object (the 1st condition you removed)

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/icon/IconData.java
```diff
@@ -38,6 +38,15 @@
...
     private byte[] data = null;
 
+    /**
+     * Setter used only for testing.
+     *
+     * @param data
+     */
+    protected void setData(final byte[] data) {
+        this.data = data;
+    }
```
antares1470, line 48: is there a way your unit tests can be written such that this setter isn't needed? I tend to be ok with getters being added for unit testing but a little more reluctant to add setters if unit testing is their only use

sol695510: I can't seem to get data be another value without doing so, as data is assigned the value of a bytearray from an outputstream that seems to always be emtpy after the stream is used.

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,540 @@
...
+    public void testGetData_dataNotNull() {
+        System.out.println("testGetData_dataNotNull");
+
+        final IconData instance = spy(new IconDataImpl());
+
+        final int size = ConstellationIcon.DEFAULT_ICON_SIZE;
+        final Color color = null;
+
+        // Data not null.
+        instance.setData(new byte[1]);
```
antares1470, line 175: In this instance, instead of using `setData` you could call `getData` twice and verify that `createData` is only called the 1st time (in which case you could potentially merge with the test above). I don't think the exact size of data should make a difference here

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,540 @@
...
+
+        // When the class of obj is not the same class as this.
+        obj = new Object();
+        assertFalse(instance.equals(obj));
+
+        // When the class of obj is the same class as this but has a different value for data.
+        obj = new IconDataImpl();
+
+        instance.setData(new byte[1]);
+        ((IconData) obj).setData(new byte[2]);
```
antares1470, line 496: Instead of `setData` here, you could possibly do some mocking of `createData` return values (passing default values to `getData` to ensure `data` is set) and then compare

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,540 @@
...
+
+        assertFalse(instance.equals(obj));
+
+        // When the class of obj is the same class as this and has the same value for data.
+        obj = new IconDataImpl();
+
+        instance.setData(new byte[1]);
+        ((IconData) obj).setData(new byte[1]);
```
antares1470, line 504: If you use the default value here from `getData` instead of using `setData`

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/icon/IconDataNGTest.java
```diff
@@ -0,0 +1,540 @@
...
+
+        final IconData instance = new IconDataImpl();
+
+        final String expResult1 = String.format("Image of %d bytes", 0);
+        final String result1 = instance.toString();
+
+        assertEquals(result1, expResult1);
+
+        // When data is not null.
+        instance.setData(new byte[2]);
```
antares1470, line 524: could use `getData` with default values and if you don't want size 0 you could mock the value from `createData`

antares1470 (commented):
@sol695510 here's some ideas I thought of for alternatives to `setData`. They may not always look as clean but they should remove the need to create the setter

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/icon/IconData.java
antares1470: @sol695510 see my below comments. I've added some thoughts on alternatives

antares1470 (approved):

arcturus2 (approved):

PR #1566, opened 2021-11-19
🧹  Sonar smells

aldebaran30701:
### Prerequisites

### Description of the Change
**Worked on the following rules:**
Add a private constructor to hide implicit public
Standard outputs shouldn't be used
Unused private fields should be removed
Rename fields to match regular expression
Constructors of abstract classes should be protected

### Applicable Issues
#1022


CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/infomap/flow/Network.java
```diff
@@ -117,13 +117,11 @@ public void read() {
 
         Logf.printf("done! Found %d nodes and %d connections. ", rg.getVertexCount(), connectionMap.size());
         if (numDoubleLinks > 0) {
-            Logf.printf("%d connections was aggregated to existing connections. ", numDoubleLinks);
+            Logf.printf("%d connections was aggregated to existing connections. %n", numDoubleLinks);
```
antares1470, line 120: Just noting you changed from `Logf` to `LOGGER` in a previous file but here you haven't. Not sure if this needs to change to using the java logger?

aldebaran30701: I'll change those ones back to logF. I think it was early on in the branch and decided to go against changing them. I switched logf to use the logger either way.

antares1470: It looks like you've changed the `Logf` to use the java logger over sysout so I guess that makes sense. That being said, I think that class is only created to be a wrapper around sysout so potentially that class could be removed all together and then you can just use the logger of the class that function was being called in

Delphinus8821 (approved):

CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/infomap/flow/Network.java
aldebaran30701: Removed all logF usages and replaced with logger.

antares1470 (approved):

PR #1570, opened 2021-11-24
More tests for JSingleChoiceComboBoxMenu and JDropDownMenu

Auriga2:
### Prerequisites

### Description of the Change
Fixed some tests and added more tests for JSingleChoiceComboBoxMenu and JDropDownMenu which involved a bit of code refactoring.

### Benefits
More coverage

### Verification Process
All tests pass

### Applicable Issues

### DIFF
--- a/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JDropDownMenu.java
+++ b/CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JDropDownMenu.java
@@ -19,6 +19,7 @@ import java.awt.Point;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.net.MalformedURLException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -84,6 +85,30 @@ public class JDropDownMenu<E> extends JComponent implements ActionListener {
             this.button = new JButton(text, arrow);
         }
         button.setHorizontalTextPosition(SwingConstants.LEFT);
+        addButtonActionListener(button, menu);
+        add(button);
+
+        listeners = new HashSet<>();
+
+        items.forEach(item -> {
+            final JMenuItem menuItem = new JMenuItem(item.toString());
+            menuItem.addActionListener(new OpenAction(menu, button));
+            addMenuItemActionListener(menuItem);
+            menuItems.put(menuItem, item);
+            menu.add(menuItem);
+        });
+    }
+
+    protected final void addMenuItemActionListener(final JMenuItem menuItem) {
+        menuItem.addActionListener(event -> {
+            final JMenuItem changedMenuItem = (JMenuItem) event.getSource();
+            final E changedItem = menuItems.get(changedMenuItem);
+            final ActionEvent actionEvent = new ActionEvent(changedItem, ActionEvent.ACTION_PERFORMED, null);
+            actionPerformed(actionEvent);
+        });
+    }
+
+    protected final void addButtonActionListener(final JButton button, final JPopupMenu menu) {
         button.addActionListener(event -> {
             if (!menu.isVisible()) {
                 final Point p = button.getLocationOnScreen();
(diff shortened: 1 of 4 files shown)


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JDropDownMenu.java
```diff
@@ -84,6 +85,30 @@ private JDropDownMenu(final String text, final Icon icon, final List<E> items) {
...
+            menu.add(menuItem);
+        });
+    }
+
+    protected final void addMenuItemActionListener(final JMenuItem menuItem) {
+        menuItem.addActionListener(event -> {
+            final JMenuItem changedMenuItem = (JMenuItem) event.getSource();
+            final E changedItem = menuItems.get(changedMenuItem);
+            final ActionEvent actionEvent = new ActionEvent(changedItem, ActionEvent.ACTION_PERFORMED, null);
+            actionPerformed(actionEvent);
```
antares1470, line 107: a bit confused by this change. Previously this was done in a thread for each listener but that seems to have been omitted here so this would behave differently.

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JSingleChoiceComboBoxMenu.java
```diff
@@ -158,10 +141,50 @@ public final void setSelectedItem(final E item) {
...
+    }
+
+    protected final void addMenuItemActionListener(final JMenuItem menuItem) {
+        menuItem.addActionListener(event -> {
+            final JMenuItem changedMenuItem = (JMenuItem) event.getSource();
+            final E changedItem = menuItems.get(changedMenuItem);
+            setSelectedItem(changedItem);
+            final int menuIndex = menu.getComponentIndex(changedMenuItem);
+            final ListSelectionEvent selectionEvent = new ListSelectionEvent(changedItem, menuIndex, menuIndex, false);
+            valueChanged(selectionEvent);
```
antares1470, line 155: similar situation here

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/gui/JDropDownMenu.java
Auriga2: The actionPerformed it calls executes in a thread for each listener:

> 
> 
>     public void actionPerformed(final ActionEvent event) {
>         listeners.forEach(listener -> {
>             final Thread thread = new Thread("JDropDownMenu Worker") {
>                 @Override
>                 public void run() {
>                     listener.actionPerformed(event);
>                 }
>             };
>             thread.start();
>         });
>     }

antares1470: gotcha. Didn't see that but that makes sense

antares1470 (approved):

Auriga2:
Below test fails in the CI, which passes locally. It's the the same function which executes ` listener.actionPerformed(event);` in a thread for each listener. I added a `Thread.sleep` but didn't work. Any suggestion to fix this please?
image

Auriga2:
Ok it has worked:
image

antares1470:
@Auriga2 so it worked in this case but generally the advice is against using Thread.sleep is a unit test if you can avoid it. The reason being is that different computers and systems will take different periods of time to complete a task and so sleeping isn't a guarantee for it to work.

In this case, the issue is a thread doesn't wait on another thread to finish unless you tell it otherwise so as soon as you create the thread, it's moving on to running the verify checks without being sure the code in the other thread has finished executing yet. Since your verify checks rely on that code having been executed in that thread, ideally you'd want to write your test in such a way that you can guarantee that the thread has finished its job before you run the verify checks (regardless on what system it is run on)

antares1470 (changes requested):
In light of my above comment

Auriga2:
I agree with your reasoning but I couldn't find a better option. I cannot implement the sonar suggestion because I cannot check the condition that must be fulfilled in the thread because it's just `listener.actionPerformed(event);`

 In our code base the following was used, for `Platform.runLater()`:
```
        final CountDownLatch latch = new CountDownLatch(1);
        Platform.runLater(() -> latch.countDown());
        latch.await(); 
```
So here I get this error when  using it and running the test:
image

There are a few other places Thread.Sleep is already used. Appreciate any other suggestions.

antares1470 (approved):

arcturus2 (approved):

Issue #1460, opened 2021-09-30
Runtime error with Cluster > K-truss when there are no K-trusses on graph and clicking on "V"


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Constellation is throwing a runtime error on the K-Truss screen if there are no K-trusses on the graph and the user selects the "V" button (If there are K-trusses in that section it would show them as a line graph). The run time error is happening with a new graph with only a couple of nodes, or a larger graph that has more nodes but doesn't have any triangular relationships. 
The following run time error is shown: 

java.lang.ArithmeticException: / by zero
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.ktruss.NestedKTrussDisplayPanel.rectY(NestedKTrussDisplayPanel.java:126)
	at au.gov.asd.tac.constellation.plugins.algorithms.clustering.ktruss.NestedKTrussDisplayPanel.paintComponent(NestedKTrussDisplayPanel.java:93)

### Steps to Reproduce

1.  Open a new graph and connect a couple of nodes
2.  Go to Tools > Cluster > Ktruss
3.  Click on Cluster
4. Click on Toggle Interactive Disabled
5.  Click on the "V" button which is top left of the screen > run time error
6. To get rid of this runtime error you will need to close and reopen Constellation
7.  Repeat with another graph that doesn't have the triangular relationships > same run time error

**Expected behaviour:** The runtime error is caught before it is shown to me

**Actual behaviour:** Run time error as per the test steps
**Reproduces how often:** 100%

### Additional Information

![image](https://user-images.githubusercontent.com/83679004/135386173-492ee07d-0959-4fde-a6cf-2dcda97ac59f.png)
![image](https://user-images.githubusercontent.com/83679004/135386616-74f76855-baf8-40fe-9e02-7372e856e1c9.png)

GammaVel:
[...]

PR #1604, opened 2021-12-17
Prevent exception of divide by zero in k-truss

aldebaran30701:
### Prerequisites

### Description of the Change
Pane was resizing by diving by zero because of the `totalNeededHeight` variable. Adding a check to avoid dividing by zero seemed to prevent the exception. I have spend the last few hours trying to correct the actual problem of it being zero and checking if the current solution causes any detrimental behaviour but have not found anything.

I actually could not seem to create a graph that had nested k-trusses for testing. The pane will resize just as it did in previous versions on standard working graphs so I suspect this PR is fine.

### Alternate Designs
Tried to find alternate solution which involved avoiding the  case of zero being the variable value but came up short.

Avoids exception to user by defensively checking variable values.

### Benefits
Avoids exception to user by defensively checking variable values.

### Possible Drawbacks
Possibly unfound drawbacks due to not being able to figure out how to trigger a nested ktruss.

### Verification Process
Creating and manipulating the ktruss view to try and throw an exception.
Using sphere graphs to test the functionality is not affected from this change. 
Toggling the up and down arrow 'V' and '^'

### Applicable Issues
#1460

### DIFF
--- a/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/ktruss/NestedKTrussDisplayPanel.java
+++ b/CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/ktruss/NestedKTrussDisplayPanel.java
@@ -123,12 +123,18 @@ public class NestedKTrussDisplayPanel extends JPanel implements MouseInputListen
     // Returns the y position on this panel of the given rectangle.
     private int rectY(final int rectangleNum) {
         final int y = rectangles[rectangleNum][2];
+        if(totalNeededHeight == 0){
+            return y;
+        }
         return expandHeight ? (y * PREFERRED_HEIGHT) / totalNeededHeight : y;
     }
 
     // Returns the height in this panel of the given rectangle
     private int rectHeight(final int rectangleNum) {
         final int h = rectangles[rectangleNum][1];
+        if(totalNeededHeight == 0){
+            return h;
+        }
         return expandHeight ? (h * PREFERRED_HEIGHT) / totalNeededHeight : h;
     }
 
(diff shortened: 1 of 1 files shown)


CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/ktruss/NestedKTrussDisplayPanel.java
```diff
@@ -123,12 +123,18 @@ private int rectX(final int rectangleNum) {
     // Returns the y position on this panel of the given rectangle.
     private int rectY(final int rectangleNum) {
         final int y = rectangles[rectangleNum][2];
+        if(totalNeededHeight == 0){
+            return y;
+        }
         return expandHeight ? (y * PREFERRED_HEIGHT) / totalNeededHeight : y;
```
antares1470, line 129: Given the return values here, could you potentially add your added if condition to the ternary below it?
i.e. `return expandHeight && totalNeededHeight > 0 ? ... : y`
I'm assuming here that `totalNeededHeight` isn't going negative but could also be changed from `>` to `!=` if need be

CoreAlgorithmPlugins/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/ktruss/NestedKTrussDisplayPanel.java
```diff
@@ -123,12 +123,18 @@ private int rectX(final int rectangleNum) {
...
     }
 
     // Returns the height in this panel of the given rectangle
     private int rectHeight(final int rectangleNum) {
         final int h = rectangles[rectangleNum][1];
+        if(totalNeededHeight == 0){
+            return h;
+        }
         return expandHeight ? (h * PREFERRED_HEIGHT) / totalNeededHeight : h;
```
antares1470, line 138: similar here

antares1470 (changes requested):

arcturus2 (approved):

antares1470 (approved):

Issue #1599, opened 2021-12-15
Rename Step Tabs in Data Access View


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Would be handy to be able to rename the step tabs in the Data Access View. This would allow you to then make clearer what each step is aiming to do. This could also then be saved and loaded as part of the templates.

### Behaviour

* Try to include measurable metrics that can be aimed for (if applicable to the issue),
e.g. if it's a time issue then "should take 5 seconds or less to finish"
or if it's a workflow issue "should be accomplished in 2 or less clicks".

-->

**Expected behaviour:** Step Tabs can be renamed

**Actual behaviour:** Step Tabs can't be renamed, default to 'Step 1', 'Step 2', etc

**Reproduces how often:** 100%

Auriga2:
The current version doesn't store the step caption in the json template when saving. Added a new field called "step", but there's a backward compatibility issue if the user tries to open a new template with a previous version of Consty.
![image](https://user-images.githubusercontent.com/64826928/149244966-9cf00cf4-d42b-4817-a65b-d84f1c43431c.png)


```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "step" (class au.gov.asd.tac.constellation.views.dataaccess.api.DataAccessUserPreferences), not marked as ignorable (2 known properties: "plugins", "global"])
 at [Source: (File); line: 2, column: 13] (through reference chain: java.util.ArrayList[0]->au.gov.asd.tac.constellation.views.dataaccess.api.DataAccessUserPreferences["step"])
	at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:61)
	at com.fasterxml.jackson.databind.DeserializationContext.handleUnknownProperty(DeserializationContext.java:987)
	at com.fasterxml.jackson.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:1974)
[...]

PR #1610, opened 2022-01-14
Enable renaming of Step Tabs in Data Access View

Auriga2:
### Prerequisites

### Description of the Change
Step Tabs in Data Access View can now be renamed by double clicking on them. This is also saved and loaded in the templates.

### Alternate Designs
n/a

Better UX

### Benefits
Better UX

### Possible Drawbacks
If the user tries to open a new template with a previous version of Consty, com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException is thrown as explained in https://github.com/constellation-app/constellation/issues/1599

### Verification Process

1. Open DAV and rename the Step tab by double clicking on the default tab caption `Step 1` 
2. Add more tabs and rename 
3. Save template and see if it loads the new names correctly when the template is loaded again.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1599


arcturus2:
@Auriga2 just in case you didn't realise there is a unit test failing.

Auriga2:
Yeah saw it. A few tests fail and working on them. 2 of them are a bit tricky because it fails with "the mock exception could not initialize mocked construction"

Auriga2:
In DAV, currently the tabs are renamed in order if any tab in the middle is removed. With this new functionality that feature is no longer needed, hence removed. E.g., If `Step 2` is closed from `Step 1, Step 2 , Step 3`, it was renaming the remaining tabs to `Step 1, Step 2`
Was there a specific reason for this functionality anyway? One guess is that because it wasn't saving the tab captions in templates, it was adding them back in order when loading the template anyway.

Auriga2:
The other issue is, when adding new tabs after closing a tab it can duplicate tab names. I guess that's okay now because the user can rename any time. Or should we handle this differently?

arcturus2:
> Was there a specific reason for this functionality anyway? One guess is that because it wasn't saving the tab captions in templates, it was adding them back in order when loading the template anyway.

I think you've understood it correctly. Given the tab names were not stored and the index was used then it made sense for each step to just be sequential numbers. Also from a user perspective it makes sense to have numbers in order without skipping numbers because it does not make much sense to jump from step 1 to 3 if you are not trying to skip step 2.

>The other issue is, when adding new tabs after closing a tab it can duplicate tab names. I guess that's okay now because the user can rename any time. Or should we handle this differently?

I don't think it is a major problem if the number recalculation feature was removed though not all users will be renaming each step and will just keep the default step names so having it recalculate is handy as per my comments above. What do you think?

Auriga2:
@arcturus2 

> I don't think it is a major problem if the number recalculation feature was removed though not all users will be renaming each step and will just keep the default step names so having it recalculate is handy as per my comments above. What do you think?

So we shouldn't recalculate the users renamed tabs. Do you mean do recalculate by checking whether they are of the default format `Step <Number>`? 

E.g. Consider a  set of tabs `Step 1, Step 2, New step 3`
When a new tab is added it will be ->`Step 1, Step 2, New step 3, Step 3`
 If `Step 1`or `Step 2` is removed ->  `Step 1, New step 3, Step 2`

Consider a  set of tabs `Step 1, Step 2, Step 3, Step 4` If you rename `Step 3` to `Some random text` -> Tabs will be renamed to `Step 1, Step 2, Some random text, Step 3`

antares1470:
@Auriga2 @arcturus2 the way I see it is that tabs that have been renamed (i.e. aren't the default given name) won't be recalculated when tabs are removed. The other thing that I think should happen is that the default tab name number applied should reflect the number of tabs, not the number of tabs with default names. I'll demonstrate with an example:

Start: Step 1, Step 2, blah, Step 78 (first 2 tabs are default names, 3rd and 4th tabs has been renamed)

_add a new tab_

Result: Step 1, Step 2, blah, Step 78, Step 5 (default name of new tab is Step 5 since it is the 5th tab)

_remove Step 2_

Result: Step 1, blah, Step 78, Step 4 (renamed tabs doesn't change, default named Steps after the one removed decrease number by one which reflects current behaviour)

Another thing I think would be nice add is that if renamed tab is set to empty (i.e. delete all the text from the tab name and accept that change) that it revert back to default naming (rather than being empty)

A couple of things I thought of though that may need to be considered:

1) What would be expected to happen if you rename a tab to what the default name would be? e.g. if you rename Step 2 (from my above example) to 'Step 3', that is a name change and so removing a tab wouldn't change it's name. On the other hand, if I then renamed the tab again to 'Step 2', would it be considered renamed (i.e. if Step 1 were removed it would retain its name 'Step 2') or part of the default naming again (i.e. if Step 1 were removed it would change to 'Step 1')
2) If I create a tab with a name that may clash with another tab either currently or possibly in the future (e.g. rename Step 2 to 'Step 1' when there already exists a Step 1 or rename Step 2 to 'Step 5' and then add a 5th tab which would by default become 'Step 5'), what happens?

Auriga2:
In your example `Step 78` is considered a renamed tab. Imagine that was `Step 1`, then it'd be very confusing for the users to distinguish what they changed, and it could appear as a bug.

To answer 1 and 2, I guess that's why I suggested we won't automatically rename any name initially. Since we have added the ability to rename tabs, the user can rename if there are duplicates.

If we really want to rename, I guess considering anything of the format `Step <Number>` as a default format and renaming them all, rather than tracking whether any of them is user edited would make it less complicated.  Of course this wouldn't let a user to have a `Step 78,` as in your example, but they can instead add something like `Step-78', so shouldn't be problem. What do you think?

antares1470:
> In your example `Step 78` is considered a renamed tab. Imagine that was `Step 1`, then it'd be very confusing for the users to distinguish what they changed, and it could appear as a bug.

You're correct in saying that. The point of my post and the example was to demonstrate firstly that it could be possible to rename it to something similar to the default convention (I would assume the average wouldn't do such things but this would be an edge to factor in) and that being able to do so raises some potential issues

> To answer 1 and 2, I guess that's why I suggested we won't automatically rename any name initially. Since we have added the ability to rename tabs, the user can rename if there are duplicates.

I still think we want to rename simply because that is the current behaviour (and makes sense intuitively) and so if there is a way we can add this feature to complement the current behaviour rather than replace it that is ideal.

One way you could possibly address all the issues above is by not actually allowing users to change the tab numbers themselves but merely allowing adding a name to the tab. e.g. If I take Step 2 and give it the name 'bob', the name of the tab would become something like 'Step 2 - bob' rather than simply 'bob'. If you removed Step 1 then, the name would change to 'Step 1 - bob' and if you removed bob from the name it would become 'Step 1'. The tradeoff is that this will naturally make the tab names longer if you want to add a name but it may be a necessary tradeoff if we want to avoid the issues mentioned earlier without getting silly.

How does that sound @Auriga2?

Auriga2:
@antares1470 That sounds less complicated. I'll implement it. It'll even be possible to allow them to just edit the 2nd part, making the default part non editable.

Auriga2:
The default text is added in the end, only when it's added by a user. E.g.: `New name - Step 1` 
image

Auriga2:
Can't figure out how to fix the failing unit tests here. They seem to be in `constellation.utilities.icon` which doesn't fail locally.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -140,20 +158,43 @@ public QueryPhasePane newTab(final PluginParameters globalPrameters) {
      * 
      * @param queryPane the pane that will be added as the content of the new tab
      */
-    public void newTab(final QueryPhasePane queryPane) {
-        final Tab newTab = new Tab(String.format(TAB_TITLE, getTabPane().getTabs().size() + 1));
-        
+    public void newTab(final QueryPhasePane queryPane, final String tabCaption) {
+        final Label label = new Label(tabCaption);
+        String defaultCaption = String.format(TAB_TITLE, getTabPane().getTabs().size() + 1);
```
antares1470, line 163: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -140,20 +158,43 @@ public QueryPhasePane newTab(final PluginParameters globalPrameters) {
...
+
+        // This is only needed to make the unit test work because the `Tab` is mocked with mockConstruction.
+        // final Tab newTab = new Tab() fails the unit test.
+        final Tab newTab = new Tab(tabCaption);
+        newTab.setText(null);
+
+        if (StringUtils.isBlank(tabCaption)) {
+            label.setText(defaultCaption);
+        } else {
+            if (!defaultCaption.equals(tabCaption.trim())) {
```
antares1470, line 173: can be merged to become `} else if {...`

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -140,20 +158,43 @@ public QueryPhasePane newTab(final PluginParameters globalPrameters) {
...
-        final Optional<EventHandler<Event>> origOnClose =
-                Optional.ofNullable(newTab.getOnClosed());
+        final Optional<EventHandler<Event>> origOnClose
+                = Optional.ofNullable(newTab.getOnClosed());
         newTab.setOnClosed(event -> {
             int queryNum = 1;
             for (Tab tab : getTabPane().getTabs()) {
-                tab.setText(String.format(TAB_TITLE, queryNum));
+                if (tab.getText() == null) {
+                    Label currentDefaultCaption = (Label) tab.getGraphic();
```
antares1470, line 190: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -179,24 +220,61 @@ public void newTab(final QueryPhasePane queryPane) {
...
+                tab.setGraphic(label);
+            });
+
+            field.focusedProperty().addListener((ObservableValue<? extends Boolean> observable, Boolean oldValue,
+                    Boolean newValue) -> {
+                if (!newValue) {
+
+                    label.setText(field.getText().trim());
+                    tab.setGraphic(label);
+                    String newDefaultCaption = String.format(TAB_TITLE, getTabPane().getTabs().indexOf(tab) + 1);
```
antares1470, line 256: final

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -179,24 +220,61 @@ public void newTab(final QueryPhasePane queryPane) {
...
+
+                    if (field.getText() == null || StringUtils.isBlank(field.getText())) {
```
antares1470, line 258: can remove the null check (`StringUtils.isBlank` will do this for you)

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -179,24 +220,61 @@ public void newTab(final QueryPhasePane queryPane) {
...
+                        Label currentDefaultCaption = (Label) tab.getGraphic();
```
antares1470, line 259: final

antares1470 (changes requested):
A few other things as well:
1) I'm noting that the new name is being added before the "-" whereas my head expects it to be after. Is it possible to make that change (naming a tab "blah" would make it "Step x - blah" rather than "blah - Step x")?
2) Also noted after renaming a tab that if you want to rename the tab again, you have to click on the renamed part of the tab (double-clicking on the Step name doesn't do anything). Would be good to be able to rename after double-clicking on any part of the tab if possible.
3) If you're able to add a whatsnew entry, that would be great

Auriga2:
1 is a bit tricky because I use Tab text to store the default Caption and the inserted label to store the user edited part. I couldn't make the label to appear after the Tab text, but I'll give it another try.

Auriga2:
Tried all solutions in https://stackoverflow.com/questions/30525821/in-javafx-is-there-a-way-to-display-tab-name-first-and-then-graphic but they didn't work. Can't figure out why because it should work as it `Specifies the positioning of the graphic relative to the text.`  (https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Labeled.html) 

> graphicProperty
An optional icon for the Labeled. This can be positioned relative to the text by using [setContentDisplay](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Labeled.html#setContentDisplay-javafx.scene.control.ContentDisplay-)

Could it be expecting only am image as the graphic? 

My next option is adding a HBox, but it needs more refactoring.

Auriga2:
Ok the property works within the label- the label text and the icon would position as specified, but it doesn't work for Tab text and the label which kind of makes sense.

antares1470 (approved):

antares1470 (changes requested):
Actually my apologies, if you're able to add a whatsnew entry, that would be great

antares1470 (approved):

arcturus2 (changes requested):

arcturus2:
image

Can we have it the other way round so its "Step 1 - userCaption"? I feel that makes more sence in terms of reading and knowing the order. The unit tests will need an update but the `whatsnew.txt` mentions it in this way :smile:.

arcturus2:
> image
> 
> Can we have it the other way round so its "Step 1 - userCaption"? I feel that makes more sence in terms of reading and knowing the order. The unit tests will need an update but the `whatsnew.txt` mentions it in this way 😄.

Ignore me @Auriga2. It turns out my branch was not up to date locally.

arcturus2:
image

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -140,20 +160,39 @@ public QueryPhasePane newTab(final PluginParameters globalPrameters) {
      * 
      * @param queryPane the pane that will be added as the content of the new tab
      */
-    public void newTab(final QueryPhasePane queryPane) {
-        final Tab newTab = new Tab(String.format(TAB_TITLE, getTabPane().getTabs().size() + 1));
-        
+    public void newTab(final QueryPhasePane queryPane, final String userCaption) {
+        final Label label = new Label(userCaption);
+        final Label defaultLabel = new Label(userCaption); // `userCaption` is only needed for
+        //unit test, because the `Label` is mocked with mockConstruction containing a parameter.
```
arcturus2, line 166: I would like to better understand this. Is this a limitation with Mockito or our code/implementation. What is the actual issue?

Auriga2: I think it's a limitation with Mockito.MockedConstruction.  Because the Label is mocked using MockedConstruction, once with a string parameter, it didn't allow a second label to be mocked with no parameters. Perhaps there's a way to fix it but I couldn't find.

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 -   Removed unused classes `DecoratorUtilities` and `LabelUtilities` from Core 
     Visual Schema
 
-## Changes in January 2022
+## Changes in February 2022
```
antares1470, line 8: You'll need to just remove this as there is already Changes in February header above this one

antares1470 (changes requested):

CHANGELOG.md
Auriga2: Oops, sorry! Removed it now.

antares1470 (approved):

arcturus2 (approved):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
arcturus2: Thanks @Auriga2 I think that this is OK, one less assertion but makes the implementation just that little bit simpler.

PR #1640, opened 2022-02-09
IconManager test fixes

antares1470:
### Prerequisites

### Description of the Change

There were a couple of issues with the tests for IconManager which I have cleaned up:
- one of the tests relied on a lookup returning a particular answer which can't be guaranteed if people create their own implementations of interfaces. I couldn't find a better way to write the test so omitted it
- one test actually had some incorrect expected values so I fixed those up
- some tests using a mock static didn't ensure that other functions called from the class called the actual method which was causing issues. Fixed this as well by making the real method call the default (unless explicitly mentioned otherwise)
- there was a also a lack of cleanup after the tests were run which meant there were issues if the tests were run more than once so I have added a cleanup

As an aside, there was also an unnecessary conditional found in one of the functions in `DefaultCustomIconProvider` so this conditional was removed

fixes unit tests

### Benefits

fixes unit tests


Auriga2 (approved):
All tests pass for me in Utilities now.

Issue #1600, opened 2021-12-15
Viewing Property Value in Histogram View


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

In the Histogram, the property values are sometimes cut off and the column is unable to be stretched out to fit. The count bars end up taking up the majority of the space and when you make the view wider, it is the count column rather than the property value column that expands. 

What I propose is altering this such that when you widen the histogram view that the property value column is the one that expands and the count column remains static (i.e. property value width is dynamic and count width is static, not the other way around). The count bars themselves could probably also be smaller by default given the counts themselves are unlikely to ever be much bigger than 7-8 digits (whereas they property values themselves can often easily exceed that length)

### Steps to Reproduce

1. Open Histogram and a non-empty graph
2. Select a property in the drop-down in the Histogram
3. Make the view wider by dragging it across from the side

**Expected behaviour:** Property Value column should have all values fully visible by default or if not, the view can be adjusted to make the values fully visible

**Actual behaviour:** Property values usually not fully visible by default, widening view makes count column longer

**Reproduces how often:** 90%

### Additional Information

Here's a gif of the issue. Note that the Property Value column title itself is cut off. A similar thing happens for the actual values if they're long enough:
![histogramPropertyValueVisibility](https://user-images.githubusercontent.com/62311243/146125144-e9914b7c-b1fa-4f36-83d6-b4182cb937d9.gif)

GammaVel:
This is a really great improvement to have, thanks @Delphinus8821  for working on this and thanks @antares1470 for coming up with the concept. I've got a couple of minor suggestions to improve this if possible:

1. The count text is getting overlaid by the selected text when you narrow the view. It stops at a certain point. Would it be possible to stop it sooner, here is a gif of the issue:
![constellation text overlay](https://user-images.githubusercontent.com/83679004/152266644-92e5607a-1ac6-4072-a908-291758cc766a.gif)

[...]

PR #1644, opened 2022-02-14
Fixed Property Value Text Size in Histogram

Delphinus8821:
### Prerequisites

### Description of the Change
The property value minimum text width has been updated to be the same width as the Property Value column title text and now expands as required when the Histogram is expanded. 

### Alternate Designs
N/A

Fixes previous change that made the histogram bars not viewable when a vertex attribute is too long. 
Also stops the Count and Selected / Total Bin Count text for overlapping. 

### Possible Drawbacks
N/A

### Applicable Issues
#1600


antares1470 (changes requested):
histogramPropertyValue

It seems that there is still a struggle to see the property value. Even when expanding the view, it stops after a particular point (the value length is far from the longest you could have as well). Ideally if possible, it would be good to have the count bar as a fixed length (because that really doesn't need to change length) and then expanding the view expands the column width of the property value

antares1470 (approved):

aldebaran30701 (approved):

Issue #1653, opened 2022-02-24
New Find View Advanced Find Null Pointer 


Delphinus8821:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When using the Advanced Find in the new Find View to search for an attribute that some elements of the graph do not have a value for, a null pointer is thrown. 

### Steps to Reproduce

1.  Create a new sphere and then add an additional node, ensuring this node does not have a value for the attribute Geo.Country 

2.  Use the Advanced Find to search for the node attribute Geo.Country using either 'Contains', 'Doesn't Contain', 'Begins With', 'Ends With' or 'Matches', and search for any string value - example pictured below uses 'Australia'. 
![image](https://user-images.githubusercontent.com/73083035/155455836-c6e19302-744d-4148-ab02-00e6f91d00e0.png)

3.  Click Find All, Find Previous or Find Next and the null pointer will be thrown

**Expected behaviour:** Search to run without issue

**Actual behaviour:** A null pointer is thrown 

**Reproduces how often:** 100%

### Additional Information
Null Pointer Exception:
![image](https://user-images.githubusercontent.com/73083035/155455772-9a6659ac-6118-4207-bfb9-1ebed9af6760.png)

GammaVel:
Thanks @Delphinus8821 for detecting this error and fixing it up. It is working in the latest nightly build.

PR #1657, opened 2022-02-28
New Find View Null Pointer Check 

Delphinus8821:
### Prerequisites

### Description of the Change
Added null checks to the new advanced find to stop null pointer exceptions from being thrown. 

### Alternate Designs
None realised 

Less bugs in core.

### Possible Drawbacks
None realised

### Verification Process
Attempted reproducer from the original ticket (#1653) in multiple scenarios and could not reproduce the null pointer exception. 

### Applicable Issues
#1653


antares1470 (commented):
if `value` is blank, would it be better to just do a blanket check over the whole function (i.e. once `value` is extracted, check if it's blank and only go ahead with the rest of the function it it's not)? Would save having to do individual checks before every time it gets used

Delphinus8821:
@antares1470 If I put a blanket check over the whole function then it doesn't correctly pick up values for the 'Is Not' filter

antares1470 (approved):

aldebaran30701 (approved):

PR #1639, opened 2022-02-09
Testing <Multiple>IOProvider Files.

serpens24:
### Description of the Change

Providing uinit tests of BlazeIOProvider

### Alternate Designs

nil

### Why Should This Be In Core?

Improve code quality

### Benefits

Improve code quality, reduce code duplication.

### Possible Drawbacks

None known

### Verification Process

Run tests and confirm tests pass and coverage is acceptable.

### Applicable Issues

nil

PR #1665, opened 2022-03-08
Visual Schema Models test and code comments/cleanup.

serpens24:
### Description of the Change
Unit tests and code comment additoon/cleanup for code in Visual Schema Models

### Alternate Designs
Nil

Improve code quality

### Benefits
Improve code quality

### Possible Drawbacks
None known

### Verification Process
Test all pass

### Applicable Issues
Nil

### DIFF
--- a/CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/BlazeNGTest.java
+++ b/CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/BlazeNGTest.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2010-2022 Australian Signals Directorate
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.graph.schema.visual.attribute.objects;
+
+import au.gov.asd.tac.constellation.graph.schema.visual.attribute.objects.Blaze.IllegalBlazeFormatException;
+import au.gov.asd.tac.constellation.utilities.color.ConstellationColor;
+import static org.testng.Assert.*;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+/**
+ *
+ * @author serpens24
+ */
+public class BlazeNGTest {
+    
+    public BlazeNGTest() {
+    }
+
+    @BeforeClass
+    public static void setUpClass() throws Exception {
+    }
+
+    @AfterClass
+    public static void tearDownClass() throws Exception {
+    }
+
+    @BeforeMethod
+    public void setUpMethod() throws Exception {
+    }
+
+    @AfterMethod
+    public void tearDownMethod() throws Exception {
+    }
+
+    /**
+     * Test of getAngle method, of class Blaze.
+     */
+    @Test
+    public void testGetAngle() {
+        System.out.println("BlazeNGTest.testGetAngle");
+        final int angle = 23;
+        Blaze instance = new Blaze(angle, ConstellationColor.YELLOW);
+        assertEquals(instance.getAngle(), angle);
+    }
+
+    /**
+     * Test of getColor method, of class Blaze.
+     */
+    @Test
+    public void testGetColor() {
+        System.out.println("BlazeNGTest.testGetColor");
+        Blaze instance = new Blaze(0, ConstellationColor.YELLOW);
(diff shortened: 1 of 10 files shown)


serpens24:
Reviewers, SonarQube seems to be failing due to duplicate code .. .there are X,Y,Z setters. I don't see any value in abstracting that away given the context of where the otehr setters are (in uncollide code). Is it possible as marking this as acceptable?

Note, also havent tried to tackle mocking the ImageIO stuff ATM.

antares1470:
> Reviewers, SonarQube seems to be failing due to duplicate code .. .there are X,Y,Z setters. I don't see any value in abstracting that away given the context of where the otehr setters are (in uncollide code). Is it possible as marking this as acceptable?

Yeah abstracting makes no sense here. Sonar has picked it up because of the name for getters and setters but as you say, the context is completely different between the 2 classes in question. Mark it as won't fix makes sense here

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/BlazeNGTest.java
```diff
@@ -0,0 +1,185 @@
...
+        
+        // invalid string (bad color)
+        try {
+            Blaze instance = Blaze.valueOf("99;fruitsalad");
+            fail("Exception not thrown");
+        } catch (IllegalBlazeFormatException ex) {
+            assertTrue("Undefined colour for blaze.".equals(ex.getMessage()));
+        } catch (Exception ex) {
+            fail("Incorrect exception thrown");
+        }
```
antares1470, line 96: wondering whether it may be better to separate out this use case into a separate unit test. The benefit there is that you can annotate the test with the expected exception (IllegalBlazeFormatException) and the expected message (Undefined colour for blaze)

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/ConnectionModeNGTest.java
```diff
@@ -0,0 +1,352 @@
...
+    public void testValueOf() {
+    }
+
+    /**
+     * Test of getConnection method, of class ConnectionMode.
+     */
+    @Test
+    public void testGetConnection() {
+
+        resetMocking();
```
antares1470, line 91: if you're needing to reset after each use case, it could be worth separating each one into its own test (and then resetting in setUpMethod). Else in this case (and subsequent tests), there may not be any harm in defining all the whens together and the verifies are then cumulative (the reset can be done inbetween tests as well)

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/BlazeNGTest.java
serpens24: I looked at that and beleive it makes the test suite bulkier wihout any real gain, but can change if you want it that way.

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/ConnectionModeNGTest.java
serpens24: I had considered, but then you need to invent a heap of differen ttest case names like "TestGetConnectionWhenSomeFunctionReturnsNull" and "TestGetConnectionWhenSomeFunctionReturnsInvalidId" etc ..... doing it this way test cases can reflect the function under test and have muchcleaner names.

I do the reset as it makes it 'much'clearer to the reader how many instances of a call should occur without having to count back.

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/PlaneNGTest.java
```diff
@@ -0,0 +1,241 @@
...
+     * Test of getImage method, of class Plane.
+     */
+    @Test
+    public void testGetImage() {
+        System.out.println("PlaneNGTest.testGetImage");
+        final BufferedImage value = instance.getImage();
+        assertEquals(value.getWidth(), bufferedImage.getWidth());
+        assertEquals(value.getHeight(), bufferedImage.getHeight());
+        assertEquals(value.getType(), bufferedImage.getType());
+        assertEquals(value.getType(), bufferedImage.getType());
```
arcturus2, line 70: duplicate?

arcturus2 (approved):
I found one duplicate entry but besides that it looks good.

Note that my personal position regarding testing is to test 1 thing at a time to stay true to the core unit testing principle. I would have to agree with @antares1470 that for the cases in this PR and #1639 rather than doing the `resetMocking()`, instead each test be done in a separate method. Noted that there is a lot more typing and coming up with good method names (which will be long) as a result but I think it makes it clearer.  I am not too fussed with the extra boiler plate it adds because Java is like that.

**HOWEVER** we already have a precedent of multiple things being tested in the same test method and I think what you have works well and there are more important things to focus on the backlog than theoretically perfect unit tests. We are still very much in the initial "accumulation" phase for unit tests and the goal right now is to increase the code coverage than anything else.

CoreVisualSchema/test/unit/src/au/gov/asd/tac/constellation/graph/schema/visual/attribute/objects/PlaneNGTest.java
serpens24: it does look lile 69 and 70 are duplicates .... I'll look and fix.

antares1470 (approved):

Issue #1663, opened 2022-03-08
Add Excel DateTime parsing option


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Given Constellation is able to import excel files through the File Importer, it would be helpful to be able to add the native excel datetime as an option for parsing into Constellation as currently that is not an option (you have to convert to Epoch or similar).

For reference, the native format for Excel is days after 1/1/1900. e.g. 1/1/2020 12pm would be 43831.5

### Steps to Reproduce

1. Create an excel spreadsheet with a column with datetimes represented as a numeric value
2. Open File Importer in Constellation and load file in
3. Try to parse the intended column as a datetime

**Expected behaviour:** Option to parse excel datetime

**Actual behaviour:** Can't parse as a datetime field, have to create a seperate column in the spreadsheet which converts to Epoch or similar

**Reproduces how often:** 100%

GammaVel:
This is working a real treat now. Thanks @Auriga2  for all your efforts with setting this up. It is so handy to be able to import from excel now.

PR #1668, opened 2022-03-15
Add Excel DateTime parsing option in File Importer

Auriga2:
### Prerequisites

### Description of the Change
The `EXCEL` is added as an option for parsing native excel datetime into Constellation which is available in the  `DateTime` `Formatter` in the `File Importer`.

### Verification Process

1. Create an excel spreadsheet with a column with datetimes represented as a numeric value
2. Open File Importer in Constellation and load file in
3. Try to parse the intended column as a datetime
4. Right click and set the `DateTime` attribute format to `Datetime`->`EXCEL` and now parse the column as a datetime

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1663

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -60,8 +60,8 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
         DATETIME_FORMATS.put("yyyyMMdd HHmmss'Z'", "yyyyMMdd HHmmss'Z'");
         DATETIME_FORMATS.put("yyyyMMddHHmmss", "yyyyMMddHHmmss");
         DATETIME_FORMATS.put("EPOCH", null);
-        DATETIME_FORMATS.put(CUSTOM, null);
         DATETIME_FORMATS.put(EXCEL, null);
+        DATETIME_FORMATS.put(CUSTOM, null);
     }
 
     public DatetimeAttributeTranslator() {
@@ -115,10 +115,9 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
                     dateTime = df.parse(value);
                     break;
                 }
-                case EXCEL: {
+                case EXCEL: 
                     dateTime = TemporalFormatting.zonedDateTimeFromLong(DateUtil.getJavaDate(Double.parseDouble(value)).getTime());
-                    break;
-                }
+                    break;                
                 default: {
                     format = DATETIME_FORMATS.get(format);
                     final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.poi.ss.usermodel.DateUtil;
 import org.openide.util.lookup.ServiceProvider;
 
 /**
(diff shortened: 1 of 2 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -112,6 +115,10 @@ public String translate(final String value, final PluginParameters parameters) {
                     dateTime = df.parse(value);
                     break;
                 }
+                case EXCEL: {
+                    dateTime = TemporalFormatting.zonedDateTimeFromLong(DateUtil.getJavaDate(Double.parseDouble(value)).getTime());
+                    break;
+                }
```
antares1470, line 121: I know it's there in a couple other cases but if you're able to remove the curly braces (they're not needed), that would be great

antares1470 (changes requested):
I don't think the conversion is working quite right. I passed in a file with a datetime value of 1.00 which should be 1/1/1900 00:00:00 but when I applied the excel formatter it became 31/12/1899 14:00:00. A similar margin of error was occurring for other datetimes as well.

A couple of other points:
1) Are you able to order the formatters so that Excel goes before Custom rather than after? I think Custom makes sense to have as the last option
2) If you're able to add a whatsnew entry, that would be great

arcturus2 (approved):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
Auriga2: @antares1470, When I debugged, the epoch time for above example is -2209023000000 
(https://www.epochconverter.com/) produces both days below, so looks like all dates are converted to GMT. Should we display it in local time zone?
image

Auriga2:
Fixed other requested changes.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
antares1470: Right I guess that makes sense. If converting the excel time to epoch still produces the same value in Constellation and then I suppose for now it makes sense to keep as is. I suppose from my perspective it just looks a little weird inputing a time into a spreadsheet and then it importing as a different time in Constellation. Maybe what I will do though is put in a new issue to give an option to specify a time zone when applying the formatter (excel or epoch)

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,9 @@
 == 3030-12-31 Getting Started
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2022-03-22 New Datetime formatter in the File Importer
+<p>New Datetime formatter 'EXCEL' is added as an option for parsing native excel datetime into Constellation which is available as an 'Datetime Format' in the 'Datetime Parameters' dialog when the DateTime attribute is right clicked and 'Formatter' -> 'Datetime' is choosen.</p>
+
```
antares1470, line 6: I think I would split up this sentence a little. Maybe something like "A new Datetime formatter has been added to the File Importer for parsing native Excel datetimes into Constellation. You can use it by right-clicking an attribute, navigating to Formatter -> Datetime and choosing 'EXCEL' from the format dropdown list"

antares1470 (changes requested):

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
Auriga2: Yes it' much better. Changed now.

antares1470 (approved):

Auriga2:
Thanks @GammaVel for thorough testing. A further investigation revealed the problem.  `DateUtil.getJavaDate` I use in the conversion applies the user's time zone if a time zone is not specified, which made all the confusion. If I specify "GMT" there, it works similar to EPOCH converter. I'll fix that here and the rest below can be handled separately.

https://github.com/constellation-app/constellation/issues/1673 still applies after fixing this, because the last part of the converted datetime implies they are GMT, where it could actually be any time zone. (Added more details in that issue.)
image

Auriga2:
Created https://github.com/constellation-app/constellation/pull/1685.

Issue #48, opened 2019-07-26
Improving UX on Error Message Handling


arcturus2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/faq
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/repositories-catalog

### Description

As a user
i want better indications about errors
so that my experience is enhanced and i get more contextual information.

There is potential for UX improvements around error handling. Some examples of improvements that could be added include:
* changing the colour of the plugin field that caused the error to occur (e.g. red)
* easy-to-understand messages making it clear to the user why an error has occurred.
* replace stack trace prints and other exception-generated messages with custom messages in accordance with the previous point

Discussion has allowed a better understanding to occur - notably that this ticket will hold information about the handling of expected errors and not runtime null pointer errors.

* [ ] Validate input fields
* [ ] Validate preconditions before execution
* [ ] Handle expected errors gracefully
* [ ] Design an extensible way so all plugins have consistent behaviour upon errors.

aldebaran30701:
This issue may need some more clarification - Does it refer the error messages that occur when something fatal has happened (null pointer etc.) If it does, some digging may need to occur to decide if there is a way to capture that exception and log it as an exception window to the user. 

If it refers to an error in an input field, like trying to create more than 32 layers - then maybe a systematic review of input cases should be had. ie. launching a dialog box which describes the error to the user. This would take much longer to implement as it's a case-by-case solution.

arcturus2:
It is a review for all the above you mentioned @aldebaran30701. We have the ability to show popups and there is also the ability for parameters to indicate something is wrong. This isn't widely used so some consistency with the Data Access View and other views is required. I am OK if all we do is document what we expect to do in each case and take it from there.

![Screenshot from 2021-06-02 16-00-48](https://user-images.githubusercontent.com/39325530/120431465-df71a600-c3bb-11eb-8102-0608fbbf204f.png)

antares1470:
Now that the changes I've made have been merged in, this is what I think could still be done as a part of ticket:
- ensure that any required parameter fields for a plugin are filled with sensible default values
[...]

PR #1674, opened 2022-03-25
Plugin Parameter ux improvements

antares1470:
### Prerequisites

### Description of the Change

Have made a number of UX improvements, mostly for plugin parameters:
- added some default values to required parameters where appropriate
- marked required parameters as required (*), making it clear in the interface when parameters are required
- GUI for String, File, Single Choice, and Multi Choice parameters now goes red if you input a blank answer into a required parameter
- replaced remaining stack trace prints with log entries

### Alternate Designs

Considered adding a key to top of the parameter pane (to indicate what * meant) but decided it should hopefully be clear enough without (due it being commonly used for the same thing in other applications and forms)

Improves user experience around plugins. Changes made to parts of Core which will benefit any created modules on top of Core.

### Benefits

Improved user experience

### Applicable Issues

#48

uxParameterImprovementsDemo


mimosa2:
In Analytic view when deselecting transaction type(Unknown) changes the color red as expected. When moving to other analytics and back the previous one doesn't keep the color change.

antares1470:
> In Analytic view when deselecting transaction type(Unknown) changes the color red as expected. When moving to other analytics and back the previous one doesn't keep the color change.

I wonder if this is caused by the analytic view possibly creating a new pane each time you select a different analytic? The catch with my implementation is that required parameters only go red when the value has been changed to something invalid (normally blank). I intentionally did it this way so that the parameter doesn't show red before you have the chance to edit it. This means though that a parameter will never display in the pane as red by default, regardless of whether or not the value contained in the parameter is actually valid.

I think the only way this could effectively be resolved then would be for me to change the logic slightly and allow a field to be red by default if need be. The catch with this though is that it will be blank required fields will all be red, regardless of whether you've actually touched them (or intend to touch them). Personally I prefer what I've done already, but I'm biased to my own thoughts so interested to hear what other people think

aldebaran30701:
If you think about it, the * will tell the user that it is required, and if they miss that it will error or prompt for completion of that field. I don't see a way around apart from validating on creation of the input. Which as you said will just display as red boxes all the time

aldebaran30701 (approved):

arcturus2 (approved):
Nice work @antares1470 and apologies it took so long to review this one.

mimosa2 (approved):

Issue #1673, opened 2022-03-23
Add option to specify display timezone for Datetime formatters


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

In the File or Database Importer, if you use a datetime formatter, by default it will present the time as GMT unless the timezone is otherwise specified. What might be a nice addition to these formatters is the option to specify the output timezone that you import the data as, should you want your values to be of a different timezone.

e.g. Applying the EPOCH datetime formatter will by default import as GMT time but perhaps you would prefer to view the datetimes in local datetime instead.

### Steps to Reproduce

1. Open Constellation and the File Importer
2. Load a file in to the importer
3. Right-click on the Datetime transaction attribute
4. Navigate to Formatter -> Datetime
5. Select EPOCH from the format dropdown list

**Expected behaviour:** Option to specify the timezone to apply, when datetime attribute is applied, it is applied with the specified timezone

**Actual behaviour:** No option to specify timezone, when datetime attribute is applied, it is applied with GMT

**Reproduces how often:** 100%

Auriga2:
The actual issue here is because the last part of the converted time implies they are GMT, where it could actually be any time zone. 
![image](https://user-images.githubusercontent.com/64826928/164358598-d67bb6f8-a07f-4048-8c19-f26b567600b2.png)

For e.g, If the date in the excel is AEST (say 11/06/2017  3:08:00 PM), the converted date will be the same datetime in GMT (2017-06-11 15:08:00 +00:00 ) which can only be `converted` to AEST in the graph that would be 2017-06-12 01:08:00 +10:00. What we really want is 2017-06-11 15:08:00 +10:00. 
This also applies to other listed formats and custom formats as well:
E.g. The 2nd option in the drop down list, "dd/MM/yyyy HH:mm:ss" converts `15/12/2009 23:59:59` and `25/12/2009 23:59:59` to:
[...]

PR #1685, opened 2022-04-21
Remove the user's time zone difference added in the Excel datetime parser

Auriga2:
### Prerequisites

### Description of the Change
The dates formatted by the Excel datetime parser was introducing a time difference equal to that of the constellation user's time zone difference from GMT. This has been fixed now.

### Alternate Designs
The correct time zone details are still not included as explained in https://github.com/constellation-app/constellation/issues/1673

### Verification Process

1. Create an excel spreadsheet with a column with datetimes represented as a numeric value
2. Open File Importer in Constellation and load that file
3. Right click and set the DateTime attribute format to Datetime->EXCEL and now parse the column as a datetime
4. The same datetimes entered in the excel file should be displayed in the converted column, assuming they are in GMT

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1663
https://github.com/constellation-app/constellation/pull/1668

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 import org.apache.poi.ss.usermodel.DateUtil;
 import org.openide.util.lookup.ServiceProvider;
 
@@ -116,8 +117,8 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
                     break;
                 }
                 case EXCEL: 
-                    dateTime = TemporalFormatting.zonedDateTimeFromLong(DateUtil.getJavaDate(Double.parseDouble(value)).getTime());
-                    break;                
+                    dateTime = TemporalFormatting.zonedDateTimeFromLong(DateUtil.getJavaDate(Double.parseDouble(value), TimeZone.getTimeZone("GMT")).getTime());
+                    break;
                 default: {
                     format = DATETIME_FORMATS.get(format);
                     final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);


arcturus2 (commented):
Works well.

image

I think i am keen to see #1673 implemented soon also but for the scope of this work its working.

antares1470:
@Auriga2 I think it would be good to add the dropdown as specified in #1673 before merging this in. The reason for that is that based on my research and understanding, excel doesn't really have a concept of timezone which means that the timezone of a datetime in excel is...whatever timezone you want it to be. That means that it is incorrect to assume that the time specified in the spreadsheet is GMT, but also incorrect to assume that it is local datetime (which is what we currently have). 

Happy for GMT to be the default option, but I think the dropdown is needed to allow users to specify the imported datetimes being of another timezone

antares1470:
> @antares1470 There are some issues we need to discuss in #1673 I think this issue handles the original issue and now excel datetimes behave as same as Epoch datetimes (and other formats). IMO all other date time formats are also affected by this, as explained in #1673 which we can handle seperately.

@Auriga2 so I think your comments in #1673 mainly refer to the output timezone. The issues I mentioned though refer to the input timezone. The problem is that excel doesn't have a concept of timezones and so the timezone of an excel datetime is whichever timezone you want it to be. e.g. if in excel I have a cell with value 1/1/2000 12:00:00, it could be GMT, GMT-5, GMT+5 or any other timezone. That means that converting the time on its own to GMT makes no sense because you have no way of knowing what timezone you're actually converting from (as the input may not be local datetime).

Originally the ticket mentioned both epoch and excel because I didn't realise epoch is based off GMT (for some reason I had in my head it was local time) and so have removed the epoch requirement subsequently. It sounds like though that maybe there was some confusion in my wording as to what I was referring to (being able to specify the output timezone would be a nice addition though wasn't necessarily what I had in mind). I'll adjust the wording to make that more clear.

Does that make sense?

Auriga2:
@antares1470 Yes this is exactly my point. But in my understanding there are no 2 time zones as input and output. Since Excel (And epoch) doesn't contain time zone details we will let the user specify it (in https://github.com/constellation-app/constellation/issues/1673), then we will apply it in the formatted datetime.

 If they wish to convert it into a different time zone later, they can use the attribute editor.

antares1470:
> @antares1470 Yes this is exactly my point. But in my understanding there are no 2 time zones as input and output. Since Excel (And epoch) doesn't contain time zone details we will let the user specify it (in #1673), then we will apply it in the formatted datetime.
> 
> If they wish to convert it into a different time zone later, they can use the attribute editor.

@Auriga2 The thing with Epoch is that it is measured from an absolute point in time (1/1/1970 midnight GMT) and so the value that we import for it will be the correct time, just maybe not necessarily in the desired timezone. With Excel on the other hand, this is not the case. If I have a column of local datetimes and then import them in with the excel formatter, this current solution will import them will the same numbers but with GMT slapped on the end. This means that unless the local timezone was GMT, the data is being **incorrectly** imported as the output isn't equivalent to the input.

The purpose of #1673 is to convert a datetime to the desired timezone, not to just slap a desired timezone on the end.
e.g. `1/1/1970 0:00 GMT` could change to `1/1/1970 10:00 GMT+10`, not `1/1/1970 0:00 GMT+10`

Auriga2:
From what I read, EPOCH is the same.

https://www.techtarget.com/searchdatacenter/definition/epoch

> What is epoch?
> In a computing context, an epoch is the date and time relative to which a computer's clock and timestamp values are determined. The epoch traditionally corresponds to 0 hours, 0 minutes, and 0 seconds (00:00:00) Coordinated Universal Time (UTC) on a specific date, which varies from system to system. Most versions of [Unix](https://www.techtarget.com/searchdatacenter/definition/Unix), for example, use January 1, 1970 as the epoch date; [Windows](https://www.techtarget.com/searchwindowsserver/definition/Windows) uses January 1, 1601; Macintosh systems use January 1, 1904, and Digital Equipment Corporation's Virtual Memory System (VMS) uses November 17, 1858.

image

 Also as I mentioned in https://github.com/constellation-app/constellation/issues/1673, the formatting adds GMT on all other formats as well. So IMO, that ticket applies to all formats not just excel.

antares1470:
> Also as I mentioned in #1673, the formatting adds GMT on all other formats as well

Ok I didn't realise that was a thing for other timezones. I guess that means I won't make you do it here but it should definitely be addressed at some stage. I've created a ticket for it: #1698

antares1470 (approved):

arcturus2 (approved):

Issue #1608, opened 2022-01-14
IllegalStateException in Data access view when multiple tabs are added


Auriga2:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
In v2.5.0 , Data access view throws the below exception if the user loads a template containing few tabs and then immediately click on Tab2. The tabs should have plugins selected, and if you wait a while before clicking on Tab 2, the exceptions is not thrown. 
Sometimes the exception is thrown when a new tab is added and clicked on it immediately, without loading any template.
```
java.lang.IllegalStateException: Not on FX application thread; currentThread = ForkJoinPool.commonPool-worker-3
	at javafx.graphics/com.sun.javafx.tk.Toolkit.checkFxUserThread(Unknown Source)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.checkFxUserThread(Unknown Source)
	at javafx.graphics/javafx.scene.Parent$3.onProposedChange(Unknown Source)
	at javafx.base/com.sun.javafx.collections.VetoableListDecorator.setAll(Unknown Source)
	at javafx.base/com.sun.javafx.collections.VetoableListDecorator.setAll(Unknown Source)
	at javafx.controls/javafx.scene.control.skin.LabeledSkinBase.updateChildren(Unknown Source)
	at javafx.controls/javafx.scene.control.skin.LabeledSkinBase.lambda$new$11(Unknown Source)
	at javafx.controls/com.sun.javafx.scene.control.LambdaMultiplePropertyChangeListenerHandler.lambda$new$1(Unknown Source)
[catch] at javafx.base/javafx.beans.value.WeakChangeListener.changed(Unknown Source)
	at javafx.base/com.sun.javafx.binding.ExpressionHelper$SingleChange.fireValueChangedEvent(Unknown Source)
	at javafx.base/com.sun.javafx.binding.ExpressionHelper.fireValueChangedEvent(Unknown Source)
[...]

PR #1691, opened 2022-05-04
#1608

spica198:
### Prerequisites

### Description of the Change

llegalStateException in Data access view when multiple tabs are added. JavaFX code allows updating the UI from an JavaFX application thread. Need to put the code which is changing the tab pane values under application thread.

### Verification Process

Open DAV, select 4-5 plugins, leaving the default parameters.

Open a second tab, and click on it immediately. There should be no exception thrown.

### Applicable Issues

#1608

### DIFF
--- a/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/ValueInputPane.java
+++ b/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/ValueInputPane.java
@@ -294,18 +294,21 @@ public class ValueInputPane extends HBox implements RecentValuesListener {
     @Override
     public void recentValuesChanged(final RecentValuesChangeEvent e) {
         if (recentValuesCombo != null && parameterId.equals(e.getId())) {
-            recentValuesCombo.getSelectionModel().selectedIndexProperty().removeListener(recentValueSelectionListener);
-            final List<String> recentValues = e.getNewValues();
-            if (recentValues != null) {
-                recentValuesCombo.setItems(FXCollections.observableList(recentValues));
-                recentValuesCombo.setDisable(false);
-            } else {
-                final List<String> empty = Collections.emptyList();
-                recentValuesCombo.setItems(FXCollections.observableList(empty));
-                recentValuesCombo.setDisable(true);
-            }
-            recentValuesCombo.setPromptText("...");
-            recentValuesCombo.getSelectionModel().selectedIndexProperty().addListener(recentValueSelectionListener);
+            //#1608: JavaFX code allows updating the UI from an JavaFX application thread.
+             Platform.runLater(() -> {             
+                recentValuesCombo.getSelectionModel().selectedIndexProperty().removeListener(recentValueSelectionListener);
+                final List<String> recentValues = e.getNewValues();
+                if (recentValues != null) {
+                    recentValuesCombo.setItems(FXCollections.observableList(recentValues));
+                    recentValuesCombo.setDisable(false);
+                } else {
+                    final List<String> empty = Collections.emptyList();
+                    recentValuesCombo.setItems(FXCollections.observableList(empty));
+                    recentValuesCombo.setDisable(true);
+                }
+                recentValuesCombo.setPromptText("...");
+                recentValuesCombo.getSelectionModel().selectedIndexProperty().addListener(recentValueSelectionListener);
+            });
         }
     }
 }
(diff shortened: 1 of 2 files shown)


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/ValueInputPane.java
```diff
@@ -294,18 +294,21 @@ protected void updateItem(final String item, final boolean empty) {
...
-                recentValuesCombo.setDisable(false);
-            } else {
-                final List<String> empty = Collections.emptyList();
-                recentValuesCombo.setItems(FXCollections.observableList(empty));
-                recentValuesCombo.setDisable(true);
-            }
-            recentValuesCombo.setPromptText("...");
-            recentValuesCombo.getSelectionModel().selectedIndexProperty().addListener(recentValueSelectionListener);
+            //#1608: JavaFX code allows updating the UI from an JavaFX application thread.
+             Platform.runLater(() -> {
```
antares1470, line 298: Seems this has an extra space at the front (relative to column I think it is meant to sit in). Might possibly apply to the whole code block(?)

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/ValueInputPane.java
```diff
@@ -294,18 +294,21 @@ protected void updateItem(final String item, final boolean empty) {
...
-                recentValuesCombo.setItems(FXCollections.observableList(recentValues));
-                recentValuesCombo.setDisable(false);
-            } else {
-                final List<String> empty = Collections.emptyList();
-                recentValuesCombo.setItems(FXCollections.observableList(empty));
-                recentValuesCombo.setDisable(true);
-            }
-            recentValuesCombo.setPromptText("...");
-            recentValuesCombo.getSelectionModel().selectedIndexProperty().addListener(recentValueSelectionListener);
+            //#1608: JavaFX code allows updating the UI from an JavaFX application thread.
```
antares1470, line 297: Minor but are you able to omit the ticket number in this comment? Just thinking that it might be a bit confusing if you don't have the context that this number is a ticket number (I think removing will be better than clarifying the context)

antares1470 (changes requested):
Looks fine otherwise

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/ValueInputPane.java
spica198: Done

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/RecentParameterValues.java
```diff
@@ -108,7 +109,8 @@ public static void removeListener(RecentValuesListener listener) {
     public static void fireChangeEvent(RecentValuesChangeEvent e) {
         synchronized (LISTENERS) {
             for (RecentValuesListener listener : LISTENERS) {
-                listener.recentValuesChanged(e);
+                //JavaFX code allows updating the UI from an JavaFX application thread.            
+                Platform.runLater(() -> listener.recentValuesChanged(e));
             }
```
antares1470, line 114: Given this is the only code in the for loop, I'm unsure whether it is better to have the `Platform.runLater` call inside the loop or outside?

arcturus2: @antares1470 what do you mean? The existing logic was already making the call for each listener. So shouldn't we just continue doing this but @spica198 is just making sure it is run on the JavaFX thread?

antares1470: @arcturus2 what I meant was comparing this solution to queuing the actual for loop itself, given everything in the for loop in getting queued on the FX thread. i.e.
```
Platform.runLater(() -> {
  for (listener : LISTENERS) {
  ...
});
```

Auriga2: Both options fix the issue here, but not sure which one is best. Kind of agree with @antares1470. Otherwise it's  overwhelming the event queue with little runnables, most of them would just return in ValueInputPane.recentValuesChanged as the first if condition fails.

mimosa2 (approved):

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/RecentParameterValues.java
spica198: Agree with @antares1470 .. Covering the loop within thread make sense instead of putting each listener in thread.

Auriga2 (approved):

antares1470 (approved):

Issue #1698, opened 2022-05-05
Datetime formatters with no timezone assume GMT


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When importing data using the file or database importers, if you apply a datetime formatter to an attribute and the formatter doesn't specify a timezone, it will by default apply GMT (+0:00) to all the datetimes. This is ok if the data is either already GMT or if timezone isn't important to the data. On the other hand if the data was to be based on another timezone (e.g. perhaps a local datetime not GMT), this can result in some serious data integrity issues as the data itself will have now been imported incorrectly.

I'm proposing for all datetime formatters that don't specify timezones that there is a drop-down list to select the intended timezone. GMT is fine as a default but users need the option to specify otherwise if GMT is incorrect. 

### Steps to Reproduce

1. Open the File Importer, add a file with some datetimes with no timezone specified
2. Change the formatter on the Datetime transaction attribute to datetime and specify the appropriate format (e.g. EXCEL, dd/MM/yyy HH:mm, etc)
3. Apply the Datetime attribute to the column of datetimes and observe the change

**Expected behaviour:** Formatter change should provide option to specify input timezone of datetimes

**Actual behaviour:** GMT automatically slapped on the end if no timezone is specified

**Reproduces how often:** 100% for all formats with no specified timezone

### Additional Information

While seemingly related to #1673, this ticket is focused on the timezone of the input itself (i.e. input timezone) compared to the other ticket which is focused on displaying the data in the desired timezone (i.e. output timezone)

This issue gets a little more tricky for the CUSTOM formatter in which you may or may not specify a timezone. My guts says the best way to resolve that is having the option to specify the timezone available for the CUSTOM formatter but only using it if no timezone is specified.

GammaVel:
This is working really nicely now, thanks @Auriga2 for correcting the way this works. I can import different file formats (Excel, CSV) and DB connection and the timezone conversions are available. This is a really nifty feature and great to see in Constellation. Thanks again

PR #1705, opened 2022-05-16
Add option to specify input timezone for Datetime formatters in Importers

Auriga2:
### Prerequisites

### Description of the Change
Users can specify the input time zone used in the Datetime fields in imported data files. 

Right-click on the `Datetime` transaction attribute and navigate to `Formatter` -> `Datetime` then select the `Time Zone`. 

It'll be empty by default which will display the formatted columns as if they are in GMT, similar to the previous behavior.
image

When a value is selected:
image

### Alternate Designs
n/a

Better ux

### Benefits

This is specially useful when Excel files are imported where the Excel datetime number doesn't contain time zone information.

### Verification Process

1. Open Constellation and the File Importer
2. Load a file in to the importer containing date time columns of different formats
3. Right-click on the Datetime transaction attribute
4. Navigate to Formatter -> Datetime
5. Select different formats and time zones and verify the datetime values in the table format correctly.
6. Click `Import button and check the date time column is imported in the graph with the specified time zone.
7. Test a custom format with the time zone and see if it retains when the  time zone field is left blank but overwrites when a value is specified.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1673

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -189,9 +189,9 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
         final String timeZone = timeZoneParam.getStringValue();
         // ZonedDateTime.parse requires a time zone identifier in the string (`value`)
         // hence zonedDateTime = ZonedDateTime.parse(value, df); doesn't work for all other formats
+        if (timeZoneParam.isEnabled() && StringUtils.isNotBlank(timeZone)) {
             ZonedDateTime dateTimeInSpecifiedTimeZone = ZonedDateTime.parse(value, df.withZone(TimeZone.getTimeZone(StringUtils.substringBetween(timeZone, "[", "]")).toZoneId()));
             return TemporalFormatting.ZONED_DATE_TIME_FORMATTER.format(dateTimeInSpecifiedTimeZone);
-        if (timeZoneParam.isEnabled() && StringUtils.isNotBlank(timeZone)) {
         } else {
             final TemporalAccessor temporalAccessorDateTime = df.parse(value);
             return TemporalFormatting.formatAsZonedDateTime(temporalAccessorDateTime);
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -37,6 +37,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.regex.Pattern;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import org.apache.commons.lang3.StringUtils;
@@ -117,6 +118,16 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
             if (change == ParameterChange.VALUE) {
                 final PluginParameter<?> slave = params.get(CUSTOM_PARAMETER_ID);
                 slave.setEnabled(master.getStringValue().equals(CUSTOM));
+
+                //disable Time Zone if the Datettime Format contains a Time Zone
+                params.get(TIMEZONE_PARAMETER_ID).setEnabled(!Pattern.matches(".*[XxZzOV']$", master.getStringValue()));
+            }
+        });
+
+        parameters.addController(CUSTOM_PARAMETER_ID, (final PluginParameter<?> master, final Map<String, PluginParameter<?>> params, final ParameterChange change) -> {
+            if (change == ParameterChange.VALUE) {
+                //disable Time Zone if the Custom Format contains a Time Zone
+                params.get(TIMEZONE_PARAMETER_ID).setEnabled(!Pattern.matches(".*[XxZzOV']$", master.getStringValue()));
             }
         });
 
@@ -164,7 +175,7 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
     }
     
     private String translateFromTemporalAccessorDateTime(final String value, final String format, final String timeZone){
-         final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);
+        final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);
         // ZonedDateTime.parse requires a time zone identifier in the string (`value`)
         // hence zonedDateTime = ZonedDateTime.parse(value, df); doesn't work for all other formats
         if (!StringUtils.isBlank(timeZone)) {
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
@@ -138,7 +138,7 @@ public class DatetimeAttributeTranslator extends AttributeTranslator {
         params.get(TIMEZONE_PARAMETER_ID).setEnabled(!containsTimeZone(format) && !format.equals(EPOCH));
     }
 
-    private boolean containsTimeZone(String format){
+    private boolean containsTimeZone(final String format){
         return Pattern.matches(".*[XxZzOV']$", format);
     }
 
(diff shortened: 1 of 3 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -104,36 +128,59 @@ public String translate(final String value, final PluginParameters parameters) {
...
-
-            return TemporalFormatting.formatAsZonedDateTime(dateTime);
-
         } catch (final DateTimeException | IllegalArgumentException ex) {
             return "ERROR";
         }
     }
 
+    private String translateFromZonedDateTime( final ZonedDateTime zonedDateTime, final String timeZone){
+        if (!StringUtils.isBlank(timeZone)) {
```
antares1470, line 158: `StringUtils.isNotBlank`

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -104,36 +128,59 @@ public String translate(final String value, final PluginParameters parameters) {
...
+        } else {
+            return TemporalFormatting.formatAsZonedDateTime(zonedDateTime);
+        }
+    }
+    
+    private String translateFromTemporalAccessorDateTime(final String value, final String format, final String timeZone){
+         final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);
+        // ZonedDateTime.parse requires a time zone identifier in the string (`value`)
+        // hence zonedDateTime = ZonedDateTime.parse(value, df); doesn't work for all other formats
+        if (!StringUtils.isBlank(timeZone)) {
```
antares1470, line 170: same here

antares1470 (changes requested):
Few things I found from testing:
1) Given this is just changing the actual timezone as appose to converting to a timezone, I wonder whether this is actually addressing #1698 instead. #1673 is more about converting the datetime to the preferred timezone as apposed to specifying the timezone itself. If it is addressing #1698, I think it would be good if possible to disable the timezone field when you select a datetime format that already contains one

2) I've noted that when going through the list of timezones that a) they weren't all in timezone order (started out fine but got a bit muddled in the middle), and b) some of the zones didn't match up (e.g. had +5:00 (GMT -5) and vise-versa)

3) Converting timezones after applying the attributes appears to have some issues. I'll update this with a gif demonstrating but basically I changed to a timezone of +0:00 and it applied in one row as -1:00 and another row as -1:54 (both of which are obviously wrong)

Auriga2:
@antares1470 For (1) you said you'll let me know whether we still need a time zone converter in the importers because we can already convert them in the Attribute editor. There we can convert on one transaction or a bulk selected (different times will be replaced with the time you enter-so not really useful?) , but couldn't fine a global time zone setting in Graph attributes as @arcturus2 mentioned. (Could it be available somewhere else?)

antares1470:
> @antares1470 For (1) you said you'll let me know whether we still need a time zone converter in the importers because we can already convert them in the Attribute editor. There we can convert on one transaction or a bulk selected (different times will be replaced with the time you enter-so not really useful?) , but couldn't fine a global time zone setting in Graph attributes as @arcturus2 mentioned. (Could it be available somewhere else?)

@Auriga2 so there isn't a global timezone graph attribute. The datetimes and their timezones are all local to the node/transaction. Yes you can convert to the desired timezone in the attribute editor, but I still think it would be good to include an option to convert the timezone in the importer as I describe in #1673. That though is more a quality-of-life improvement (so that you don't have to then go to the attribute editor afterwards). Your current solution here though definitely seems more like it is aiming to fix #1698 which is focused on ensuring data with no timezone specified can still be correctly imported.

Auriga2:
Agree that this is fixing the https://github.com/constellation-app/constellation/issues/1698. Updated the description here. I'll add a separate PR for https://github.com/constellation-app/constellation/issues/1673

Auriga2:
These are the possible formats with time zones:
The Z (Zulu) on the end means UTC. that is, an offset-from-UTC of zero hours-minutes-seconds. 
 E.g.
"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"(eg  "2011-08-12T20:17:46.384Z")
 yyyy-MM-dd'T'HH:mm:ss'Z'(eg "2015-09-29T18:46:19Z")
 
These as a custom format, but I couldn't find a proper format string to get them formatted in consty yet.
 YYYY-MM-DDThh:mm+hh:mm (eg 1997-07-16T19:20+01:00)
YYYY-MM-DDThh:mm:ss+hh:mm (eg 1997-07-16T19:20:30+01:00)
YYYY-MM-DDThh:mm:ss.s-hh:mm (eg 1997-07-16T19:20:30.45-01:00)

 YYYY-MM-DD'T'HH:mm:ss Z ?(eg  2016-06-10T13:53:22 +0200)

Auriga2:
Looks like there could be many time zone formats the [DateTimeFormatter ](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#:~:text=Offset%20X%20and%20x%3A%20This,%2C%20such%20as%20'%2B0130') can handle. I could see below formats working in consty. 
yyyy-MM-dd'T'HH:mmXXX (eg 1999-07-16T19:20-03:30)
yyyy-MM-dd HH:mm:ss.SSS XXX (eg 2020-05-09 22:14:14.037 +09:00)
yyyy-MM-dd'T'HH:mm:ssXX (eg 2016-06-10T13:53:22+0400)

I'll add a check assuming any format with trailing **X** and **'Z'** as containing time zone information.
Should probably also add trailing lowercase **x**, **O, V**

image
image

antares1470 (changes requested):
A few things I found (on top of the code comments above yet to be changed):

1) Epoch is measured from GMT and so should disable the timezone parameter when selected
2) When the datetime attribute is applied, there is a check for whether the timezone parameter is empty, but no check if it is disabled. This means currently that if you select a value in the parameter before it gets disabled and then apply that attribute, it will still use the timezone value picked (disabling doesn't remove the value, simply removes the ability to change it)

Side note: If you're also able to include a whatsnew for this, that would be great

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -116,69 +118,78 @@ public PluginParameters createParameters() {
...
         return parameters;
     }
 
+    private void enableDisableTimeZone(final PluginParameter<?> master, final Map<String, PluginParameter<?>> params){
+        // Disables the Time Zone field if the Date time Format or Custom Format contains a Time Zone while it's not an EPOCH time
+        final String format =  master.getStringValue();
+        params.get(TIMEZONE_PARAMETER_ID).setEnabled(!containsTimeZone(format) && !format.equals(EPOCH));
+    }
+
+    private boolean containsTimeZone(String format){
```
antares1470, line 141: parameter final

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -116,69 +118,78 @@ public PluginParameters createParameters() {
...
         } catch (final DateTimeException | IllegalArgumentException ex) {
             return "ERROR";
         }
     }
 
-    private String translateFromZonedDateTime( final ZonedDateTime zonedDateTime, final String timeZone){
-        if (!StringUtils.isBlank(timeZone)) {
+    private String translateFromZonedDateTime( final ZonedDateTime zonedDateTime, final PluginParameter timeZoneParam){
+        final String timeZone = timeZoneParam.getStringValue();
+        if ( timeZoneParam.isEnabled() && !StringUtils.isBlank(timeZone)) {
```
antares1470, line 179: `!StringUtils.isBlank(timeZone)` -> `StringUtils.isNotBlank(timeZone)`

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/translator/DatetimeAttributeTranslator.java
```diff
@@ -116,69 +118,78 @@ public PluginParameters createParameters() {
...
     }
     
-    private String translateFromTemporalAccessorDateTime(final String value, final String format, final String timeZone){
+    private String translateFromTemporalAccessorDateTime(final String value, final String format, final PluginParameter timeZoneParam){
         final DateTimeFormatter df = DateTimeFormatter.ofPattern(format);
+        final String timeZone = timeZoneParam.getStringValue();
         // ZonedDateTime.parse requires a time zone identifier in the string (`value`)
         // hence zonedDateTime = ZonedDateTime.parse(value, df); doesn't work for all other formats
-        if (!StringUtils.isBlank(timeZone)) {
+        if (timeZoneParam.isEnabled() && !StringUtils.isBlank(timeZone)) {
```
antares1470, line 192: same here

antares1470 (changes requested):

antares1470 (approved):

mimosa2 (approved):

Delphinus8821 (approved):

PR #1712, opened 2022-06-02
fixed issue with some default parameter values not triggering controllers

antares1470:
### Prerequisites

### Description of the Change

A couple of plugins had a default value added to one of the parameters. The problem was that changing the value in these parameter was meant to trigger a controller but since the default value was set before the controller was defined, the expected actions didn't happen (e.g. other parameters in the plugin being enabled or disabled). This small change just moves the setting of those default parameter values to after the controller definition so that the controller code gets triggered.

Ensures plugins in Core follow expected behaviour

### Benefits

parameters don't have to be re-selected to trigger intended behaviour

### Possible Drawbacks

Nil

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
@@ -140,7 +140,6 @@ public class MergeNodesPlugin extends SimpleQueryPlugin implements DataAccessPlu
         mergeType.setRequired(true);
         final List<String> mergeTypes = new ArrayList<>(MERGE_TYPES.keySet());
         SingleChoiceParameterType.setOptions(mergeType, mergeTypes);
-        SingleChoiceParameterType.setChoice(mergeType, mergeTypes.get(0));
         params.addParameter(mergeType);
 
         final PluginParameter<IntegerParameterValue> threshold = IntegerParameterType.build(THRESHOLD_PARAMETER_ID);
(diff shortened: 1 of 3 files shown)


Auriga2:
Looks good in the Merge nodes and Transactions plugins. The default threshold value in Merge nodes plugin is 1000 while  it's 0 in Merge Transactions plugin. (In v2.7.rc1 both are 0.) Should we make the latter a realistic number or make both 0? 

Could there be other plugins affected, because we set values before defining the controller everywhere?

Also get this exception in both versions if you delete the threshold value and click elsewhere. Get the same every time the up and down arrows are clicked afterwards.
image

Entering a number after that throws this, but goes away if you enter a number again, 
image

antares1470:
> Looks good in the Merge nodes and Transactions plugins. The default threshold value in Merge nodes plugin is 1000 while it's 0 in Merge Transactions plugin. (In v2.7.rc1 both are 0.) Should we make the latter a realistic number or make both 0?

@Auriga2 the threshold is set to 1000 when the merge by parameter is set to Geospatial Distance. The reason it was 0 before that was because there was no default value explicitly set for the parameter (so default value for integer parameters is 0). So I think these values are correct given the what merge by values are being set by default

> Could there be other plugins affected, because we set values before defining the controller everywhere?

The answer is possible. I did a check of other plugins that I marked with required parameters (and set default values for) and these were only the ones I noticed required the update. An easy way to check is to set the parameter to the same value and see whether anything changes. I'm fairly confident I've got most if not all of them.

> Also get this exception in both versions if you delete the threshold value and click elsewhere. Get the same every time the up and down arrows are clicked afterwards.
> 
> Entering a number after that throws this, but goes away if you enter a number again,

This is an existing issue with the NumberSpinner FX object. You'll find that you'll get the same thing occurring in versions even earlier than 2.7-rc1. I tried to fix it on Constellation side but was ultimately unsuccessful so it will have to be something fixed on the FX side

sol695510 (approved):

Auriga2 (approved):
Thanks for the explanation.

Issue #1718, opened 2022-06-27
Consolidate Import from GraphML, GML, Pajek into one Plugin in Core


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Presently in Constellation Adaptors there are a few Data Access plugins for working with different graph file types including GraphML, GML, and Pajek. The aforementioned types though appear to be open standards and so these might instead make more sense to be be in Core. What I'm going to propose therefore is 2 things:
1) Import from GraphML, Import from GML, and Import from Pajek be moved to Core
2) Those plugins then be merged together into one "Import Graph File" plugin

The drawback with merging them into one plugin is that you won't be able to import different types of graph files in the same step (instead you'll need to be run the same plugin across multiple steps) but depending on how it is refactored, I think it could really help with extensibility (e.g. could use a lookup and hence include other graph file types not just in Core) and maintainability.

### Additional Information

Note that I've only mentioned the Import plugins and not the Enrich or Extend variants which are also available in Adaptors. I believe in the general case you probably prefer to just bring in a whole file rather than just part of it, and to my understanding those Enrich and Extend versions are really there to demonstrate what Constellation can do (since such plugins don't really exist in Core).

GammaVel:
Thanks @antares1470 for adding in handy import from file functionality. I have tried this out with GML, GraphML and Pajek (*.net) and it is working. The help doco has been updated too which is great!

PR #1730, opened 2022-07-18
Import Graph File

antares1470:
### Prerequisites

### Description of the Change

Added a new Data Access Plugin called "Import Graph File" which allows you to import different kinds of graph files previously made available through Constellation Adaptors.

There were a few bug I identified with the implementations of classes I had moved from adaptors so I fixed those at the same time.

Being able to import graph files should be a core part of a graph analysis application. This has also been written in such a way that the plugin can be extended with other kinds of graph file types that people may want to add to their own versions.

### Benefits

Ability to import other graph file types now part of Core

### Possible Drawbacks

GML, GraphML, and Pajek were previously each their own plugin in Constellation adaptors which meant you could import a graph of each type in one step. Consolidating these into one plugin does mean now that the same process requires 3 data access steps to import each graph. Long term though I think this is the right move as it gives future graph file type implementations a framework to work with. Also I could be wrong, but I suspect there won't be too many cases of wanting to import multiple of these file types at the same time. 

### Applicable Issues

#1718


Delphinus8821:
Reviewing this PR and it looks good to me. In testing locally it looks like it's generating a new entry in the toc.md for the new help page that's been created, would you be able to update the PR to include that change to the toc.md?

Delphinus8821 (approved):

mimosa2 (approved):

Issue #1249, opened 2021-08-19
Notes View - Suggestion to disable "Add" and "Remove" selected elements when note is applied to the whole graph


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

On the Notes view I can add User notes to an entire graph or an individual selection of Node and Transactions. If I add a note to the entire graph and Click Edit I am presented with three options to Keep, Add or Remove.  As an improvement it would be good if the options to Add or Remove are greyed out or hidden as the Note already applies to the entire graph so you can't really add to the selection. If I do try and remove a node from the selection, the note remains for the entire graph and if I click on "Select on graph" I can see that the node has not been removed from the Note. 

### Steps to Reproduce

1.  From the Notes View add a Note to the entire graph. 
2. Select one of the nodes on the graph
3.  Right click the note and select Remove Selected Elements
4. Navigate back to the note and R mouse click to Select on graph > Expecting node is not included in the selection

**Expected behaviour:** The options to Remove and Add don't get shown for a note that applies to the whole graph so I don't see the issue if I try to remove a node. 

**Actual behaviour:** \The node is still part of the Note as it applies to all of the graph. 

**Reproduces how often:**100%

### Additional Information
![image](https://user-images.githubusercontent.com/83679004/129995371-8d4dff08-1392-49dd-ad90-878b33c554dc.png)
new release: 
![image](https://user-images.githubusercontent.com/83679004/134610803-4a8e3486-df60-4854-8df6-4fd59d6998ab.png)

GammaVel:
Hi Constellation developers!  I updated my screenshot here as I noticed the functionality has changed with the latest build. This is still an issue, I can select to Add or Remove to a note that applies to the whole graph but it doesn't actually do anything. It would be great if this could be fixed.

Atlas139mkm:
[...]

PR #1743, opened 2022-07-29
Notes View "Remove Selected" Menu Item Disabled

altair1673:
### Prerequisites

### Description of the Change
When a note is not associated with any vertexes or transactions the "Remove Selected" option is the context menu is disabled.

Reduces confusion on notes associated with the whole graph when the "remove selected" option does nothing.

### Benefits
Improves suability

### Verification Process
When no graph elements are associated with note the "Remove Selected" option is disabled. When I add elements to the note the option becomes enabled again.

### Applicable Issues
#1249

### DIFF
--- a/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
+++ b/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
@@ -42,7 +42,6 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import javafx.application.Platform;
 import javafx.collections.FXCollections;
@@ -489,7 +488,6 @@ public class NotesViewPane extends BorderPane {
 
             synchronized (LOCK) {
                 notesViewEntries.forEach(entry -> {
-
                     // Add note to render list if its respective filter is selected.
                     if ((selectedFilters.contains(USER_NOTES_FILTER) && entry.isUserCreated())) {
                         notesToRender.add(entry);
@@ -503,7 +501,6 @@ public class NotesViewPane extends BorderPane {
                         // If no nodes or transactions are selected, show notes applied to the whole graph.
                         if (entry.isGraphAttribute()) {
                             notesToRender.add(entry);
-
                         }
                         // Show notes related to the selected nodes.
                         for (final int node : nodesSelected) {
@@ -772,7 +769,6 @@ public class NotesViewPane extends BorderPane {
                 newNote.setNoteContent(contentTextArea.getText());
 
                 removeFromSelectedElements(newNote);
-
                 final Graph activeGraph = GraphManager.getDefault().getActiveGraph();
                 if (activeGraph != null) {
                     updateNotesUI();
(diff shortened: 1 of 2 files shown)


antares1470 (changes requested):
Are you also able to disable the add selected option as well? When a note applies to the graph, any subsequently added nodes and transaction are added to the note, and hence add selected is also redundant

Delphinus8821:
> Are you also able to disable the add selected option as well? When a note applies to the graph, any subsequently added nodes and transaction are added to the note, and hence add selected is also redundant

@antares1470 I believe with the way it was written, if a note was applied to the graph and then the user added nodes or transactions to it, it changed the note from being one applied to the graph to being one only applied to those newly added selected elements.

antares1470:
> @antares1470 I believe with the way it was written, if a note was applied to the graph and then the user added nodes or transactions to it, it changed the note from being one applied to the graph to being one only applied to those newly added selected elements.

So I can't recall exactly how it's being stored under the hood but if I select the whole graph and then add a user note, if it is applied to selected it specifies as being applied to x nodes and y transactions and if not it specifies being applied to the graph. If I then add a node to the graph, the note that applies to the selection won't include this new node whereas the one applied to the graph does. See the gif below demonstrating this:

notesViewSelectOnGraph

Delphinus8821:
> > @antares1470 I believe with the way it was written, if a note was applied to the graph and then the user added nodes or transactions to it, it changed the note from being one applied to the graph to being one only applied to those newly added selected elements.
> 
> @Delphinus8821 So I can't recall exactly how it's being stored under the hood but if I select the whole graph and then add a user note, if it is applied to selected it specifies as being applied to x nodes and y transactions and if not it specifies being applied to the graph. If I then add a node to the graph, the note that applies to the selection won't include this new node whereas the one applied to the graph does. See the gif below demonstrating this:
> 


@antares1470 This is a different case to when you use the 'Add Selected' button on the note. A note that is applied to a graph will always be applied to the whole graph but if the user uses the 'Add Selected' button, then the selection applied to the note updates. See the gif demonstration below:
NotesAdd

antares1470:
> @antares1470 This is a different case to when you use the 'Add Selected' button on the note. A note that is applied to a graph will always be applied to the whole graph but if the user uses the 'Add Selected' button, then the selection applied to the note updates.

@Delphinus8821 Ok I didn't realise that was a thing, my bad there.

That being said, I would argue that what is being demonstrated there isn't "Add Selected" but instead something like "Change to Selected". When I see "Add Selected", I think of appending to the existing selection for the note, rather than replacing. This is certainly the case where if you have a note that applies to 1 node and you then select a different node (with the node the note is applied to de-selected) and add selected, the note is now applied to 2 nodes. So in the case of a note being applied to the whole graph, there isn't anything you can append to the note that isn't already applied.

Consequently, I actually think "Add Selected" for the note applied to the whole graph is misleading and therefore should still be disabled. There may be an argument for including an option to change the selection instead but I note that currently in master that while "Add Selected" will change the note selection to the selected graph elements, "Remove Selected" doesn't do the opposite, i.e. change the selection to everything bar the selected graph elements (presumably because of the way a note is stored).

antares1470 (approved):

Delphinus8821 (approved):

PR #1778, opened 2022-11-07
fixed potential issues with workflow plugins

antares1470:
### Prerequisites

### Description of the Change

Workflow Query plugins are a great tool in Constellation to run multiple plugins sequentially without having to execute each plugin individually. Looking at the code though, it appears that a lot of the logic in and around `WorkflowQueryPlugin` only really works for running node-querying plugins rather than transaction-querying plugins, and so will present issues when you want to include a plugin of the latter style in your workflow. This PR adds some changes to allow transactions to be properly considered when running a workflow plugin.

An issue was also identified in which the batch subgraphs being created by the workflow plugin wasn't always applying results correctly back to the original graph due to the way graph element id's were being created and handled in the subgraphs vs the original. This PR also corrects that.

Fixes issues with workflow plugins

### Benefits

Bugfixes, workflow plugins can now be utilised with more plugins

### Possible Drawbacks

Nil


arcturus2 (approved):
Nice!

Issue #1695, opened 2022-05-05
New Find screens - Improve "Current Selection" option, consider changing to post find actions and add delete option


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The New Find (which is available under Experimental) introduces a new option for Current Selection. This replaces the previous tick box which was used for "Add results to the current selection". In the new Find we have Current Selection with the options of Ignore, Add To, Find In and Remove from.
![image](https://user-images.githubusercontent.com/83679004/166853648-672acffd-36ba-45b6-831d-776fd76525aa.png)

When I first started to use the new Find one of the things I found most confusing was "Ignore". I wasn't sure what it meant and why I would need to ignore my current selection.  Also, with the default set to Ignore it makes it seem like the Current Selection should not be used to find in (ie - I am searching outside of the current selection). This is not the case, when it is set to Ignore it doesn't pay any attention to the selection. 

As this can easily trip people up, it might be better to have this section divided up into a Search Options/Post Actions with more options shown when it is expanded.  We also need to have an option to delete the search item from the graph.  
 Search options could be set to:  Selection, Graph, All open graphs. Post actions could be: Add to current selection, Remove, Delete. This could be similar to excel where initially there is a very streamlined screen with minimal options.

This is open for discussion and it would be great to see some different mock designs.

### Steps to Reproduce

1.  Open up the new Find screen
2.  Attempt to find information > confused by the ignore option which is shown in the new option

**Expected behaviour:** More user friendly options here

**Actual behaviour:**Initially I was unsure how to make use of the Ignore Option though I did get used to it after a while. 

### Additional Information
Just wondering if there would be a better word here that would be less confusing, or if Ignore could be changed to a blank. 
![image](https://user-images.githubusercontent.com/83679004/165403834-c285aabe-a924-450c-8c81-d2904200041d.png)

GammaVel:
[...]

PR #1799, opened 2023-01-15
Improve Current Selection in New Find View

Delphinus8821:
### Prerequisites

### Description of the Change
Updated the current selection drop down list in the new find view and split it into two drop downs, search in and post results action. The search in drop down shows where the search will be performed, e.g. the current selection, current graph or all open graphs. The post search actions drop down has the option to choose whether the search will replace the current selection, add to or remove from the current selection. 

A delete results button was also added to delete results from the graph quickly. It only becomes available after a find all search as it will delete all results found. After pressing the delete results button, the user is then displayed a dialog box which shows how many results will be deleted across x graphs and whether they wish to continue deleting these. 
The delete results option was originally going to be added as part of the Post Search Actions drop down but it was discovered that unless the search with the same results had already been made using a different option in the Post Search Actions drop down, completing the delete results option would cause Constellation to become unresponsive. 

These new options appear in both the basic find and the advanced find. 

### Alternate Designs
N/A

Improves the usability of the new find view by expanding the different options. Also adds new functionality of being able to delete results from within the find view. 

### Benefits
Easier use of the new find view. 

### Verification Process
Tested all options in both drop down lists on both the basic find and the advanced find to ensure all functionality still works as intended. 
Also tested the delete option on both tabs. 

### Applicable Issues
#1695


CoreFindView/src/au/gov/asd/tac/constellation/views/find2/components/BasicFindTab.java
```diff
@@ -140,19 +143,29 @@ public BasicFindTab(final FindViewTabs parentComponent) {
...
+        searchInChoiceBox.getSelectionModel().selectedItemProperty().addListener(new ChangeListener<String>() {
             @Override
             public void changed(ObservableValue<? extends String> observableValue, String oldElement, String newElement) {
                 updateSelectionFactors();
+                updateBasicFindParamters();
+            }
+        });
+
+        // set the action for changing the seleciton in the postSearchChoiceBox
+        postSearchChoiceBox.getSelectionModel().selectedItemProperty().addListener(new ChangeListener<String>() {
```
antares1470, line 156: can the listener here and the one above be converted to a lambda? Or is there an issue with doing that? (If there's a problem, make the parameters final in the function)

CoreFindView/src/au/gov/asd/tac/constellation/views/find2/plugins/BasicFindPlugin.java
```diff
@@ -196,8 +194,10 @@ protected void edit(final GraphWriteMethods graph, final PluginInteraction inter
         foundResult.clear();
         foundResult.addAll(distinctValues);
 
+        // If the results list is null, has different parameters to the current list, and is not searching all graphs, then create a new results list
         if (ActiveFindResultsList.getBasicResultsList() == null || !ActiveFindResultsList.getBasicResultsList().getSearchParameters().equals(this.parameters)
-                || (!this.parameters.isSearchAllGraphs() && ActiveFindResultsList.getBasicResultsList().get(0) != null && !ActiveFindResultsList.getBasicResultsList().get(0).getGraphId().equals(graph.getId()))) {
+                || (!this.parameters.isSearchAllGraphs() && ActiveFindResultsList.getBasicResultsList() == null && ActiveFindResultsList.getBasicResultsList().get(0) != null
```
antares1470, line 199: `ActiveFindResultsList.getBasicResultsList() == null && ActiveFindResultsList.getBasicResultsList().get(0) != null` would appear to cause an issue (if the 1st is true, the 2nd will throw a null pointer)

CoreFindView/src/au/gov/asd/tac/constellation/views/find2/plugins/advanced/AdvancedSearchPlugin.java
```diff
@@ -269,8 +271,10 @@ protected void edit(GraphWriteMethods graph, PluginInteraction interaction, Plug
         foundResult.clear();
         foundResult.addAll(distinctValues);
 
+        // If the results list is null, has different parameters to the current list, and is not searching all graphs, then create a new results list 
         if (ActiveFindResultsList.getAdvancedResultsList() == null || !ActiveFindResultsList.getAdvancedResultsList().getAdvancedSearchParameters().equals(this.parameters)
-                || (!this.parameters.isSearchAllGraphs() && ActiveFindResultsList.getAdvancedResultsList().get(0) != null && !ActiveFindResultsList.getAdvancedResultsList().get(0).getGraphId().equals(graph.getId()))) {
+                || (!this.parameters.getSearchInLocation().equals(ALL_OPEN_GRAPHS) && ActiveFindResultsList.getAdvancedResultsList() == null
+                && ActiveFindResultsList.getAdvancedResultsList().get(0) != null && !ActiveFindResultsList.getAdvancedResultsList().get(0).getGraphId().equals(graph.getId()))) {
```
antares1470, line 277: Similar here

antares1470 (changes requested):
Some other things from testing:
- When the post-action option is Replace Selection, Find All doesn't highlight any of the found terms and Find Next and Previous append to the selection
- In the Replace tab I'm a bit confused to what "Replace In" means. Replace in this case also seems to have 2 different meanings in this tab (with respect to text and with respect to selection) which I think might be confusing
- I would possibly look at the Delete Results button being a Post-search action option instead. I do like the inclusion of the prompt to check beforehand but I think it doesn't really make sense to be a separate button.
- The term "Delete Results" can also be a little confusing. When I first saw it, I interpreted it as clearing the search rather than deleting things from the graph

antares1470 (changes requested):
Given there is only one option now for the Post actions in the Replace tab, is it possible for that dropdown to be removed?

antares1470 (approved):

mimosa2:
I tested Basic Find with All Open Graphs and Find All picked up all the hits but I had the default Post-Search Action option as Add To Selection. This option didn't enable Find Prev and Find Next although there were results in the other graphs (they all high lighted correctly).
 I had the similar situation the when the default was set to Remove from Selection.  Is it possible to make no default option for this drop down ( Post-Search Action)?

mimosa2:
After discussed with @Delphinus8821 and @antares1470  I understood that The Find Next & Previous are not meant to be enabled for Add To or Remove From options.

mimosa2 (approved):

Issue #653, opened 2020-07-09
Thread.sleep() called within loop


aldebaran30701:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Calling sleep within a loop can cause performance issues, and maybe a redesign of the classes utilizing this may be necessary.
if the implementation is waiting for another thread to fill a buffer, a blocking queue could be used to free up CPU.

The following modules include usages:
Attribute editor - top component 
Data access view
Graph node
Interactive Graph
Open GL Display
Plugin Reporter View
Table View
Testing
Web server

### Additional Information

These implementations may not be able to be improved on, so an individual analysis may be required on each case.

aldebaran30701:
Looking into CompletableFutures might help with knowing when to proceed with an action instead of sleeping the current thread.
The data access view and table view have been rewritten, so might not need this fixed on some sections.

PR #1811, opened 2023-03-08
Removed some thread.sleeps from WebServer and Attribute Editor package

altair1673:
### Prerequisites

- [ X] Reviewed the [checklist](CHECKLIST.md)

### Description of the Change

The thread.sleep() in the attribute editor has been removed since a similar effect can be reached without it. The RestServiceUtilities class has had a thread.sleep replaced with a CompletableFuture.

After some investigation:

Data access view - the thread.sleep() here seems to be an intentional thing where the user can choose to sleep the thread for a certain period of time.

Graph Node - It is unclear how to make this any better. The sleep is in the DefaultPluginInteraction class

Core Interactive Graph - The sleep is in the Animation class. The sleep is used pause the thread after each frame of the animation is displayed. Since each animation runs on its own thread I don't think the sleep affects performance much since nothing else will be running on that thread.

Open GL display - Cannot find the Thread.sleep() in this

Plugin Reporter View - I did draft some changes for this where a sleep isn't used but they were all less efficient than using the sleep. When I find a better solution I will comeback for this. The sleep is in the PluginReportTimeUpdater class.

Table View - Unclear on how to improve things, seems like the last guy was unsure about it as well. This one is in the top component.

Testing - The thread.sleeps here are not directly inside a loop so they are probably being called from a loop somewhere else. When searching for the usages of the function containing the sleeps no results are found. I have been told that these are used with experimental stuff and is important to keep.

Web server - this looks like it can be improved. The sleep is in the FileListener class. Need to learn more about it.

Small performance benefit
### Benefits

Unsure if user will feel things since so little has changed but theoretically AttributeEditor and RestServiceUtilities should perform better
### Possible Drawbacks

Things become slower and I have to start from square one again.
### Verification Process

Still need to verify. This one will probably need a few more go's.
### Applicable Issues

#653

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
@@ -24,9 +24,7 @@ import au.gov.asd.tac.constellation.graph.node.GraphNode;
 import au.gov.asd.tac.constellation.preferences.utilities.PreferenceUtilities;
 import au.gov.asd.tac.constellation.views.JavaFxTopComponent;
 import java.util.ArrayList;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.LinkedBlockingQueue;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.prefs.PreferenceChangeListener;
 import java.util.prefs.Preferences;
@@ -104,8 +102,6 @@ import org.openide.windows.TopComponent;
     "HINT_AttributeEditorTopComponent=Attribute Editor"
 })
 public final class AttributeEditorTopComponent extends JavaFxTopComponent<AttributeEditorPanel> implements GraphManagerListener, GraphChangeListener, UndoRedo.Provider, PreferenceChangeListener {
-    
-    private static final Logger LOGGER = Logger.getLogger(AttributeEditorTopComponent.class.getName());
 
     private static final String ATTRIBUTE_EDITOR_GRAPH_CHANGED_THREAD_NAME = "Attribute Editor Graph Changed Updater";
     private final AttributeEditorPanel attributePanel;
@@ -127,13 +123,13 @@ public final class AttributeEditorTopComponent extends JavaFxTopComponent<Attrib
 
             final ArrayList<Object> devNull = new ArrayList<>();
 
-            while (queue.size() > 0) {
-                    queue.drainTo(devNull);
+            while (!queue.isEmpty()) {
+                queue.drainTo(devNull);
             }
 
-                if (reader != null) {
-                    attributePanel.updateEditorPanel(reader.refreshAttributes());
-                }
+            if (reader != null) {
+                attributePanel.updateEditorPanel(reader.refreshAttributes());
+            }
 
         };
 
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
@@ -124,29 +124,20 @@ public final class AttributeEditorTopComponent extends JavaFxTopComponent<Attrib
         setToolTipText(Bundle.HINT_AttributeEditorTopComponent());
 
         refreshRunnable = () -> {
-            try {
+
                 final ArrayList<Object> devNull = new ArrayList<>();
 
-                while (queue.drainTo(devNull) > 0) {
-                    Thread.sleep(50);
-                }
 
-                /*while (queue.size() > 0) {
+            while (queue.size() > 0) {
                     queue.drainTo(devNull);
-                }*/
+            }
 
                 if (reader != null) {
                     attributePanel.updateEditorPanel(reader.refreshAttributes());
                 }
-            } catch (final InterruptedException ex) {
-                LOGGER.log(Level.SEVERE, "Thread was interrupted");
-                Thread.currentThread().interrupt();
-            }
+
         };
 
-        CompletableFuture<void> future = CompletableFuture.runAsync(() -> {
-
-        });
 
         GraphManager.getDefault().addGraphManagerListener(AttributeEditorTopComponent.this);
         newActiveGraph(GraphManager.getDefault().getActiveGraph());
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
@@ -24,6 +24,7 @@ import au.gov.asd.tac.constellation.graph.node.GraphNode;
 import au.gov.asd.tac.constellation.preferences.utilities.PreferenceUtilities;
 import au.gov.asd.tac.constellation.views.JavaFxTopComponent;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.logging.Logger;
 import java.util.prefs.PreferenceChangeListener;
@@ -121,7 +122,7 @@ public final class AttributeEditorTopComponent extends JavaFxTopComponent<Attrib
 
         refreshRunnable = () -> {
 
-            final ArrayList<Object> devNull = new ArrayList<>();
+            final List<Object> devNull = new ArrayList<>();
 
             while (!queue.isEmpty()) {
                 queue.drainTo(devNull);
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
@@ -24,6 +24,7 @@ import au.gov.asd.tac.constellation.graph.node.GraphNode;
 import au.gov.asd.tac.constellation.preferences.utilities.PreferenceUtilities;
 import au.gov.asd.tac.constellation.views.JavaFxTopComponent;
 import java.util.ArrayList;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -125,9 +126,14 @@ public final class AttributeEditorTopComponent extends JavaFxTopComponent<Attrib
         refreshRunnable = () -> {
             try {
                 final ArrayList<Object> devNull = new ArrayList<>();
+
                 while (queue.drainTo(devNull) > 0) {
                     Thread.sleep(50);
                 }
+
+                /*while (queue.size() > 0) {
+                    queue.drainTo(devNull);
+                }*/
 
                 if (reader != null) {
                     attributePanel.updateEditorPanel(reader.refreshAttributes());
@@ -137,6 +143,10 @@ public final class AttributeEditorTopComponent extends JavaFxTopComponent<Attrib
                 Thread.currentThread().interrupt();
             }
         };
+
+        CompletableFuture<void> future = CompletableFuture.runAsync(() -> {
+
+        });
 
         GraphManager.getDefault().addGraphManagerListener(AttributeEditorTopComponent.this);
         newActiveGraph(GraphManager.getDefault().getActiveGraph());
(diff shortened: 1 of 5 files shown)


CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,7 +116,17 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+        final String newId;
+
+        try {
+            newId = RestServiceUtilities.waitForGraphChange(existingId).get(10, TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            throw new RestServiceException(ex);
+        } catch (ExecutionException ex) {
+            throw new RestServiceException(ex);
+        } catch (TimeoutException ex) {
+            throw new RestServiceException(ex);
```
Delphinus8821, line 128: finals on exceptions

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/OpenGraph.java
```diff
@@ -113,6 +120,12 @@ public void callService(final PluginParameters parameters, final InputStream in,
             mapper.writeValue(out, root);
         } catch (final GraphParseException | FileNotFoundException ex) {
             throw new RestServiceException(HTTP_UNPROCESSABLE_ENTITY, ex.getMessage());
+        } catch (InterruptedException ex) {
+            throw new RestServiceException(ex);
+        } catch (ExecutionException ex) {
+            throw new RestServiceException(ex);
+        } catch (TimeoutException ex) {
```
Delphinus8821, line 127: finals on exceptions

Delphinus8821 (changes requested):

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,13 +116,25 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+
+        if (!newId.isBlank() && !newId.isEmpty()) {
         final ObjectMapper mapper = new ObjectMapper();
         final ObjectNode root = mapper.createObjectNode();
         root.put("id", newId);
         root.put("name", GraphNode.getGraphNode(newId).getDisplayName());
         root.put("schema", schemaName);
-        mapper.writeValue(out, root);
+            mapper.writeValue(out, root);
+        }
```
Delphinus8821, line 138: Fix indentation here

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,13 +116,25 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+        String newId = "";
 
+        try {
+            newId = RestServiceUtilities.waitForGraphChange(existingId).get(10, TimeUnit.SECONDS);
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
+        } catch (final ExecutionException ex) {
+            throw new RestServiceException(ex);
+        } catch (final TimeoutException ex) {
+            throw new RestServiceException(ex);
```
OrionsGuardian, line 128: Can these be combined to a single catch ?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/OpenGraph.java
```diff
@@ -113,6 +120,12 @@ public void callService(final PluginParameters parameters, final InputStream in,
             mapper.writeValue(out, root);
         } catch (final GraphParseException | FileNotFoundException ex) {
             throw new RestServiceException(HTTP_UNPROCESSABLE_ENTITY, ex.getMessage());
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
+        } catch (final ExecutionException ex) {
+            throw new RestServiceException(ex);
+        } catch (final TimeoutException ex) {
+            throw new RestServiceException(ex);
```
OrionsGuardian, line 128: Combined into a single catch ?

OrionsGuardian (changes requested):

Delphinus8821:
I think this is good to go once the code smells get fixed.

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/OpenGraph.java
```diff
@@ -102,7 +105,11 @@ public void callService(final PluginParameters parameters, final InputStream in,
             final Graph g = new GraphJsonReader().readGraphZip(fnam, new HandleIoProgress(String.format("Loading graph %s...", fnam)));
             GraphOpener.getDefault().openGraph(g, name, false);
 
-            final String newId = RestServiceUtilities.waitForGraphChange(existingId);
+            final String newId;
+
+            newId = RestServiceUtilities.waitForGraphChange(existingId).get(10, TimeUnit.SECONDS);
```
Delphinus8821, line 110: Is there a reason for splitting this into two lines?

mimosa2 (approved):

OrionsGuardian (approved):

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestServiceUtilities.java
```diff
@@ -170,36 +171,22 @@ public static String activeGraphId() {
...
-            // We only have so much patience.
-            //
-            if (++waits > 10) {
-                throw new RestServiceException("The new graph has taken too long to become active");
+    public static CompletableFuture<String> waitForGraphChange(final String existingId) {
+
+        return CompletableFuture.supplyAsync(() -> {
+            while (true) {
+                if (GraphManager.getDefault().getActiveGraph() != null) {
+                    final Graph newGraph = GraphManager.getDefault().getActiveGraph();
```
antares1470, line 179: Are you possibly able to extract the `getActiveGraph` call into a separate variable and use that? This just ensures that both calls here are using the same graph (which appears to be what you want here, without it there is a small chance these would return different results if you so happen to manage to switch graphs between the 2 calls).

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestServiceUtilities.java
```diff
@@ -170,36 +171,22 @@ public static String activeGraphId() {
...
+                    final String newId = newGraph != null ? newGraph.getId() : null;
```
antares1470, line 180: the if statement this sits in already checks that the active graph (which is what `newGraph` is set to) is not null so this check shouldn't be necessary (especially if combined with the above comment)

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestServiceUtilities.java
```diff
@@ -170,36 +171,22 @@ public static String activeGraphId() {
...
+                    if ((existingId == null && newId != null) || (existingId != null && newId != null && !existingId.equals(newId))) {
+                        // - there was no existing graph, and the new graph is active, or
+                        // - there was an existing graph, and the active graph is not the existing graph.
+                        // - we assume the user hasn't interfered by manually switching to another graph at the same time.
```
antares1470, line 184: As explained in the earlier comment, I probably wouldn't assume the user won't manually switch for the duration of this call. Instead, implement my previous comments which will mitigate the consequences if that happens

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,13 +116,23 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
 
-        final ObjectMapper mapper = new ObjectMapper();
-        final ObjectNode root = mapper.createObjectNode();
-        root.put("id", newId);
-        root.put("name", GraphNode.getGraphNode(newId).getDisplayName());
-        root.put("schema", schemaName);
-        mapper.writeValue(out, root);
+        try {
+            newId = RestServiceUtilities.waitForGraphChange(existingId).get(10, TimeUnit.SECONDS);
+        } catch (final InterruptedException ex) {
```
antares1470, line 123: are you able to throw an exception here as well (I assume if we interrupt here we probably don't want to continue execution)? Or if it that isn't the intended execution, perhaps a log entry to indicate the interrupt has occurred?

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,13 +116,23 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+            Thread.currentThread().interrupt();
+        } catch (final ExecutionException | TimeoutException ex) {
+            throw new RestServiceException(ex);
+        }
+
+        if (!newId.isBlank() && !newId.isEmpty()) {
```
antares1470, line 129: `StringUtils.isNotBlank(newId)`

Side note: empty implies blank (i.e. the empty string is a subset of the blank strings which contain at most only whitespace characters) so you don't need to specify both

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -113,13 +116,23 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+            final ObjectMapper mapper = new ObjectMapper();
+            final ObjectNode root = mapper.createObjectNode();
+            root.put("id", newId);
+            root.put("name", GraphNode.getGraphNode(newId).getDisplayName());
+            root.put("schema", schemaName);
+            mapper.writeValue(out, root);
```
antares1470, line 135: I could have this wrong but based on the code above, would it make more sense for this code to be in the try block? Doesn't seem like you want this code to execute if you get any exceptions (incl. interrupt)

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/OpenGraph.java
```diff
@@ -113,6 +116,10 @@ public void callService(final PluginParameters parameters, final InputStream in,
             mapper.writeValue(out, root);
         } catch (final GraphParseException | FileNotFoundException ex) {
             throw new RestServiceException(HTTP_UNPROCESSABLE_ENTITY, ex.getMessage());
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
```
antares1470, line 120: Similar to `NewGraph` are you able to add either a log entry or throw an exception here?

CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorTopComponent.java
```diff
@@ -123,21 +120,20 @@ public AttributeEditorTopComponent() {
...
-                    Thread.sleep(50);
-                }
 
-                if (reader != null) {
-                    attributePanel.updateEditorPanel(reader.refreshAttributes());
-                }
-            } catch (final InterruptedException ex) {
-                LOGGER.log(Level.SEVERE, "Thread was interrupted");
-                Thread.currentThread().interrupt();
+            final ArrayList<Object> devNull = new ArrayList<>();
```
antares1470, line 124: noting it appears you've just moved this out of the try block but are you possibly able to change this declaration to `List` rather than `ArrayList` (instantiation is fine)?

antares1470 (changes requested):

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/restapi/RestServiceUtilities.java
antares1470: This check can still be removed (as the check is already done to get into the if block)

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/NewGraph.java
```diff
@@ -120,19 +124,20 @@ public void callService(final PluginParameters parameters, final InputStream in,
...
+                final ObjectMapper mapper = new ObjectMapper();
+                final ObjectNode root = mapper.createObjectNode();
+                root.put("id", newId);
+                root.put("name", GraphNode.getGraphNode(newId).getDisplayName());
+                root.put("schema", schemaName);
+                mapper.writeValue(out, root);
+            }
         } catch (final InterruptedException ex) {
             Thread.currentThread().interrupt();
+            LOGGER.log(Level.SEVERE, "Thread interrupted");
```
antares1470, line 137: Are you able to add the exception as an argument here (just add ex as a 3rd parameter)

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/OpenGraph.java
```diff
@@ -118,6 +122,7 @@ public void callService(final PluginParameters parameters, final InputStream in,
             throw new RestServiceException(HTTP_UNPROCESSABLE_ENTITY, ex.getMessage());
         } catch (final InterruptedException ex) {
             Thread.currentThread().interrupt();
+            LOGGER.log(Level.SEVERE, "This thread has been interrupted");
```
antares1470, line 125: same here

antares1470 (changes requested):

antares1470 (approved):

PR #1819, opened 2023-03-24
Update VisualGraphOpener.java

serpens24:
### Prerequisites

- [X ] Reviewed feedback from the "Sonar Cloud" bot. Note that you have to wait

### Description of the Change
Added code to handle case where there is a valid graph backup, but also a corrupt grpah and autosave file. Previously in this case, if the user elected to open autosave, the graph is copied over top of the backup - even if its corrupt (which isn't known until opening it) - so if the autosave is also corrupt, then the only file that wasnt (the backup) was deleted and the two corrupt files remained.
Now in this case, if a user is opening an autosave, and there is a backup, that backup is temporarily backed up and only used if the autosave and main file are corrupt.

### Alternate Designs

Could have tried to oppen the graph and catch read errors before copying it over backup, but this is expected to take more time.

Reduce chance of user losing work

### Benefits

As above

### Possible Drawbacks

Potentially if user killed application at exact time between creating the secondary backup, and it being used or deleted it could persist afterwards.

### Verification Process

Follow steps called out in the relatred ticket.

### Applicable Issues

nil

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/IconEditorFactory.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/editors/IconEditorFactory.java
@@ -171,19 +171,33 @@ public class IconEditorFactory extends AttributeValueEditorFactory<Constellation
             refreshIconList(iconFile);
         }
 
-        private List<File> pngWalk(File path) {
+        /**
+         * Search for all valid icons, starting at the supplied path.
+         * @param path Path to search for icons.
+         * @return List of all icon files found.
+         */
+        private List<File> iconWalk(File path) {
             final List<File> files = new ArrayList<>();
-            pngWalk(path, files);
+            iconWalk(path, files);
             return files;
         }
 
-        private List<File> pngWalk(final File path, final List<File> files) {
+        /**
+         * Navigate through all files in path and its subdirectories, adding any
+         * files found to match icon file types to the list of icons returned.
+         * @param path Path to search for icons.
+         * @param files Files already found.
+         * @return Complete list of all icon files found (new + those passed in).
+         */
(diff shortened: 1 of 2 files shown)


CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphOpener.java
```diff
@@ -121,7 +121,17 @@ public void openGraph(final GraphDataObject gdo) {
                         // Backup the current actual file and replace it with the autosave file.
                         final File autosaved = new File(AutosaveUtilities.getAutosaveDir(), props.getProperty(AutosaveUtilities.ID) + FileExtensionConstants.STAR);
                         try {
-                            AutosaveUtilities.copyFile(autosaved, f);
+                            // make temp copy of any backup file that exists to try if we find both autosave and main file are corrupt
+                            final File toBak = new File(f.getPath() + FileExtensionConstants.BACKUP);
+                            if (toBak.exists()) {
+                                // Make a backup of our backup which is about to be overwrtten by the Autodsave call
+                                final File toBakBak = new File(toBak.getPath() + FileExtensionConstants.BACKUP);
+                                final boolean toRenamed = toBak.renameTo(toBakBak);
```
OrionsGuardian, line 129: The variable `toRenamed` is unused.
Should this be checking if the rename operation was successful?

OrionsGuardian (changes requested):

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphOpener.java
antares1470: Appears to be used in the line below it

Issue #1762, opened 2022-09-06
Optimise notes display in Notes View


sol695510:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Changes to be implemented better display notes in the Notes View:

- Move ‘new note’ UI to a pop-up to free space to view more notes.

- Move ‘edit’ and ‘remove’ buttons of each note to the top as depending on the size of the note, having them centred vertically can be difficult for usability.

- Add a 'show more / show less' toggle when the information on a note exceeds a defined limit to reduce how information on notes are displayed, as to reduce size of notes and increase number of notes viewable.

### Additional Information

N/A

GammaVel:
Thank you @altair1673 for improving the notes view here, this looks great. I like how the Show More and Show less is working.

PR #1820, opened 2023-03-27
Moved note creation to popup window

altair1673:
### Prerequisites

### Description of the Change

Note creation controls  which are the add/cancel butts, title text box and the content text area has been moved to a popup window which is spawned by clicking a button in the Notes View. A show more/less button is added to notes that have more than 500 characters in their description. The edit and delete buttons are now at the top of the note instead of to the right.

With the note creation controls moved there is more space in the view and more notes can be seen. The show more/less button keep notes to a default height so that extremely long notes don't take up the whole view when created. Moving the edit and delete buttons to the top of the note allow for more space in the note description so the users can see more of their text before the default text limit is reached.
### Benefits

Quicker navigation of notes since there height is capped based on a set number of characters allowed for notes in their default state. Show more/less button can be used to hide description of notes that are not currently relevant. UI looks cleaner with edit and delete buttons at the top. The bottom bar only has one button to create a new note instead of all the controls which reduces clutter when user is scrolling through notes.
### Possible Drawbacks

The default limit for notes is 500 characters before the show more/less button pops up. I chose this because the height if notes by default looked appropriate.
### Verification Process

Tested that clicking add on the popup window adds the note to the view.
### Applicable Issues

#1762


mimosa2 (approved):

OrionsGuardian:
Please address the code smells.
They are mostly trivial and easy to comply with.

antares1470 (changes requested):
Can you also add a whatsnew entry for this (and some of the other notes view changes that have already been merged in)?

From testing, I'm not sure whether it was introduced this PR or a previous, but there appears to be issues with the display in the notes view. Namely:
1) The width of the notes content in the notes appears to be wider than the default width for the view (i.e. you haven't manually adjusted the width previously) and has no horizontal scroll bar so you have to widen the view to see everything (at the very least a scroll bar should come up but personally I think everything should be accessible from the default width no scrolling required)
2) Values of drop-downs at the top of the view (e.g. which notes are displaying) aren't displaying when they are ticked (previously when you were no longer in focus, it would display a comma-separated list of all the ticked values).
notesViewSizeAndDisplayIssues

Also, I don't think I'm a huge fan of the buttons at the top because at the moment there just looks like there is a unnecessary gap between the note text and the top of the note (as shown in the user note in the gif). Perhaps if the buttons were left-aligned rather than right it might be better? Personally I think the vertical arrangement (as it was) is still better as it keeps the text in a similar place as their Auto note counterparts for consistency which I think looks a little cleaner. The two button heights together also never really went below the lowest point of the note text even for the smallest user notes which I think is important to note as while moving the buttons to the top might make reduce the size of larger notes, you may actually be increasing the size of shorter notes

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewTopComponent.java
```diff
@@ -85,6 +86,7 @@ public NotesViewTopComponent() {
             }
             notesViewPane.updateNotesUI();
         });
+        jfxContainer.setMinimumSize(new Dimension(1000, 500));
```
OrionsGuardian, line 89: A minimum width of 1000 is far too wide for a view pane.
Can we try a value around 400.

OrionsGuardian:
The first time you try to create a note, the popup window is displayed on the wrong monitor.
It goes to the default system monitor, rather than the monitor where constellation is running.

Also, I'd suggest some minor changes to the create new note screen:
image

OrionsGuardian (changes requested):

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewTopComponent.java
antares1470, line 89: I'd prefer to make the view usable in the existing default width rather than simply changing the default width. Having tested this on my computer, the view by default now takes up more than half the width of my screen which just isn't practical (and reducing the width manually brings up the same issues as before). 

One idea might be to look what was done at the top of the Data Access View and have elements (e.g. buttons) on the screen flow onto the next line if they don't fit in the current width

antares1470 (changes requested):

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewTopComponent.java
antares1470: As per my comment below, I'd actually prefer to avoid playing around the view sizes altogether. Rather the view should be made to be usable with the default size determined by the application

OrionsGuardian:
The Show More / Show Less buttons should dynamically appear if the note text will go over more than 5 lines.
The code currently only offers the Show More / Show Less buttons if the text contains more than 500 characters of data.

image

In the above image, Note A is showing the first 500 characters of data on 5 lines across the screen, but if the width of the view panel is reduced, it will wrap the text and use more lines to continue to show the first 500 characters of data.
Note B has many newlines in the text, and the 500th character appears on the 49th line, so it uses 49 lines of space to show the first 500 characters.

sol695510:
Currently this is what the Notes View looks like, with the time range menu closed and opened respectively:

NV current look tr closed

NV current look tr open

Here's some mocked up design suggestions for minimising the space used by the currently implemented design:

### **Notes View, normal width**
New NV, expanded
- Moving the tags menu below the filters menu
- Moving the create note button to the top since note creation is no longer at the bottom
- Utilising the time range menu used in the Data Access View as it is more compact

### **Notes View, condensed width**
New NV, condensed
- Elements within the time range menu shift to accommodate the more vertical size instead of restricted the view's width

In addition, current the tags menu contains no prompt text like the filters menu which says "Select a filter...". I propose not to hide the tags menu but rather disable it when the Auto Notes selection is deselected in the filters. This lets the user know there is a function there but not available with the current selection instead of hiding the element and leaving a blank space in the UI.

### **No filters selected**
filter unselected, tags unselected (disabled)

### **Only User Notes selected**
filter selected (user), tags unselected (disabled)

### **User Notes and Auto Notes selected**
filter selected (user, auto), tags unselected (enabled)

### **Only Auto Notes selected**
filter selected (auto), tags unselected (enabled)


Also the new note creation pop up could be condensed in its design as well. Here is the current design as well as a mocked up suggestion:

### **Current**
NV create note dialog

### **Mocked up suggestion**
New NV, create note
- I think the "Enter note info" text is unnecessary as the pop up window already has a title with similar text

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NewNotePane.java
```diff
@@ -163,13 +173,23 @@ public void showPopUp() {
...
+
+            final Scene s = new Scene(dialogPane);
+            s.getStylesheets().add(getClass().getResource("resources/notes-view.css").toExternalForm());
+
+            stage.setScene(s);
+
             isFirstTime = false;
         }
 
+        List<Screen> screens = Screen.getScreensForRectangle(window.getX(), window.getY(), window.widthProperty().get(), window.heightProperty().get());
```
antares1470, line 188: final

antares1470 (changes requested):
Some things from testing:
1) I'm a little confused by the Get Height button on user notes. Not exactly of its purpose (maybe it was a debugging thing?)
2) I'm finding that the Show more/less button displays on all user notes, regardless of what size it is. I think this should only be displaying if the contents is hidden in any way
3) It seems with the css that some of the previous formatting has been lost. Namely the note title is no longer bold
4) The auto notes also appear to no longer be displaying the datetime information
5) Agreeing with @sol695510, I think the Create Note button may be better placed in the top corner. Truthfully I found once the view was populated with notes that the button almost seemed to get lost at the bottom of the view (but that might be just me)
6) This may not be related to this particular ticket (possibly one of the others) but I'm finding the time filter isn't actually working. I created a note and then did a filter on a time range which should've included that note -> no notes were displayed.

OrionsGuardian:
image

antares1470:
> 6\. This may not be related to this particular ticket (possibly one of the others) but I'm finding the time filter isn't actually working. I created a note and then did a filter on a time range which should've included that note -> no notes were displayed.

I gave this another test today and it appears to be working fine. Not sure what happened previously but I'll let you know if I see it again

Delphinus8821 (commented):
Looks like it works well.
Still needs a what's new entry and for the code smells to be cleaned up.

antares1470 (changes requested):
Looking good. On top of what @Delphinus8821 has said, I have a couple other things:
1) Are you possibly able to update the notes view help page, namely the main screenshot and the information for user-generated notes? These will be out-of-date.
2) Found an odd bug whereby sometimes when I click the Create Note button, the popup displays at a window size smaller than it should (and cutting off some of the dialog box). Seems to be a bit random as to when it pops up at the correct window size vs the reduced size. See gif
notesViewNotePopup

antares1470 (changes requested):
1) Can you add a whatsnew entry for this (and possibly some of the other recent notes view changes you've made)?
2) Found another bug whereby if you have a note that shows more and then edit it so it no longer needs it, it removes the button, but if you then edit it again so that it is required, the button doesn't pop up.
notesViewShowMoreDisappear

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewTopComponent.java
```diff
@@ -17,11 +17,15 @@
 
 import au.gov.asd.tac.constellation.graph.Graph;
 import au.gov.asd.tac.constellation.graph.manager.GraphManager;
+import au.gov.asd.tac.constellation.graph.monitor.GraphChangeEvent;
 import au.gov.asd.tac.constellation.graph.schema.visual.concept.VisualConcept;
 import au.gov.asd.tac.constellation.plugins.reporting.GraphReportListener;
 import au.gov.asd.tac.constellation.plugins.reporting.GraphReportManager;
 import au.gov.asd.tac.constellation.plugins.reporting.PluginReport;
 import au.gov.asd.tac.constellation.views.JavaFxTopComponent;
+import java.awt.Dimension;
```
OrionsGuardian, line 26: Remove unused imports (lines 20 and 26)

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/DateTimeRangePicker.java
```diff
@@ -23,22 +23,20 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.logging.Logger;
```
OrionsGuardian, line 26: Remove unused import

OrionsGuardian (changes requested):

antares1470 (changes requested):
Can you add an entry to whatsnew.txt for this (and possibly some of the other recent notes view changes you've made)?

OrionsGuardian:
image

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,12 @@
 == 3030-12-31 Getting Started
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2023-05-05 Notes View UI Change
+<p>The controls to add a new note are no longer at the bottom and have been moved to a pop-up window which appears when the "Create Note" button is pressed on the top left.</p>
+<p>A colour picker has been added to the note creation window as well as the inidividual notes in the edit mode so now there can be different coloured notes.</p>
+<p>Notes are now a fixed size and a "Show More" button will appear on notes with content that exceeds the minimum height.</p>
+<p>The Time-Ranger filter control is in a drop down menu at the top and takes up less space.</p>
```
antares1470, line 8: Given time-range (not time-ranger 😃) hasn't been mentioned previously, you can probably omit the bit about taking up less space and just state that the ability to filter on time-range has been added

CoreWhatsNewView/src/au/gov/asd/tac/constellation/views/whatsnew/whatsnew.txt
```diff
@@ -1,6 +1,12 @@
 == 3030-12-31 Getting Started
 <p>If you're new to Constellation, read the <a href="" helpId="au.gov.asd.tac.constellation.functionality.gettingstarted">getting started guide</a>.</p>
 
+== 2023-05-05 Notes View UI Change
+<p>The controls to add a new note are no longer at the bottom and have been moved to a pop-up window which appears when the "Create Note" button is pressed on the top left.</p>
+<p>A colour picker has been added to the note creation window as well as the inidividual notes in the edit mode so now there can be different coloured notes.</p>
+<p>Notes are now a fixed size and a "Show More" button will appear on notes with content that exceeds the minimum height.</p>
```
antares1470, line 7: Not sure it's strictly true that notes are a fixed size (notes with less text don't pad with extra blank lines) so it's not technically a minimum height either. Perhaps it would more accurate to suggest that larger notes (>5 lines) will be compressed with Show More/Less options

antares1470 (changes requested):
Just some minor changes for the whatsnew. Looks good in testing

antares1470 (approved):

Delphinus8821 (approved):

Issue #1796, opened 2023-01-11
Minor wording update suggested to File Import message for rows being skipped


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

After importing a file a message pops up to let you know how many rows were imported. This is super useful. The message currently has "Skipped rows <0/a number> due to import error. 
At the moment it looks like this could be referring to the row number that was skipped or the number of rows skipped.
![image](https://user-images.githubusercontent.com/83679004/211700947-22385e04-c4df-48e7-886a-3e464807085d.png)

As an enhancement it would be better if this was: 
Skipped <0/a number> rows due to import error. 

OR
Skipped rows: <0/a number> due to import error.  (with a colon)

OR 
Rows skipped: <0/a number> due to import error.  (with a colon and reorder of wording)

### Steps to Reproduce

1.  Import a file that contains an error in the date column, but most of the dates are in correct format
2. Click "Skip invalid rows"
3. Import > Observe message that is displayed

**Expected behaviour:** Small wording change suggested

**Actual behaviour:** NA

**Reproduces how often:** 100%
### Additional Information

![image](https://user-images.githubusercontent.com/83679004/211701133-935e92f2-b465-4e09-968d-965d28b6f319.png)

Auriga2:
It'd be nice if we could word `row or rows` instead of `row(s)`  similar to `file or files` in the error message, as that's a simple check in the code.

![image](https://github.com/constellation-app/constellation/assets/64826928/22efb455-0d72-4c4a-a3df-f2dda6bed0bd)

![image](https://github.com/constellation-app/constellation/assets/64826928/b7ddf62c-783c-40b1-8d87-a8b5bb9f21ae)

GammaVel:
Thanks for fixing the wording here, looks great!

Here is the new text:
![image](https://github.com/constellation-app/constellation/assets/83679004/9919ec15-3962-4dcf-9740-25581d352ed0)

[...]

Issue #1795, opened 2023-01-11
CSV File import - Runtime error if there is an error with the Date and "Skip Invalid rows" is not ticked


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Constellation has the option now to "Skip Invalid Rows" when importing a file that contains errors. If I don't tick this option and the CSV file contains an error with the date field, a runtime error message is displayed (shown in the screenshot section). This is 100% reproducible. It would be good if a more user friendly error message was presented instead of the Runtime error message to help me know how to resolve this issue.  

I can select "Skip Invalid Rows" to get around this error but it wasn't immediately apparent to me to do this. 
This is more of an issue when the error is outside of the 100 preview rows as I do not know there is an error present until I do the import. 

### Steps to Reproduce

1.  Set up a file that has an error on a date field that is outside of the first 100 rows
2.  Use the Import from file functionality and select file
3. Perform the field mapping, including the date field
4. Click import > the run time error is shown
5. Click ignore invalid rows > the runtime error is not shown, and the import completes as expected

### Behaviour

**Expected behaviour:** A message displays saying "Unable to complete import due to error with data. The file can be imported if you select Ignore invalid rows "

**Actual behaviour:** Run time error:
![image](https://user-images.githubusercontent.com/83679004/211699113-03fff940-1d23-403a-8c6d-6230b402156f.png)

**Reproduces how often:** 100%

### Additional Information

This issue relates to https://github.com/constellation-app/constellation/issues/1449

Auriga2:
It'd be nice to word the message to `.........if you select the Ignore invalid rows checkbox next to the Import button` or similar, specially for new users to locate this check box. I initially checked the options at the top.

GammaVel:
Thanks @Auriga2 and @spica198 for looking into this issue. I tried this out again and there is no run time exception being shown which is perfect. I noticed a small typo that would be good to fix, so I made a new issue for that.

Issue #1731, opened 2022-07-19
Importing from file - java.lang.NumberFormatException: For input string: "ERRO" when header row is included in the data


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

This is a small edge case to do with importing from a CSV file which has date/time field mapping set and the header row included in the loaded data.

A runtime error is being displayed during the import from file process, if you have not selected the "File has headers" and the header row is included on the first line of the import and the date mapping has been applied to the column.  

When the date time field is applied to the column, the header row contents are being shown as ERROR. 
It is still possible to select the import button. When this is done a runtime error is displayed : java.lang.NumberFormatException: For input string: "ERRO" (full details below) 

### Steps to Reproduce

1.  Select a sample CSV file which includes a date/time field and has a header row which is the column titles
2.  Choose Destination - Analytic graph - do not select the "File has headers" option
3.  Observe in the preview that the column headings are present as the first line
4.  Apply the date/time to the date time column
5. Observe the column heading for the date/time has ERROR as expected
![image](https://user-images.githubusercontent.com/83679004/179640327-e9734a65-c0a2-4424-9097-3c94a10eae8c.png)

6. Complete the rest of the file mapping and select import > Run time error is displayed > 
7.  Repeat with another file to make sure this is repeatable > it is

**Expected behaviour:** The error field is ignored and the import completes with a message saying how many error rows were ignored

**Actual behaviour:** Run time error > I need to manually correct this error by turning on the option for file has headers (hence this is not a big issue)
**Reproduces how often:** 100%

### Additional Information

![image](https://user-images.githubusercontent.com/83679004/179639738-422ad07e-41bf-4ab4-9ef9-f859e4b7350e.png)

Auriga2:
I now get `java.lang.IllegalArgumentException` exception when recreating this. Can you please check @spica198 
[...]

PR #1821, opened 2023-04-03
1731 file import error numberformtexception

spica198:
### Prerequisites

### Description of the Change

When the date time field is applied to the column, the header row contents are being shown as ERROR. 

Expected behaviour: The error field is ignored and the import completes with a message saying how many error rows were ignored

Cover the filter with try-catch and increment skipInvalidRows counter
### Applicable Issues
#1731 #1795 #1796

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportDelimitedPlugin.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportDelimitedPlugin.java
@@ -167,7 +167,7 @@ public class ImportDelimitedPlugin extends SimpleEditPlugin {
                 // At least 1 object was successfully imported. List all successful file imports, as well as any files
                 // that there were issues for. If there were any files with issues use a warning dialog.
                 final String fileFiles = (validFilenames.size() == 1) ? "file" : "files";
-                sbMessage.append(String.format("Extracted data from %d rows in %d %s. Skipped %d rows due to import error.",
+                sbMessage.append(String.format("Extracted data from %d row(s) in %d %s. Skipped %d row(s) due to import error.",
                         importedRows, validFilenames.size(), fileFiles, skippedRows));
                 sbMessage.append(" Files with data: ");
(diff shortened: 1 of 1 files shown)


CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportDelimitedPlugin.java
```diff
@@ -384,31 +386,40 @@ private static Map<String, Integer> processVertices(final ImportDefinition defin
...
+                            throw ex;
+                        }
                     }
                 }
+            } catch (final DateTimeException | IllegalArgumentException | SecurityException ex) {
+                if (skipInvalidRows) {          
+                    ++skippedRow;
+                } else {                    
+                    throw ex;
+                }
```
OrionsGuardian, line 422: There are nested try-catch blocks with exactly the same catch conditions.
This can be re-written to only use the outer try-catch block. We can move the `graph.removeVertex(vertexId);` to the outer catch block, but make it conditional on whether there was a new vertex added.
ie. initialise vertexId to -1 before the if statement, then in the catch block,  if vertexId != -1, remove the vertex.

OrionsGuardian (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/delimited/ImportDelimitedPlugin.java
```diff
@@ -484,6 +495,7 @@ private static Map<String, Integer> processTransactions(final ImportDefinition d
                         graph.removeVertex(destinationVertexId);
                         ++skippedRow;
                     } else {
+                        NotifyDisplayer.display("Unable to complete import due to error with data. The file can be imported if you select Ignore invalid rows", NotifyDescriptor.PLAIN_MESSAGE);
```
antares1470, line 498: maybe make this an error message rather than plain message

antares1470 (changes requested):

antares1470 (approved):

OrionsGuardian (approved):

Issue #1682, opened 2022-04-19
Enhance Data Access view Configuration to show order of categories


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Data Access View (DAV) options under configuration shows a list of available categories to show on screen. The order here does not get translated to the order shown in the actual Data Access View. 

That is, if I change the order of the DAV categories in the configuration, I still have the default order shown in the data access view and my changes are not reflected. As an enhancement it would be great to expand the functionality to control the order of the Categories on display. 

### Steps to Reproduce

1.  Go Set up > Options > Data Access View
2. Change the order around to be different to the default. This can be done by hiding all categories and selecting the category to be visible in the desired order
3. Save and exit Constellation for the changes to apply
4. Restart constellation and check the order of Categories > this is the default order and not the order I have entered in the DAV set up options

**Expected behaviour:** as an enhancement the order of categories in DAV config is shown in the DAV and there is an option to change the ordering within the visible categories

**Actual behaviour:** 
![image](https://user-images.githubusercontent.com/83679004/163908281-83609496-0857-49ea-86d4-51dd8cba2504.png)

**Reproduces how often:** 100%

### Additional Information

It would be great to build on the awesome work that was done to be able to hide the categories.

GammaVel:
Hi @altair1673, thanks very much for working on this. It looks like there could be an issue here if you have hidden a category in version 2.10 and then open up the latest nightly build. 

In version 2.10 I had hidden one of the columns from the DAV. When I loaded the latest windows nightly build (v2.11) it didn't shown any of my categories, it was blank. Here is what it looked like:

![image](https://github.com/constellation-app/constellation/assets/83679004/749b48c2-0713-4fb9-9301-8db36ce7cda8)

I tried to replicate this bug by repeating the exact same steps and the next time Enrichment was shown both as a visible and a hidden column

[...]

PR #1842, opened 2023-06-01
Data Access View now orders plugins as configured

altair1673:
### Prerequisites

### Description of the Change

The order of the "visible" list under the DAV configurations in options now determine the order of plugins shown in the DAV. 

Gives user control of the order of categories to be displayed in the DAV. Possibly increases efficiency if user is allowed to customize the layout for productivity purposes.
### Possible Drawbacks

Currently to re-arrange everything in the visible list they must all be moved to the hidden list and then moved back in the order the user wants the categories to be.
### Verification Process

Tested by changing the order of categories in the config window, closing consty and reopening it to see the DAV show the categories in the specified order.
### Applicable Issues

#1682


antares1470 (changes requested):
I'm not sure how I feel about this as a solution. Moving categories back and forth between lists isn't particularly intuitive and feels more like a workaround than an actual solution. The expected behaviour in the original ticket identified wanting to be able move categories within the visible list itself and I think this is still the desired behaviour to pull off this request. 

Perhaps an interim solution addressing the ticket is to reset the order presented in the preference window (to match the current order, as this was an issue identified in the ticket). Until we can move categories in the visible list itself though, I don't feel it is worth including re-ordering.

As an aside, I did some testing on the current solution. If you haven't previously moved any categories into the hidden list, you get presented with this in the Data Access View 😃:
davCategoriesError

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanel.java
```diff
@@ -289,10 +323,72 @@ private void buttonLeftActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FI
...
+        else
+        {
+            final int[] selectedIndices = visibleList.getSelectedIndices();
+
+            if(selectedIndices[0] != 0)
+            {
+                final int beforeIndex = selectedIndices[0]-1;
+                final List<String> selectedItems = visibleList.getSelectedValuesList();
+
+                visibleListModel.removeRange(selectedIndices[0], selectedIndices[selectedIndices.length - 1]);
```
OrionsGuardian, line 339: This code assumes the selected entries are in one contiguous block, when it's actually possible to have gaps in the selections. 
ie. Use Ctrl-Click to select the 2nd and 4th entries, leaving the 1st and 3rd entries unselected.
Then if you click the Up button, the 3rd entry will be removed, and not be added back in later code.

Probably best to not allow gaps in the selections.
Add these lines to the constructor:
```
        visibleList.setSelectionMode(ListSelectionModel.SINGLE_INTERVAL_SELECTION);
        hiddenList.setSelectionMode(ListSelectionModel.SINGLE_INTERVAL_SELECTION);
```

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanel.java
```diff
@@ -289,10 +323,72 @@ private void buttonLeftActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FI
...
+            final int[] selectedIndices = visibleList.getSelectedIndices();
+
+            if(selectedIndices[selectedIndices.length-1] != visibleList.getModel().getSize()-1)
+            {
+                final int afterIndex = selectedIndices[selectedIndices.length-1]+1;
+                final List<String> selectedItems = visibleList.getSelectedValuesList();
+
+                visibleListModel.removeRange(selectedIndices[0], selectedIndices[selectedIndices.length - 1]);
+
+                visibleListModel.addAll(afterIndex-selectedIndices.length, selectedItems);
```
OrionsGuardian, line 370: Missing a "+1" on this line; code should be:
```
    visibleListModel.addAll(afterIndex - selectedIndices.length + 1, selectedItems);
 ```

**Review by OrionsGuardian — CHANGES_REQUESTED** [review]

**Review by OrionsGuardian — CHANGES_REQUESTED** [review]
Other Issues:
Initial order in the Options window does not match the order in the Data Access View.
Changing the order in the Options windows doesn't enable the Apply button.

**Comment by OrionsGuardian** [comment]
image

File: CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/QueryPhasePane.java [hunk 3]
```diff
@@ -77,25 +81,31 @@ public class QueryPhasePane extends VBox {
...
-                    
-                    dataSourceList.getChildren().add(heading);
-                    dataSources.addAll(heading.getDataSources());
+
+            orderedPlugins.add(new Pair<>(pluginsOfType.getValue().getKey(), heading));
                 });
 
+        orderedPlugins.sort(Comparator.comparingInt(Pair<Integer, HeadingPane>::getKey));
+
+        for (int i = 0; i < orderedPlugins.size(); i++) {
```
antares1470, line 103: use a for each here instead. i.e. `for (final Pair<Integer, HeadingPane> plugin : orderedPlugins) {...`

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/tasks/LookupPluginsTask.java
```diff
@@ -50,7 +56,26 @@ public Map<String, List<DataAccessPlugin>> get() {
                 }
             }
         }
-        return plugins;
+
+        if (StringUtils.isNotBlank(VISIBLE_CATS)) {
+            final String[] visibleCategoriesArray = addCategoryToList(VISIBLE_CATS);
+            if (visibleCategoriesArray.length > 0) {
+                plugins.keySet().forEach(key -> {
```
antares1470, line 63: use the entry set here as well

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/tasks/LookupPluginsTask.java
```diff
@@ -50,7 +56,26 @@ public Map<String, List<DataAccessPlugin>> get() {
...
+                            pluginsWithOrder.put(key, p);
+                            break;
+                        }
+                    }
+                });
+            }
+        }
+
+        if (StringUtils.isBlank(VISIBLE_CATS) && StringUtils.isBlank(DAV_CATS)) {
+            plugins.keySet().forEach(key -> pluginsWithOrder.put(key, new Pair(0, plugins.get(key))));
```
antares1470, line 76: and here

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/tasks/LookupPluginsTask.java
```diff
@@ -50,7 +56,26 @@ public Map<String, List<DataAccessPlugin>> get() {
...
+                            final Pair<Integer, List<DataAccessPlugin>> p = new Pair<>(i, plugins.get(key));
+                            pluginsWithOrder.put(key, p);
+                            break;
+                        }
+                    }
+                });
+            }
+        }
+
+        if (StringUtils.isBlank(VISIBLE_CATS) && StringUtils.isBlank(DAV_CATS)) {
```
antares1470, line 75: given the if statement immediately above it is `StringUtils.isNotBlank(VISIBLE_CATS)`, would it make more sense to make this an else-if condition of that? It would mean you wouldn't need to include the first boolean statement here (as that will be implied by the if condition falling to the else-if). i.e.
```
if (StringUtils.isNotBlank(VISIBLE_CATS)) {
    ...
} else if (StringUtils.isBlank(DAV_CATS)) {
    ...
}
```

antares1470 (changes requested):
To add to below comments, I was wondering whether the up and down buttons could be moved somewhere to make them more obvious as to the scope of their use? Perhaps below the visible categories list? I think in their current position, it suggests the possibility of being able to move items in either list (which obviously isn't the case)

antares1470 (changes requested):
To add to the above:
1) Could you add a whatsnew entry for this?
2) given a number of public functions have had their definition changed, it might be good to have some entries in the changelog for them

Delphinus8821 (approved):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -73,6 +77,7 @@ public void applyChanges() {
                 final Preferences prefs = NbPreferences.forModule(DataAccessViewPreferenceKeys.class);
                 final DataAccessViewCategoryPanel panel = getPanel();
                 prefs.put(DataAccessViewPreferenceKeys.HIDDEN_DA_VIEW, panel.getHiddenCategory().toString().replace("[,", "["));
+                prefs.put(DataAccessViewPreferenceKeys.VISIBLE_DA_VIEW, panel.getVisibleCategory().toString().replace("[,", "["));
                 panelRefreshed = true;
```
OrionsGuardian, line 81: After this line is where you should be setting the ordering variables back to false:
```
                orderChanged = false;
                reorderButtonPressed = false;
```

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -96,8 +101,22 @@ public boolean isChanged() {
...
+
+
+        if (reorderButtonPressed) {
+            if (!visibleCategory.equals(visibleNow)) {
+                orderChanged = true;
+                visibleNow = panel.getVisibleCategory();
+            } else {
+                orderChanged = false;
+                reorderButtonPressed = false;
+            }
```
OrionsGuardian, line 114: Remove the **else** block. Reset these variables after changes have been applied.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -96,8 +101,22 @@ public boolean isChanged() {
...
+        }
+
+        return !hiddenCategory.isEmpty() || orderChanged
+                || !prefs.get(DataAccessViewPreferenceKeys.HIDDEN_DA_VIEW, DataAccessViewPreferenceKeys.HIDDEN_DA_VIEW_DEFAULT).isEmpty()
+                || !prefs.get(DataAccessViewPreferenceKeys.VISIBLE_DA_VIEW, DataAccessViewPreferenceKeys.HIDDEN_DA_VIEW_DEFAULT).isEmpty();
```
OrionsGuardian, line 119: The logic here needs to be revised.
It should be checking the current visible list against the entries that were initially in the list when the options window was opened.
```
    return orderChanged || visibleEntriesHaveChanged();
```

OrionsGuardian (changes requested):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -96,8 +107,26 @@ public boolean isChanged() {
         final Preferences prefs = NbPreferences.forModule(DataAccessViewPreferenceKeys.class);
```
OrionsGuardian, line 107: remove unused variable

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -96,8 +107,26 @@ public boolean isChanged() {
...
         final DataAccessViewCategoryPanel panel = getPanel();
         final List<String> hiddenCategory = panel.getHiddenCategory();
```
OrionsGuardian, line 109: remove unused variable

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -96,8 +107,26 @@ public boolean isChanged() {
...
+
+        return visibleEntriesHaveChanged() || orderChanged;
+    }
+
+    private boolean visibleEntriesHaveChanged() {
+        if (firstLoad) {
+            return false;
+        }
+
+        return getPanel().getVisibleCategory().size() != visibleNow.size();
```
OrionsGuardian, line 129: Checking the size is not enough, you also need to check the entries individually.
You could have a visible list containing entries A, B, and C, and a hidden list containing entry D.
Then move entry D back to the visible list and move entry A to the hidden list.
So you now have B, C, and D in the visible list, and A in the hidden list.
The size has remained the same, but the list now has different entries, 
so the **visibleEntriesHaveChanged** function should be returning **true**.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -123,7 +152,18 @@ public void removePropertyChangeListener(final PropertyChangeListener l) {
...
+            visibleNow = thePanel.getVisibleCategory();
         }
         return thePanel;
     }
+
+    public void setReorderButtonPressed(final boolean reorderButtonPressed) {
+        this.reorderButtonPressed = reorderButtonPressed;
+    }
+
+    public void setMoveButtonPressed(boolean moveButtonPressed) {
```
OrionsGuardian, line 164: final

OrionsGuardian (changes requested):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
altair1673: Yeah this was pretty dumb will fix it now.

OrionsGuardian:
Problem with the Cancel button.
It only works when you first view the Options screen. 
You can make changes to the lists and press Cancel and it will restore them to their last saved state.
But after Applying any changes to the lists, the Cancel button no longer works. 
If you make more changes in the options window, then press Cancel. ... All those changes are kept and shown the next time you open the options window, when it should have reverted the changes back to the last saved state.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanelController.java
```diff
@@ -121,6 +133,8 @@ public boolean isChanged() {
 
     private boolean visibleEntriesHaveChanged() {
         if (firstLoad) {
+            final Preferences prefs = NbPreferences.forModule(DataAccessViewPreferenceKeys.class);
+            LOGGER.log(Level.SEVERE, "visible categories: " + prefs.get(DataAccessViewPreferenceKeys.VISIBLE_DA_VIEW, DataAccessViewPreferenceKeys.HIDDEN_DA_VIEW_DEFAULT));
```
OrionsGuardian, line 137: Are these lines still needed?

OrionsGuardian (changes requested):

antares1470 (approved):

Issue #1835, opened 2023-05-24
Inline changes made to notes aren't recognised as changes to be saved


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Once a note has been created in the Notes View, you have the option to edit the note (text, colour, etc). When you make these edits though, these aren't recognised as changes to the file and so you can't save the change to the note and when you close the graph and open it again, the note will be back to the version you had when the graph was last saved.

The only way you can save the edit is by either making a change on the graph to trigger having unsaved changes, or by adding a new note.

### Steps to Reproduce

1. Open a graph with at least one user note (alternatively create a graph, add a note and then save the graph)
2. With the Notes View open, edit one of the user notes in some way (change the text, change the colour) and press ok
3. Note that the graph doesn't highlight as having unsaved changes so Ctrl-S won't do anything
4. Close the graph and re-open it

**Expected behaviour:** Changes made to note should be visible

**Actual behaviour:** Changes made to note are lost

**Reproduces how often:** 100%

GammaVel:
Thank you @altair1673 for completing these changes, they are working extremely well. When I edit a note and then save it the graph is showing up as unsaved and warns me when I go to close it. Works with notes for the graph and notes for a selection.

PR #1850, opened 2023-06-13
Notes View graph save

altair1673:
### Prerequisites

### Description of the Change

This is a quick fix to make sure that when the user saves a note the graph needs to be saved as well. This way user changes to notes are not suddenly lost when they close consty.

Currently notes do not get saved to file when closing a graph because the user isn't given the option to save. But now thats fixed and the graph save dialog is set to trigger after a note is saved. 
### Benefits

Users can save notes to file using Ctrl+S instead of having to make changes to the graph itself to trigger the save event.
### Possible Drawbacks

The fix seemed simple enough and nothing should be this easy, drawbacks will appear after testing.
### Verification Process

Made changes to note and then saved it. The title of the graph at the top highlights and when closing consty the save dialog appears.
### Applicable Issues

#1835

### DIFF
--- a/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
+++ b/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
@@ -1022,6 +1022,14 @@ public class NotesViewPane extends BorderPane {
                 noteButtons.getChildren().addAll(showMoreButton, gap, editTextButton, deleteButton);
                 noteButtons.setSpacing(DEFAULT_SPACING);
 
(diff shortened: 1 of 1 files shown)


antares1470 (changes requested):
it appears saving isn't possible after adding a new note.
Steps to reproduce:
1) Open a new graph, add a node
2) Open notes view
3) Press Ctrl-S and verify the save dialog pops up. Go to File -> Save and see it is enabled
4) Add a new note
5) Try repeating step 3. Note when you go to File -> Save it has been disabled

antares1470 (approved):

Delphinus8821 (approved):

Issue #1067, opened 2021-05-28
More flexibility in overriding preset blaze colours via Add Custom Blaze


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When you add a blaze via Add Custom Blaze, you get the option to set it as a preset. If you have already filled up all 10 presets though, it by default replaces the 10th preset and does so without warning. Rather than that, I think it would be a better experience to notify the user if they need to replace an existing preset and then allow them to choose either which preset they want to replace or cancel the request to make the chosen custom blaze colour a preset.

### Steps to Reproduce

1.  Open a non-empty graph
2.  Ensure all of the slots for preset blaze colours are filled
3.  Select a node and go Selection > Add Custom Blaze (or use any other way to select Add Custom Blaze)
4. Pick a blaze colour (ideally different to all of your presets so you can see the change) and select that you want to make it a preset
5. Observe the preset list

**Expected behaviour:** Custom colour automatically replaces 10th preset colour with no warning

**Actual behaviour:** User has option to choose which preset gets replaced

**Reproduces how often:** 100%

GammaVel:
HI @altair1673, I tried this out in the latest nightly build and I didn't get the extra dialog box, the blaze was being replaced but there wasn't any additional message. Can you please check with the latest nightly build?  Thanks.

altair1673:
I don't think this has even been merged in.

GammaVel:
Looks like I was too eager to look at this one the first time :-)   I tried this again with the latest nightly build and it is working well, I had a little chuckle when I saw the name of the message box!
![image](https://github.com/constellation-app/constellation/assets/83679004/46af5eb0-7c49-4638-8e9f-813e61a4d2ab)

Thanks @altair1673 for adding this handy little bit of customisation.

PR #1856, opened 2023-06-22
Blaze colour replacement dialog.

altair1673:
### Prerequisites

### Description of the Change

When there is already 10 preset blaze colours and the user wants to add in a new preset, a dialog window will popup asking the user which preset they would like to replace with their new one.

Currently the last blaze colour preset would be replaced with the new one with no warning which isn't ideal since if that last preset colour is important it'll be lost without giving the user a choice.
### Benefits

Gives user more control over which preset they would like to replace.
### Possible Drawbacks

Dialog windows are annoying to work with in JavaFX and making changes in the code can have unexpected effects.
### Verification Process

Saved 10 blaze colours as presets. Then set 10 new presets by replacing the existing ones via this new poop-up window.
### Applicable Issues

#1067


antares1470 (changes requested):
Otherwise looks good in testing

Delphinus8821 (approved):

antares1470 (approved):

Issue #1861, opened 2023-07-12
Incorrect undo caption `Process Gestures` for plugins run in the same thread


Auriga2:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
For a graph action performed with a plugin run in the same thread, the undo caption is incorrect as it's always `Undo Process Gestures`
E.g. Create Transaction 

### Steps to Reproduce

1. Open a new graph, add vertex and check Edit->undo which will be `Undo Create Vertex`

2. Add a transaction and check Edit->undo which will be `Undo Process Gestures`

 Create Vertex: 
![image](https://github.com/constellation-app/constellation/assets/64826928/184cd580-ca27-4180-bcf7-5663f183a97e)
 
Create Transaction :
![image](https://github.com/constellation-app/constellation/assets/64826928/e464617b-1df5-437b-8bb2-0d1ebf28d816)

**Expected behaviour:** Undo caption should be `Undo Create Transaction `

**Actual behaviour:** Undo caption is `Undo Process Gestures`

**Reproduces how often:** 100%

### Additional Information

Blocks https://github.com/constellation-app/constellation/issues/1141

Auriga2:
It's the `executeNow` in the `DefaultInteractionEventHandler.createTransaction` causes it. Changing it to `.executeLater(graph) `similar to ` DefaultInteractionEventHandler.createVertex` fixed it. But not sure of the consequences since I am not using the `GraphWriteMethods` anymore, specially refering the param descrition quoted below. Everything seems working fine with that though. All other plugin execution calls use `executeLater` so apparently this is the only occurrence of this issue. 

(changing the Add vertex to run in the current thread created the same caption Process gestures)

```
     * @param wg Write access to the graph, corresponding to the lock on which
     * gestures are currently being processed.
```

Auriga2:
In the current thread the actual undo operation is called by the Flush. 

[...]

PR #1863, opened 2023-07-17
Fix the incorrect caption `Undo Process Gestures` to `Undo Create Transaction`

Auriga2:
### Prerequisites

### Description of the Change
Fixed the undo caption of `Create Transaction` which was incorrect as  `Undo Process Gestures`, by executing it later rather than in the current thread, similar to Create Vertex, as per the discussions.

Removed the explicit call to flush in the `DefaultInteractionEventHandler.startHandlingEvents` only for Create Transaction, but kept the other half of the `||` operation JIC. Or can we remove all that code?

### Alternate Designs
In SimpleEditPlugin-> public void run(final GraphWriteMethods graph,......) , somehow set a LockingManager.LockingEdit when the plugin is run. 

Bug fix

### Benefits
Better UX

### Verification Process
Open a new graph, add a transaction and check Edit->undo which should be `Undo Create Transaction `

### Applicable Issues
https://github.com/constellation-app/constellation/issues/1861


antares1470 (approved):

OrionsGuardian (approved):

Delphinus8821 (approved):

GammaVel:
Great to see this working @Auriga2!  Thanks for looking closely into all the underlying issues with the code. The correct message is shown when going to undo a transaction now. 
image

PR #1887, opened 2023-08-31
fixed some potential issues around data access category ordering

antares1470:
### Prerequisites

### Description of the Change

Add some fixes to potential issues that could arise in extending Constellation.

For instance, it appeared the code was assuming that any categories being looked at were one word with no spaces so a category such as 'My New Category' could disappear after ordering as it would be looking for 'My' rather than 'My New Category'. Also appeared the ordering was no longer looking at the category position for default ordering so this has been fixed.

Code in Core still needs to account for extending the application beyond Core.

### Benefits

Fixes potential issues with data access category ordering

### Possible Drawbacks

Nil

### Applicable Issues

#1682

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanel.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/DataAccessViewCategoryPanel.java
@@ -17,6 +17,7 @@ package au.gov.asd.tac.constellation.views.dataaccess.panes;
 
 import au.gov.asd.tac.constellation.utilities.text.SeparatorConstants;
 import au.gov.asd.tac.constellation.views.dataaccess.plugins.DataAccessPlugin;
+import au.gov.asd.tac.constellation.views.dataaccess.plugins.DataAccessPluginType;
 import au.gov.asd.tac.constellation.views.dataaccess.tasks.LookupPluginsTask;
 import au.gov.asd.tac.constellation.views.dataaccess.utilities.DataAccessUtilities;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


OrionsGuardian:
There is an issue when you first run Constellation from a clean build:
image

If you have not opened the Data Access View, and there is no currently saved properties file for the Data Access View settings, then the Options screen will be blank.

antares1470:
> There is an issue when you first run Constellation-Adaptors from a clean build: image
> 
> If you have not opened the Data Access View, and there is no currently saved properties file for the Data Access View settings, then the Options screen will be blank.

I can confirm this. I also got this doing the same for core only so that suggests to me that there is something wrong with the way it is finding the categories

antares1470:
@OrionsGuardian I've resolved this by generating the list if DAV hasn't been opened (and no order preferences exist). The issue was that the function in `LookupPluginsTask` was only being called from the `DataAccessPaneState` (i.e. wasn't going to be called before Data Access View was opened).

The catch with this change is that it does make opening the panel take a little longer if it has to generate the list but I couldn't find any way around it (you can quickly get all possible categories, but ultimately you need to look at all data access plugins to determine which ones are and aren't being used). This timing issue isn't a problem once Data Access View has been opened or the preferences exist.

OrionsGuardian:
> @OrionsGuardian I've resolved this by generating the list if DAV hasn't been opened (and no order preferences exist). The issue was that the function in `LookupPluginsTask` was only being called from the `DataAccessPaneState` (i.e. wasn't going to be called before Data Access View was opened).
> 
> The catch with this change is that it does make opening the panel take a little longer if it has to generate the list but I couldn't find any way around it (you can quickly get all possible categories, but ultimately you need to look at all data access plugins to determine which ones are and aren't being used). This timing issue isn't a problem once Data Access View has been opened or the preferences exist.

This seems fine to me. It's not a frequently used screen.
There's only a very small delay in loading the screen for the first time, which I would consider to be acceptable.

OrionsGuardian (approved):

Delphinus8821 (approved):

Issue #1860, opened 2023-07-10
Rename star graphs from client


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

It would be nice to be able to specify or change Constellation graph title names within scripts using the client to improve efficiency when using something Jupyter Notebooks to generate multiple Constellation graphs

### Steps to Reproduce

1.  \[First Step\]

2.  \[Second Step\]

3.  \[and so on…\]

**Expected behaviour:** Be able to rename from script

**Actual behaviour:** Have to rename afterwards via normal means (e.g. File explorer)

**Reproduces how often:** 100%

Auriga2:
@antares1470 Jupyter notebook uses new_graph end point to create a new graph. We can add an optional parameter to specify the graph name there when creating the graph. In absence of that, it'll be the current behavior- generate the graph with the default name `analytic graph1` etc.

Other than that do we need a separate endpoint to rename a graph? If so which parameters would the user specify :

- graph ID? 
- The fully qualified path of the .star file? 
- graph name/The default name `analytic graph1` etc.? 
- Use the current active graph?

HolandCrick:
@Auriga2 Thanks for adding the enhancement to change Constellation graph title names within using Jupyter scripts. It works as expected. 

![image](https://github.com/constellation-app/constellation/assets/157770877/632221a8-aa05-470b-ba4d-a48424c18928)

PR #1913, opened 2023-11-06
Rename graph through API

Auriga2:
### Prerequisites

### Description of the Change
Jupyter notebook users can now specify or change Constellation graph title names within scripts using the client.

When creating a new graph (`new_graph`) the graph name can be specified as an optional parameter. In absence of that, it'll generate the graph with the default name, `analytic graph1` etc.

There's also a separate endpoint to rename a graph specified by the graph ID (`rename_graph`). If a graph ID is not specified, the current active graph is renamed. It returns the id, previous name and the new name of the graph.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/1860


CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/RenameGraph.java
```diff
@@ -0,0 +1,147 @@
...
+            dataObject.rename(newGraphName);
+
+            final ObjectMapper mapper = new ObjectMapper();
+            final ObjectNode root = mapper.createObjectNode();
+            root.put("id", graphNode.getGraph().getId());
+            root.put("previous_name", previousGraphName);
+            root.put("new_name", graphNode.getDisplayName());
+            mapper.writeValue(out, root);
+
+        } catch (IOException ex) {
```
antares1470, line 143: final

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/services/RenameGraph.java
```diff
@@ -0,0 +1,147 @@
...
+            Exceptions.printStackTrace(ex);
```
antares1470, line 144: can you change this to a log entry?

antares1470 (changes requested):
I think in order for this to work, you also need to add the needed functions to the client code itself (`constellation_client.py`). There are functions in there for each of the services which call the relevant `callService` function. You could probably use those to help in adding and updating the ones you need here

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
```diff
@@ -20,7 +20,7 @@
 # For example, if a new function is added, clients that require that function
 # to be present can check the version.
 #
-__version__ = 20200326
+__version__ = 20200327
```
antares1470, line 23: I think this was a datetime stamp :)

antares1470 (changes requested):
If you could possibly add a whatsnew entry for this as well. That would be great

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
Auriga2: ` Increase this version number when a change happens.` This comment made me think I should manually increase it. Will remove.

antares1470: I think its a manual change. I was mainly noting that the version number is actually a datetime stamp probably representing when it was last changed (i.e 2020-03-26). So suggesting the version number should probably be changed to something more like 20231115 for instance

antares1470 (approved):

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
```diff
@@ -564,23 +564,45 @@ def open_graph(self, filename):
...
+
+        :returns: The id, previous name and the new name of the graph.
+        """
+
+        params = {}
+        if graph_id:
+            params['graph_id'] = graph_id
+        
+        params['new_graph_name'] = new_graph_name
+        graph = self.call_service('new_graph', verb='post', args=params).json()
```
Delphinus8821, line 602: Should this be rename_graph, not new_graph? Changing it to that when testing worked but it threw an error when it was new_graph

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
```diff
@@ -564,23 +564,45 @@ def open_graph(self, filename):
...
         params = {}
         if schema_name:
             params['schema_name'] = schema_name
+        if graph_name:
+            params['graph_name'] = graph_name
         graph = self.call_service('new_graph', verb='post', args=params).json()
 
         return graph['id']
 
+    def rename_graph(self, graph_id=None, new_graph_name):
```
OrionsGuardian, line 588: This line causes an error during testing ...
The order of the parameters needs to be changed so that values with defaults are at the end:
`def rename_graph(self, new_graph_name, graph_id=None):`

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
```diff
@@ -564,23 +564,45 @@ def open_graph(self, filename):
...
+        """
+
+        params = {}
+        if graph_id:
+            params['graph_id'] = graph_id
+        
+        params['new_graph_name'] = new_graph_name
+        graph = self.call_service('new_graph', verb='post', args=params).json()
+
+        return graph['id', 'previous_name', 'new_name']
```
OrionsGuardian, line 604: This line causes an error during testing ...
It doesn't like having multiple field names listed.
We can simply replace the line with:
`return graph`
This will return all the fields in the graph object in json format.
(which happens to be only those 3 fields that were listed)

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
OrionsGuardian: Confirming, when testing the functionality, this does need to be changed to `rename_graph` for it to work.

OrionsGuardian (changes requested):

Delphinus8821 (approved):

OrionsGuardian (approved):

PR #1920, opened 2023-11-23
Bug Fixes for the Analytic View State

Delphinus8821:
### Prerequisites

### Description of the Change
Fixed null pointer exception for old graphs.
Fixed bug when having the Conversation View open when closing the Analytic View which has some visualisations active causes Constellation to hang.
Fixed array index out of bounds exception for the hide visualisation when switching between graphs of different sizes.

Bugfix in core required for the new release. 

### Verification Process
Old Graph Null Pointer:
1. Open a graph from v2.10 which has had an analytic run on it before
2. Create a new graph
3. Run an analytic on the graph from version 2.10
4. Swap to the new graph and observe the null pointer no longer occurs

Conversation View Bug:
1. Open the Conversation View and at least 2 graphs
2. Run an analytic on each graph and selection at least one of the toggles on each graph
3. Close the Analytic View 
4. Observe the application is still responsive

Hide Visualisation Error:
1. Create a new sphere graph and another new graph with only a few nodes and transactions
2. Run an analytic on the sphere graph
3. Swap to the smaller graph and observe there is no array index out of bounds exception thrown

### DIFF
--- a/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewPane.java
+++ b/CoreAnalyticView/src/au/gov/asd/tac/constellation/views/analyticview/AnalyticViewPane.java
@@ -255,7 +255,7 @@ public class AnalyticViewPane extends BorderPane {
                 final boolean resultsVisible = state.isResultsPaneVisible();
                 final AnalyticQuestion<?> question = state.getQuestion();
 
-                if (results != null && resultsVisible && !viewPane.getChildren().contains(analyticResultsPane)) {
+                if (results != null && resultsVisible && !viewPane.getChildren().contains(analyticResultsPane) && question != null) {
                     viewPane.getChildren().add(1, analyticResultsPane);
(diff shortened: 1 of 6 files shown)


OrionsGuardian (approved):
This fixes most issues, so will approve these changes. 
However, there is still a hanging issue occurring when closing the Analytics View while the Conversation View is open.

Steps to make Constellation hang:
Launch Constellation.
Close the Analytics View and close the Conversation View.
Open 3 existing graphs.
Open the Analytics View and the Conversation View.
Run the default Analytic on each graph, and click on the size button.
Create a new Empty graph (don't do anything with this graph).
Switch back to one of the Pre-existing graphs.
Close the Analytics View.

Constellation hangs.

I suspect the problem is specifically with the Conversation View, and should probably be addressed as such.

antares1470 (approved):
Similar to @OrionsGuardian will approve for the other bugs that have been resolved but note the Conversation View one is still popping up (it doesn't happen all the time but it does happen). I'm not sure yet whether that issue is purely with the Conversation View or whether the Analytic View closing plays any part in it. Either way it can be addressed separately

Issue #1924, opened 2023-11-30
`PluginParametersSwingDialog` buttons are not customizable


capricornunicorn123:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
`DialogDisplayer`s (Specifically `PluginParametersSwingDialog`s) have default button values of "OK and "Cancel". Although generally fitting with all use cases, it would be better if these button labels were suited to the specific use case of the `DialogDisplayer`. 

### Steps to Reproduce
1.  Open new graph
2.  Build Sphere graph (note the "OK", "Cancel" buttons)
3.  Select OK.
4.  Right Click on the graph tab and select rename (note the "OK", "Cancel" buttons)

**Expected behaviour:** 
Build Graph "OK" button would be better served as "Build"
Rename Graph "OK button would be better served as "Rename"

**Actual behaviour:**
All acceptance buttons default to "OK"

**verification process:**
Execute Plugins interactively and note the label of the acceptance button.
Some known examples:
  Experimental > Build > Build Spere graph.
  File > Export> To Screenshot.

**Reproduces how often:** 
100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
The Export Action button text is right in all types except a) Screenshot and b) JSON file type.

![image](https://github.com/constellation-app/constellation/assets/157770877/6ca5d42b-d35e-49c0-975c-b7991292c488)

capricornunicorn123:
The dialog windows in the screenshot a) and screenshot b) are FileChoserDialog windows and are not PluginParameterSwingDialog windows. As a result these acceptance options are not customizable for screenshot a) and b). 

The comment is still valid but would be out of scope for this ticket and should be captured in a separate issue. Keeping in mind that the current implementation of FileChooserDialogs would not be able to support this feature, the ticket will likely go stale until a new FileChoserDialog implementation is created.

HolandCrick:
@capricornunicorn123 - 
[...]

PR #1925, opened 2023-11-30
Enable `PluginParametersSwingDialog` Dynamic Approval Option Text

capricornunicorn123:
### Prerequisites

### Description of the Change

`PluginParametersSwingDialog`'s `DialogDisplayer` window will now adjust the acceptance button text to reflect the action word in the `DialogDisplayer` title. for Example setting the title to Build Sphere Graph will result in an acceptance button of "Build"

### Alternate Designs
It was considered allowing developers to specify the action button title at the `createPluginParameters()` method. The necessary passing of a string through the path the `Plugin` > `PluginParameters` >`DefaultPluginEnvironment` > `DefaultPluginInteraction` > `PluginParametersSwingDialog`  >  `DialogDisplayer` was deemed too convoluted when a similar outcome could be generated automatically based off of the title `PluginParametersSwingDialog` title.

It is a core feature.

### Benefits
Buttons will have clearer indication of their function to users.

### Verification Process
Build a sphere graph | Experimental > Build > Build Sphere Graph (Note the Build Button)
Export to GeoJSON | File > Export > to GeoJSON (Note the Export Button)

### Applicable Issues

#1924


CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/PluginParametersSwingDialog.java
```diff
@@ -174,6 +231,38 @@ public String getResult() {
...
+    }
+    
+    /**
+     * Checks to see if the acceptance button was selected.
+     * 
+     * @param result
+     * @return 
+     */
+    public static boolean isAccepted(final String result) {
+        for (final String keyWord : acceptanceKeyWords){
```
antares1470, line 258: wonder if this function could benefit from only checking `result`? That way it only needs to check one value (which should hopefully contain the correct value) rather than re-iterating through all the possible options again to find it.

capricornunicorn123: The object attribute result and the parameter result are essentially the same reference (silly implementation on my part). As the result could be "Cancel" evaluating the attribute against the parameter would return true when we want false. I've has a look at the implementation and decided on the following change: Instead of returning the string value of the button as the result when clicking the acceptance button, we return "OK" no matter the label of the button. That way we only need to check that the result is "OK" to verify if the button was pressed.

Delphinus8821 (approved):

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/PluginParametersSwingDialog.java
```diff
@@ -143,26 +151,79 @@ public void setSize(final Dimension dimension) {
...
+     * The acceptance option is "OK" by default.
+     * The rejectionoOption is "Cancel" by default. 
+     * If a keyword is present in the title of the DialogDisplayer
+     * the acceptance option will adjust dynamically to express that keyword 
+     * (i.e. "Build", "Save", "Export")
+     */
     public void showAndWaitNoFocus() {
-        //Having 'No' button as initial value means focus is off of 'OK' and 'Cancel' buttons
-        final DialogDescriptor dd = new DialogDescriptor(xp, title, true, DialogDescriptor.OK_CANCEL_OPTION, DialogDescriptor.NO_OPTION, null);
+        final DialogDescriptor dd = createDialogDescriptor(true);
```
antares1470, line 185: it seems this is meant to pass false instead rather than true

antares1470 (changes requested):

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/PluginParametersSwingDialog.java
capricornunicorn123: Good catch, thanks :) Copy and pasting oversight on my part when I was working on the code below

antares1470 (approved):

OrionsGuardian:
I'd like this functionality to be enhanced a little further ....

**PluginParametersSwingDialog.java**
```
    private String acceptanceText = "";

    public PluginParametersSwingDialog(final String title, final PluginParameters parameters, final Set<String> excludedParameters, final String helpID) {
        this(title, parameters, excludedParameters, helpID, "");
    }
    

    public PluginParametersSwingDialog(final String title, final PluginParameters parameters, final Set<String> excludedParameters, final String helpID, final String acceptanceText) {
        this.title = title;
        this.acceptanceText = acceptanceText;

```

Add an acceptanceText parameter to the constructor to allow any text to be supplied.

Then go through in an order of preference to find which text to use for the acceptance button:
```
    private String getAcceptanceButton() {
        if (!StringUtils.isBlank(acceptanceText)) {
            return acceptanceText;
        }
        for (final String keyWord : acceptanceButtonLabels){
            if (StringUtils.containsIgnoreCase(title, keyWord)){
                return keyWord;
            }
        }
        return "OK";
    }
```

capricornunicorn123:
> I'd like this functionality to be enhanced a little further ....

Sure thing, this will be useful for devs to overwrite the dynamic approval option to a static option when relevant. There are not many standalone uses of `PluginParametersSwingDialog` (the custom blaze dialog for example) so I'll assume enabling this functionality for future use will be sufficient.  

Let me know if you've come across any existing instances of `PluginParametersSwingDialog` where a static approval option would be better suited rather than a dynamic one and I can update them as a part of this ticket.

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/PluginParametersSwingDialog.java
```diff
@@ -234,19 +236,22 @@ public String getResult() {
...
-     * the acceptance button text at the plugin creation level.
+     * If developers have specified an acceptance button text, this specified text is returned.
+     * If no text was specified a key word is extracted from the title and returned. 
+     * If no word key word is extracted, "OK" is returned
      * 
-     * @param title
      * @return 
      */
     private String getAcceptanceButton() {
+        if (!StringUtils.isBlank(acceptanceText)) {
```
antares1470, line 246: `StringUtils.isNotBlank`

antares1470 (changes requested):

Delphinus8821 (approved):

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/gui/PluginParametersSwingDialog.java
capricornunicorn123: @antares1470, I resolved this change and mustn't have pushed my commit. Sorry about that.. Should be all good to go once sonar cloud passes :)

antares1470 (approved):

Issue #1767, opened 2022-09-28
Using Jupyter notebooks to import data - improve case sensitivity


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Currently if you use Using Jupyter notebooks to import data you need to have the entities in exactly the right case or they will not get recognised. For example - person has to be "Person" and not "person" or "PERSON". If it is in uppercase it won't be recognised.  The issue arises if you are using another program to create the data it could create it as PERSON which would then need to be manually updated before importing.  

### Steps to Reproduce

1.  Refer to description above for details on what is happening here

**Expected behaviour:** When using Juypter notebooks Constellation automatically recognises entity keywords regardless of what casing the letters have. Mixed case, lower case, uppercase should all be recognised. 

**Actual behaviour:**When creating the data for import you need to set it up to match Upperletter followed by lower letters. 

**Reproduces how often:** 100%
### Additional Information

NA

HolandCrick:
Thanks to Guilty-Spark-343; altair1673 and antares1470 for raising and resolving the importing case sensitive data using Juypiter Note Book. It works well now as expected. 

![image](https://github.com/constellation-app/constellation/assets/157770877/7f738f25-38b7-41db-8286-4635a4763a16)
![image](https://github.com/constellation-app/constellation/assets/157770877/736342e2-3267-46c1-8639-dcad29dc8c34)
![image](https://github.com/constellation-app/constellation/assets/157770877/1839750b-97a4-4e22-979d-61952015e10e)
![image](https://github.com/constellation-app/constellation/assets/157770877/225d18ff-fffe-470f-803b-5122bcfbb6d7)

PR #1934, opened 2023-12-09
Feature/rest case sensitivity 1767

Guilty-Spark-343:
### Prerequisites

- [ X] Reviewed the [checklist](CHECKLIST.md)

### Description of the Change

As outlined in issue #1767 when importing from Jupyter or any other external system using the REST API the types must be the exact same case as the Schema type otherwise they won't be recognized and it'll cause duplicate nodes with slightly different type.  The example given in issue #1767 was, person has to be "Person" and not "person" or "PERSON".

The case for the type varies and can also be extended upon (below) so this implementation will attempt a case insensitive match between the type and the list of available types.  Currently the 'Label', Idenitifer', and 'Type' fields will be normalized when added to the graph.
image

Input script - Note the type should be 'Email' but 'email' and 'EMAIL' have been provided to demonstrate the bug.
image

Before the change:
image

After the change:
image
Notice the type is also recognized: 
image

### Alternate Designs

This could have also been implemented in the constellation client however this would fix the issue when other applications outside of jupyter access the REST API.  In addition the list of types isn't exposed through the web server so changes would have to be made server side anyway. 

This is a fix to the current issue of not recognizing types if they are the same case and the schema type.  

### Benefits

This will help keep a standard of types and will also deduplicate nodes that have different case types.

### Possible Drawbacks

None
### Verification Process
1. Open Constellation
2. Start Rest Server - Tools > Start Rest Server
3. Start Jupyter - Tools > Start Jupyter Notebook
4. Download the following notebook and add to your current instance of Jupyter - https://github.com/constellation-app/constellation-training/blob/master/Analyst%20Training/Exercise%2010%20-%20Network%20Analysis%20With%20Python/notebooks_and_constellation.ipynb
5. Change the types to mixed case, capitals, or lowercase and push the dataframe to constellation.
6. Verify the types have been normalized in the graph.

### Applicable Issues

#1767

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/processing/GraphRecordStoreUtilities.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/processing/GraphRecordStoreUtilities.java
@@ -88,7 +88,7 @@ public class GraphRecordStoreUtilities {
             "source.Type",
             "destination.Type"
     );
-    private static final List<String> ApprovedTypes = SchemaVertexTypeUtilities.getTypes().stream().map(i -> i.getName()).collect(Collectors.toList());;
+    private static final List<String> ApprovedTypes = SchemaVertexTypeUtilities.getTypes().stream().map(i -> i.getName()).collect(Collectors.toList());
 
 
     private static int addVertex(final GraphWriteMethods graph, final Map<String, String> values,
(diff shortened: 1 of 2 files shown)


Guilty-Spark-343:
Wondering if this should be more general and belong in SchemaVertexTypeUtilities.getType()?

Delphinus8821 (approved):

antares1470 (approved):
Agreed. If it can be built into the framework that's probably the better place to put it (so the application as a whole can benefit, not just the client)

Issue #1919, opened 2023-11-22
Reset Data Access Category Preference Lists


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Being able to choose which Data Access Plugin Categories to display and in which order is helpful but I think it might be helpful to have an option to reset everything if desired (i.e. all the categories in the visible list in the default order). Currently if I want to do that, I either have to do a bunch of playing around trying to get everything into that order (which could get tedious), or I need to use the Reset script (which requires Constellation closed and will reset other things as well)

**Expected behaviour:** Some sort of Reset to Default button that will put all the categories back in the default position

**Actual behaviour:** Doing this manually or outside the application itself

HolandCrick:
Thanks to @Guilty-Spark-343 for adding the "Reset Data Access Category". It is useful feature and works very well. 
![image](https://github.com/constellation-app/constellation/assets/157770877/4812fb17-eb7c-4303-b3e4-04dde35fcda4)

PR #1937, opened 2023-12-17
Enrichment/restore default dataview categories #1919

Guilty-Spark-343:
### Prerequisites

### Description of the Change

As mentioned in issue #1919 here is a current feature that allows the user to change the visible categories in the data view and change the order of the categories however resetting this takes some effort.  This enrichment simply adds a button, "Restore to Default",  to reset the dataview categories to the default which is all categories in the visible list in the default order.

image

### Alternate Designs

None

This is a small change that will make the user experience better by saving time restoring to the defaults.

### Benefits

Saving time for the user to restore the categories to the default settings.

### Possible Drawbacks

None

### Verification Process

1. Open Setup > Options. Click the Constellation tab, then data access view tab.
2. Change the order of the categories and change the visible and hidden categories.
3. Click 'Restore to Default' 
4. Check the hidden categories should be empty and the visible categories should include all categories in the default order.
5. Close and Open Constellation to ensure the changes have persisted.
### Applicable Issues

#1919


antares1470 (changes requested):
The default order being set looks a little wrong as it should be following the order of positions as specified by the positional data access plugins (e.g. see DataAccessPluginCoreType)
i.e. for Core this would be Import, Clean, Utility, Developer rather than Utility, Import, Developer, Clean

Guilty-Spark-343:
oh good catch! The order wasn't being preserved in the HashMap.  All Fixed and good to go, thanks!

antares1470 (changes requested):
Thanks @Guilty-Spark-343. Looks good now.
One thing I missed from my first review (sorry), if you could add an entry to whatsnew.txt for this, that would be great

Guilty-Spark-343:
updated whatsnew :)

antares1470 (approved):

Delphinus8821 (approved):

Issue #1888, opened 2023-09-01
Allow clicking in the Attribute editor field to untick the no value automatically


GammaVel:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

I needed to add some more attribute values to nodes on my graph using the Attribute Editor. It is very easy to click in the attribute field and the box displays for the field to be edited. I tried to start typing in the field however it didn't work. It wasn't until I realised the No value field was set, and I needed to untick it. 

### Steps to Reproduce

1.  Open any graph and the Attribute editor
2.  Select a node on the graph and observe the attribute editor is populated with values
3. Click on one of the fields to Edit them
4. Immediately think I can start editing the value > I can not > you first need to untick the field for no value. 
>>> To avoid a mouse click, is it possible to automatically untick the no value field so that if a user starts typing it gets turned off.
>>> Alternatively, is this field needed ? The text box field could have no value until the user started typing. 

**Reproduces how often:** 100%

### Additional Information

![image](https://github.com/constellation-app/constellation/assets/83679004/df544cec-4e30-4b47-b8c5-495f8584a290)

GammaVel:
Hi @altair1673, it is fantastic to see how quickly my feedback has been addressed, thanks for making this change!  It is much easier to use. 
There might be an issue with the Reset to Default option.  In version 2.10 when this was ticked, this would grey out the text and automatically select the No value option. This isn't happening with the latest nightly build.  It resets it back to nothing but it doesn't automatically grey out the text and select the No value option.

Here are some screenshots to explain more for the above:
![image](https://github.com/constellation-app/constellation/assets/83679004/505eb8ff-4219-4065-b21d-0e18260f78fe)



It is working as expected when I click on the No value checkbox, no issues with that one. Can you please check with is happening with the Restore default?  
Thanks!

HolandCrick:
[...]

PR #1959, opened 2024-02-19
Allow immediate editing in Attribute Editor editor fields

sol695510:
### Prerequisites

### Description of the Change
Removed the no value checkbox from classes it was present in. The no value checkbox's functionality was also replicated by the Restore Default button in many cases so the checkbox was removed and the Restore Default button was updated to disable editable fields when clicked, as it did for some fields but now for all editors, but also be able to undo the Restore Default action during editing if desired, which also re-enables the editable fields.

### Alternate Designs
N/A

Part of the core functionality of the Attribute Editor.

### Benefits
Allows for immediate editing of editor fields.

### Possible Drawbacks
N/A

### Verification Process
Checked each editor to ensure the correct values are saved with either default enabled or not.

### Applicable Issues
#1888

### DIFF
--- a/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorDialog.java
+++ b/CoreAttributeEditorView/src/au/gov/asd/tac/constellation/views/attributeeditor/AttributeEditorDialog.java
@@ -45,6 +45,7 @@ public class AttributeEditorDialog extends ConstellationDialog {
     private final Button okButton;
     private final Button cancelButton;
     private final Button defaultButton;
+    private final Button undoDefaultButton;
 
     public AttributeEditorDialog(final boolean restoreDefaultButton, final AbstractEditor<?> editor) {
         final VBox root = new VBox();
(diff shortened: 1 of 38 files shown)


antares1470 (commented):
I'm not sure how I feel about removing the No value checkbox altogether. With the attributes types that have them, the reason they have them is that no value is a valid value and so I feel people need the option to change it to that. I know its still there in Restore to Default, but I don't feel like that is particularly intuitive, and it assumes that the the default value is in fact no value which may not necessarily be the case (and if it isn't, you have no way of setting no value)

antares1470:
> I'm not sure how I feel about removing the No value checkbox altogether. With the attributes types that have them, the reason they have them is that no value is a valid value and so I feel people need the option to change it to that. I know its still there in Restore to Default, but I don't feel like that is particularly intuitive, and it assumes that the the default value is in fact no value which may not necessarily be the case (and if it isn't, you have no way of setting no value)

Just to further add, I'm not against the idea of converting No value into a button (perhaps its similar to pressing ok or cancel except when clicked it sets the value to null). I just think No value should be a clear option (even if its not set by default when you open the edit dialog)

OrionsGuardian (changes requested):
image

Delphinus8821 (approved):

antares1470 (approved):

Issue #1961, opened 2024-02-23
Word Cloud View add Scrollbar for smaller display.


HolandCrick:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Word Cloud View: Experimental -> Views -> Word Cloud
If you are running new feature "Word Cloud View" on a smaller screen such as laptop, the "Generate Button" is not displayed on the screen. It goes down and it is not visible. It would be ideal to have a scrollbar to navigate up/down.
### Steps to Reproduce

1.  Create a sphere graph
2.  Experimental -> Views -> Word Cloud
3.  Choose Element Type node
4.  Choose the Attribute Geo.Country2
5.  Change the Threshold to 1 and
6.  Hit Generate
7.  Change Attribute to different value
8. Repeat 4-6 and see the output, then some words will appear in the word cloud like this

**Expected behaviour:** After hit Word Cloud **Generate Button** - the left top window will show the result and bottom section should show the "Work Cloud View Functionality."

**Actual behaviour:** "Work Cloud View Functionality" goes down and there is no scrollbar to navigate up/down to change any attribute and Generate "Word Cloud View' again

**Reproduces how often:** 100% unless you change screen ratio.

### Additional Information
![image](https://github.com/constellation-app/constellation/assets/157770877/f18c2107-2ad0-469d-8745-12aa54799de3)

![image](https://github.com/constellation-app/constellation/assets/157770877/3c9ef017-8bfb-423a-ac1f-3ce3353c6363)

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
Thanks @ Quasar985  for adding scrollbar enhancement in Word Cloud View. It works as expected.

![image](https://github.com/constellation-app/constellation/assets/157770877/3407d628-00e5-4f38-a615-f6d4d5007042)

PR #1979, opened 2024-03-12
Enhancement/wordcloud scrollbar

Quasar985:
### Prerequisites

### Description of the Change

Added scroll bar to the "Generate Content" section of the word cloud view. Also altered height of pane that contains the contents of word cloud to be based on the window size, instead of being a constant value.

### Alternate Designs

It was considered to keep the height as a constant value, however doing so would result in an inconsistent scale of the content page depending on the screen's resolution.

Fixes issue where, depending on the user's screen resolution, the button to generate the word cloud is inaccessible.

### Benefits

Fixes issue related to word cloud generation being inaccessible.

### Possible Drawbacks

The functionality to dynamically set the height of the pane may need tweaking if the word cloud view undergoes major layout changes.

### Verification Process

Verified visually, by changing the window size, and by moving constellation to different monitors.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/1961


CoreWordCloudView/src/au/gov/asd/tac/constellation/views/wordcloud/ui/WordCloudPane.java
```diff
@@ -231,26 +236,30 @@ public WordCloudPane(final WordCloudController controller) {
         noWord.setMaxSize(0, 0);
     }
     
+    public void setEverythingHeight(int h) {
```
Delphinus8821, line 239: parameter final

Delphinus8821 (changes requested):
A couple of things:
If you make the pane narrower than it expects, it cuts off the generate button even when there is room below it to expand to 
image

If you change your windows setting scaling to something quite zoomed in, the generate tab is still off of the screen. 
image

CoreWordCloudView/src/au/gov/asd/tac/constellation/views/wordcloud/ui/WordCloudTopComponent.java
```diff
@@ -91,9 +90,14 @@ public WordCloudTopComponent() {
             final Scene scene = new Scene(wordCloudPane);
             scene.getStylesheets().addAll(JavafxStyleManager.getMainStyleSheet());
             panel.setScene(scene);
+
+            // Update word cloud pane's size when window size changes
+            scene.heightProperty().addListener((obv, oldVal, newVal) -> {
+                wordCloudPane.setContentHeight((int) (double) newVal);
```
antares1470, line 96: I think only one of these type casts needs to occur

antares1470 (changes requested):
Not sure if I'm a fan of the large gap being left at the top as it makes the Generate Word Cloud section look a little messy with its positioning. If possible, it would be good to have it as it was before with the panel showing the actual word cloud only displaying when there is something to be displayed.

wordCloudPanePositioning

Delphinus8821:
> Not sure if I'm a fan of the large gap being left at the top as it makes the Generate Word Cloud section look a little messy with its positioning. If possible, it would be good to have it as it was before with the panel showing the actual word cloud only displaying when there is something to be displayed.
> 


I agree. It also looks a bit odd when there are not many words in the word cloud 

image

antares1470 (changes requested):
On my laptop, once you generate a word cloud, the view doesn't fit horizontally within the default dimensions of the view space (including the Generate Word Cloud section which prior to generating a word didn't suffer from that issue). Perhaps an addition of a horizontal scroll bar would be useful (or better if possible, fitting the view to the default width of the view).

wordCloudHorizontal

Delphinus8821 (approved):

Delphinus8821 (approved):

antares1470 (approved):

Issue #2004, opened 2024-04-08
constellation_client.py pandas.get_graph_attributes(). Error


HolandCrick:
### Prerequisites

- [x ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The constellation_client.py .get_graph_attributes() encounters an error, due to the 'data' argument having a type of <class 'bytes'> instead of being <class 'str'>.

This may be related to a change in newer versions of Python requests or Pandas.

This same issue also exists in the Constellation.get_graph_attributes -> pd.read_json call. #1816

### Steps to Reproduce

Environment: Python 3.10.5, pandas 1.5.3, requests 2.28.1

Follow 'Constellaton-Training (Analyst training) Exercise 10 - Network Analysis With Python' to Part 2 (Getting Data from CONSTELLATION) -> https://github.com/constellation-app/constellation-training/blob/master/Analyst%20Training/Exercise%2010%20-%20Network%20Analysis%20With%20Python/notebooks_and_constellation.ipynb

Running df = cc.get_graph_attributes() df) will hit the error.

**Expected behaviour:**  Pandas correctly parsing the data variable into a graph attributes.

**Actual behaviour:** Pandas throws an error due to the 'file-like object, got <class 'bytes'> 

**Reproduces how often:**   100%

### Additional Information
![image](https://github.com/constellation-app/constellation/assets/157770877/5c782550-b251-4ce6-abe7-28b34590d43c)

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@OrionsGuardian thanks for fixing the  constellation_client.py .get_graph_attributes() . It's working nicely now.
![image](https://github.com/constellation-app/constellation/assets/157770877/d13d74c8-a604-4db2-b6b2-5f0a67fce687)

Issue #1816, opened 2023-03-17
constellation_client.py pandas read_json data type error


rozie1337:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Hello, FYI constellation_client.py `Constellation.get_dataframe` encounters an error (_Expected file path name or file-like object, got type ''_) on the call to `pd.read_json`, due to the 'data' argument having a type of <class 'bytes'> instead of being <class 'str'>.

This may be related to a change in newer versions of Python requests or Pandas.

This same issue also exists in the `Constellation.get_graph_attributes -> pd.read_json` call.

### Steps to Reproduce

1. Environment: Python 3.10.5, pandas 1.5.3, requests 2.28.1

2.  Follow 'Constellaton-Training (Analyst training) Exercise 10 - Network Analysis With Python' to Part 2 (Getting Data from CONSTELLATION) -> https://github.com/constellation-app/constellation-training/blob/master/Analyst%20Training/Exercise%2010%20-%20Network%20Analysis%20With%20Python/notebooks_and_constellation.ipynb

3.  Running `df = cc.get_dataframe()` will hit the error. 

**Expected behaviour:** Pandas correctly parsing the data variable into a dataframe.

**Actual behaviour:** Pandas throws an error due to the 'data' variable being the wrong type.

**Reproduces how often:** 100%

### Additional Information
Can be fixed by calling something like `<variable>.decode("utf-8")` before the call to `pd.read_json` if it is type `bytes`, or by updating the requests call to ensure it returns a string.

HolandCrick:
Hello @Guilty-Spark-343 for addressing the _Json data type error. It works for .get_dataframe but still need to address the  .get_graph_attributes(). The graph attribute section through the error:

![image](https://github.com/constellation-app/constellation/assets/157770877/f6333e6c-dcdb-467b-8fe9-610b669d18b7)

new ticket #2004

PR #2007, opened 2024-04-09
fixed client get_graph_attributes data issue

Guilty-Spark-343:
### Prerequisites

### Description of the Change
Similar to #1933 the constellation_client wasn't decoding the output from the server from bytes to a string to then be used by Panda's read_json as outlined in https://github.com/constellation-app/constellation/issues/1816.  This error occured when calling get_graph_attributes from the client.
Since Constellation opens the default Jupyter instance there can be many different versions of python and pandas accessing the web server using the constellation client.
The change simply checks the data returned from the Constellation web server, if it is a byte type it'll decode it to utf8, as suggested in https://github.com/constellation-app/constellation/issues/1816 . This will not effect the other accepted types such as path object of file buffers.
This has been tested on 3.7.6 (using the older version of Pandas) & 3.11 that uses the new version of Pandas that was erroring.

This is a bug fix for running Jupyter with newer versions of Pandas therefore should be in Constellation core,

### Benefits

The Constellation Core now works with newer versions of python.

### Possible Drawbacks

None

### Verification Process

The changes were tested against both the older versions and newer versions of python / Pandas.
Testing the fix:

1. Install a new python virtual environment
2. Install all relevant libraries used by constellation_client
3. Add the new kernel to a Jupyter instance.
4. Add constellation_client.py as either a relative file or in the system paths (Constellation will do this to the default once it's pushed)
5. Add and run the example notebook to read the current graph - https://github.com/constellation-app/constellation-training/blob/master/Analyst%20Training/Exercise%2010%20-%20Network%20Analysis%20With%20Python/notebooks_and_constellation.ipynb
6. get_graph_attributes should return the graph attributes as a dataframe instead of the error 'error:Expected file path name or file-like object, got <class 'bytes'> type'

### Applicable Issues

#2004

### DIFF
--- a/CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
+++ b/CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
@@ -536,7 +536,7 @@ class Constellation:
         r = self.call_service('get_graph_values', args=params)
         content = r.content
         if isinstance(content, bytes):
-        content = content.decode('utf8')
+            content = content.decode('utf8')
         df = pd.read_json(content, orient='split', dtype=False, convert_dates=False)
         df, self.types = self._fix_types(df)
 
--- a/CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
+++ b/CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
@@ -534,8 +534,10 @@ class Constellation:
             params = {'graph_id':graph_id}
 
         r = self.call_service('get_graph_values', args=params)
-
-        df = pd.read_json(r.content, orient='split', dtype=False, convert_dates=False)
+        content = r.content
+        if isinstance(content, bytes):
+        content = content.decode('utf8')
+        df = pd.read_json(content, orient='split', dtype=False, convert_dates=False)
         df, self.types = self._fix_types(df)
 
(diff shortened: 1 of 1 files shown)


CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
```diff
@@ -534,8 +534,10 @@ def get_graph_attributes(self, graph_id=None):
             params = {'graph_id':graph_id}
 
         r = self.call_service('get_graph_values', args=params)
-
-        df = pd.read_json(r.content, orient='split', dtype=False, convert_dates=False)
+        content = r.content
+        if isinstance(content, bytes):
+        content = content.decode('utf8')
```
antares1470, line 539: I think this line needs to be indented

antares1470 (changes requested):

CoreWebServer/src/au/gov/asd/tac/constellation/webserver/resources/constellation_client.py
OrionsGuardian: Yes, this does need to be indented.
Error when testing with Jupyter Notebook:
```
IndentationError: expected an indented block after 'if' statement on line 538
```

antares1470 (approved):

OrionsGuardian (approved):

Issue #1859, opened 2023-07-10
Table view edges and links


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The current table view only lets you switch between nodes and transactions. Would it be possible to also add edges and links as possible views?

### Behaviour

* Try to include measurable metrics that can be aimed for (if applicable to the issue),
e.g. if it's a time issue then "should take 5 seconds or less to finish"
or if it's a workflow issue "should be accomplished in 2 or less clicks".

-->

**Expected behaviour:** Edge and Link views available

**Actual behaviour:** Only Node and Transaction views available

**Reproduces how often:** 100%

PR #2011, opened 2024-04-11
Add edges and links to table view

Quasar985:
### Prerequisites

### Description of the Change

Added functionality to view edges and links in the table view, as well as the existing transactions are vertices.

### Alternate Designs

For selecting edges/links on the graph and have that selection be reflected in the table view:
This was achieved iterating over the selected transactions on the graph and selecting edges/links in the table view that those transactions represent. An alternate design would be to add functionality to the selection coed that handles edges/links and use that information directly.

### Benefits

Users can now view information about edges and links in table view.

### Possible Drawbacks

Aside from the usual risk of changing code, none.

### Verification Process

Verified visually and through automated testing.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/1859


Delphinus8821 (approved):

antares1470 (changes requested):
1) If you could add a whatsnew entry for this, that would be great
2) Would also be good if you could update the table view help page to include this change
3) From testing, I'm not exactly sure what the intention is to display, but it seems at the moment that if you have more than one value for an attribute in an edge/link (which you might encounter if they consist of more than one transaction), it doesn't display all the values (see the image below)
4) Also from testing, perhaps if the menu list order is modified slightly so that Vertex and Transaction are switched around, that would be good to group Transaction, Edge, and Link together
tableLinkEdgeIssue

antares1470:
> 3\. From testing, I'm not exactly sure what the intention is to display, but it seems at the moment that if you have more than one value for an attribute in an edge/link (which you might encounter if they consist of more than one transaction), it doesn't display all the values (see the image below)

I would note from this that we would likely want to cap the values somewhere (if an edge consists of 50 transactions and each has a unique value for a transaction, you probably don't want to blow out the table with 50 entries in the one cell, maybe)

Quasar985:
> > 3. From testing, I'm not exactly sure what the intention is to display, but it seems at the moment that if you have more than one value for an attribute in an edge/link (which you might encounter if they consist of more than one transaction), it doesn't display all the values (see the image below)
> 
> I would note from this that we would likely want to cap the values somewhere (if an edge consists of 50 transactions and each has a unique value for an attribute, you probably don't want to blow out the table with 50 entries in the one cell, maybe)

I'll look into that.

I also just need some further clarification, because I'm not quite sure what you mean when you say it doesn't show all values for an attribute of something like an edge. In your screenshot it shows that the two types for the transactions in the highlighted edge are 'unknown' and 'correlation'. I can see one of those in the table being highlighted, and that has a transaction.type of 'unknown' so it must be partially correct from what I can see. Or do you mean something else isn't displaying like it normally should?

Quasar985:
> > > 3. From testing, I'm not exactly sure what the intention is to display, but it seems at the moment that if you have more than one value for an attribute in an edge/link (which you might encounter if they consist of more than one transaction), it doesn't display all the values (see the image below)
> > 
> > 
> > I would note from this that we would likely want to cap the values somewhere (if an edge consists of 50 transactions and each has a unique value for an attribute, you probably don't want to blow out the table with 50 entries in the one cell, maybe)
> 
> I'll look into that.
> 
> I also just need some further clarification, because I'm not quite sure what you mean when you say it doesn't show all values for an attribute of something like an edge. In your screenshot it shows that the two types for the transactions in the highlighted edge are 'unknown' and 'correlation'. I can see one of those in the table being highlighted, and that has a transaction.type of 'unknown' so it must be partially correct from what I can see. Or do you mean something else isn't displaying like it normally should?

Wait, I think I understand the problem now. The row in the table view doesn't display the correct data for multiple transactions.

antares1470:
> Wait, I think I understand the problem now. The row in the table view doesn't display the correct data for multiple transactions.

Yes. An edge (and link) can consist of multiple transactions and therefore the edge/link could contain multiple values for a given attribute and so this should probably be reflected in the table view. I'm not exactly sure of the best approach though given as highlighted, too many values may blow out the table more than desired

Quasar985:
> Yes. An edge (and link) can consist of multiple transactions and therefore the edge/link could contain multiple values for a given attribute and so this should probably be reflected in the table view. I'm not exactly sure of the best approach though given as highlighted, too many values may blow out the table more than desired

I'm thinking the table view should display `<Multiple Values>` if that attribute is not the same for all transactions, similar to the attribute editor. If the user wants to view all the different values, they could also use the attribute editor. What do you think?

OrionsGuardian (changes requested):

OrionsGuardian (approved):

Issue #1877, opened 2023-08-08
Integration of generic input field able to facilitate multiple selection use cases


capricornunicorn123:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Input fields that permit multiple discrete inputs are implemented in a number of unique ways throughout Constellation. It would help with promoting code readability and the overall UX if all of these instances could be enabled by a single component. 

In PR #1869 some simple input fields that permit multiple discrete inputs were captured in the new Core Utility component MultiSelectInputField. This field enabled users to select multiple inputs in a CheckComboBox style input field as well as enable bulk selection options such as "Select All" and "Clear All" in a ContextMenu or a selection options MenuButton.

Ideally this MultiSelectInputField should be built upon / adapted / or replaced to provide functionality for input fields that permit multiple discrete inputs.

Known instances include:
- ErrorReportView - Report Settings DropDown
- TimelineView - Source/Destination/Transaction Filter
- MapView - Various Toolbar Inputs

### Additional Information

This Issue relates to Issue #51 PR #1869

andromeda-224:
Replacements:

- [x] Error Report > Report Settings (Filter)
- [x] Plugin Reporter > Filter

HolandCrick:
@andromeda-224 thanks for the enhancement. The select all/clear all option are very handy. In most Plugin it's working fine but some are it's not working, or feature is not added. Can you please review again.

The Analytic View - select all or clear all not working pls see the clip to replicate the issue:
![Image](https://github.com/user-attachments/assets/5ca26777-1bcb-4011-af1b-1f8aebb636c8)

The Conversation View - select all or clear all option is missing: 
![Image](https://github.com/user-attachments/assets/78a5c973-e260-4973-8237-d3747b137c50)

HolandCrick:
@andromeda-224 thanks for the enhancement. It's working nicely. After adding the "Transaction Type Value" the select all / clear all in analytic view works.

Issue #1012, opened 2021-04-14
Ability to type ahead in `MultiChoiceParameterType` drop down lists


Auriga2:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Enable type ahead in the multi choice drop downs.
Single choice drop downs are handled in https://github.com/constellation-app/constellation/issues/347
This requires fixes in controlsfx and there's a ticket in their backlog controlsfx/controlsfx#1160

### Steps to Reproduce
n/a

**Expected behaviour:** Start typing in drop down list and it starts to list options based on what you have typed

**Actual behaviour:**  Multi choice drop down lists are not editable.

**Reproduces how often:** n/a

### Additional Information
n/a

capricornunicorn123:
This Issue has been addressed in #2025

HolandCrick:
@andromeda-224  - Thanks for the enhancement for ability to type ahead in drop down list. It's working well. Start typing in drop down list and it starts to list options based on what you have typed.

![Image](https://github.com/user-attachments/assets/699256fb-963e-427c-be5f-2bb1ce232a9a)

HolandCrick:
Ability to type ahead" enhancement is not active in some places. such as
Attribute Editor -> Transactions -> Click + Sign -> Custom -> Attribute Type -> Click Dropdown list

HolandCrick:
@andromeda-224  - Thanks for the enhancement. The ability to type ahead in drop down list (multi choice) is working fine. If the Attribute Type is a single choice dropdown; then it does not work. The single choice type ahead has not been implemented yet. Relevant Ticket is #2259.

PR #2025, opened 2024-04-22
Input Field Enhancments

capricornunicorn123:
### Prerequisites

### Description of the Change

This PR aims to address limitations with the current implementation of input fields in Constellation. A suite of classes will be introduced to provide a custom implementation of various input field utilized in constellation. This will provide a number of benefits including increasing reusability, increasing visual consistency and increasing the effectiveness of their use.

Each input field will be based off of a "Magic-Wand" Style input field classed as the abstract `ConstelationInputFild`
  
BaseField

This Base `ConstellationInputField` enables a rich framework for facilitating a number of use cases. These Include:
-  **AutoComplete** An auto complete context menu is attempted to be created on each change to the value of a `ConstellationInputField`. All implementation of `ConstellationInputField`'s must be able to provide a list of `MenuItem`'s which have a defines action that modifies the value of the Input Field if the option is selected. This allows for an implementation specific autocomplete functionality.
- **DropDownMenus** A drop down menu can be extended by implementations of the `ConstellationInputField`    

Following the above image, the base class can be adapted and extended to suit any input field utilized in constellation. 

The following rules will be used to ensure consistent appearance and use.

- All input fields will have a text area where free text can be added.
- The text area shall fill the space of the entire input field unless trigger buttons are present.
- Input fields may have 0, 1 or 2 trigger buttons.
- Input fields requiring only one trigger button shall place the button on the right side of the field. 
- Buttons that trigger a context menu shall be grey
- Buttons that trigger a change in the appearance/formatting of the input field shall be grey
- Buttons that trigger a pop-up window shall be blue
- Buttons that trigger a pop-up widow shall be placed on the right.
- Input fields shall have no more than 1 pop-up triggering button.

Below are the new input fields and their current completion level:

**Password Input** - `PasswordInputField` (In Progress)
password

**Single Line Text Input** - `TextInputField(TextType.SINGLELINE)`  (In Progress)
image

**Multi Line Text Input** - `TextInputField(TextType.MULTILINE)` (In Progress)

**File Chooser Input** - `FileInput(FileKind.OPEN)` (In Progress)
image

**Single Choice Input** - `ChoiceInputField(ChoiceType.SINGLE)` (Not Started)
image

**Multi Choice Input** - `ChoiceInputField(ChoiceType.MULTIPLE)` (Not Started)
image

**Color Input** - `ColorInputField` (In Progress)
image

**Number Input** - `NumberInputField` (In Progress)
image

### Alternate Designs

Input fields have historically been extended by adding tooltips and adjacent buttons which extend the functionality of the input field. Although this approach is convenient, the intended use of this buttons is not always clear and can lead to different appearances among related input fields

Input fields are a core feature of Constellation 
### Benefits

Increased interface visual quality
Increasing code reusability
Increasing resilience to change by decoupling the application from external input field providers such as ControlsFX

### Possible Drawbacks
Input fields are used throughout Constellation and this change will require the manual updating of each module.
### Verification Process

The verification process will need to be thorough as this change has the potential to impact every view, and interactive or Data Access Plugin. In most cases the integration of the new input fields into these vies and plugins will be deliberate so the verification process of each oof these changes can be detailed on a case-by-case- basis.  
### Applicable Issues

#1877
#1012 
#1518


OrionsGuardian:
Doing some preliminary testing ...

There's an issue with the Prev/Next number input field which replaces the number spinner component:
image
Clicking on "Prev" or "Next" jumps the value to the extreme endpoint value

Also in the number input field, with a fixed range and fixed increment size, there's a rounding error in the resulting values:
image

Multi-line text field isn't big enough to at least show a full 2 lines of text
image

The dropdown box of values displayed when clicking the "Select" button doesn't extend all the way across the size of the component:
image
You should be able to click the Select button and have clickable values directly below the mouse pointer.

The multi-select dropdown window should also stretch across the full component width, and the Select All / Clear All options should be right justified since they are actions rather than data entries.
image

Typing into the select and multi-select fields should do a case insensitive search for matches, and should match anywhere in the string, not just from the beginning.

image

Also, the background shouldn't turn red while typing. It should only turn red if the component loses focus with an invalid value.

The color input field, I'd suggest using the word "Select" instead of "Swatch" as it is the name of the action you are performing.
image

OrionsGuardian:
The input field for when you can select multiple files doesn't quite look right with an oversized open button:
image

I've mocked up an alternative layout to consider:
image

This will keep the button the same size for all input fields, and the entries can either be in a list, or in a scrollpane.

OrionsGuardian:
Found an issue with the Import From Graph plugin, where it doesn't send the currently selected file type to the File Chooser.
image

Issue #2022, opened 2024-04-18
Sonar Cleanup


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

It's been a bit since there was some effort put in the cleanup some of the issues being flagged by Sonarcloud but it would be good to get some of the numbers down.

### Additional Information

If you're contributing to this, please comment what rule(s) you are looking at so that there isn't any duplication of effort

antares1470:
(Java) Private mutable members should not be stored or returned directly

Delphinus8821:
(Java) Redundant casts should not be used
(Java) Simple class names should be used

capricornunicorn123:
(Java) Unused "private" fields should be removed

Delphinus8821:
(Java) Literal suffixes should be upper case
(Java) Lambdas should be replaced with method references

Quasar985:
(Java) Math operands should be cast before assignment
(Java) Tests should use fixed data instead of randomized data
(XML) Sections of code should not be commented out

Quasar985:
(Java) Empty statements should be removed
(Java) Enum values should be compared with "=="
(Java) Lazy initialization of "static" fields should be "synchronized"

Quasar985:
(Java) Accessing an array element should not trigger an ArrayIndexOutOfBoundsException
(Java) Arrays and lists should not be copied using loops
(Java) Comparators should be "Serializable"

Quasar985:
(Java) The names of methods with boolean return values should start with "is" or "has"
(Java) Unused "private" methods should be removed
(Java) "collect" should be used with "Streams" instead of "list::add"

Delphinus8821:
(Java) Private fields only used as local variables in methods should become local variables

capricornunicorn123:
(Java) Arrays should not be created for varargs parameters

Quasar985:
(CSS) CSS properties should be valid
(Java) "switch" statements and expressions should not be nested
[...]

PR #2028, opened 2024-04-22
Sonar Code Cleanup (remove unused private fields, remove unused parameters, declare parameters as base types)

capricornunicorn123:
### Prerequisites

### Description of the Change

Changes address the following Sonar Rules

**(Java) Unused "private" fields should be removed** - All cases addressed

**(Java) Method parameters should be declared with base types** - All cases Addressed

**(Java) Unused Method parameters should be removed** - the following instances could not be addressed:
- DataAccessViewCategoryPanel (au.gov.asd.tac.constellation.views.dataaccess.panes)
- OperatorRegistry (au.gov.asd.tac.constellation.graph.value)
- HashmodPanel (au.gov.asd.tac.constellation.graph.utilities.hashmod)
- HelpOptionsPanel (au.gov.asd.tac.constellation.help.preferences)
- PluginParametersPane (au.gov.asd.tac.constellation.plugins.gui)
- ApplicationOptionsPanel (au.gov.asd.tac.constellation.preferences)
- GraphOptionsPanel (au.gov.asd.tac.constellation.preferences)

### Applicable Issues
#2022

### DIFF
--- a/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/GraphTaxonomy.java
+++ b/CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/GraphTaxonomy.java
@@ -98,7 +98,7 @@ public final class GraphTaxonomy {
     }
 
     public GraphTaxonomy(final GraphWriteMethods wg, final Map<Integer, Set<Integer>> taxa, final Map<Integer, Integer> nodeToTaxa) {
-        this(wg, taxa, null, Graph.NOT_FOUND, Graph.NOT_FOUND);
+        this(wg, taxa, nodeToTaxa, Graph.NOT_FOUND, Graph.NOT_FOUND);
     }
 
     public GraphTaxonomy(final GraphWriteMethods wg, final Map<Integer, Set<Integer>> taxa, final Map<Integer, Integer> nodeToTaxa, final int singletonKey, final int doubletKey) {
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/GLTools.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/GLTools.java
@@ -488,7 +488,7 @@ public final class GLTools {
     }
 
     // Load a TGA as a 2D Texture. Completely initialize the state
-    public static Texture loadTexture(final GL gl, final InputStream in, final String ext, final int minFilter, final int magFilter, final int wrapMode) throws IOException {
+    public static Texture loadTexture(final GL gl, final InputStream in, final int minFilter, final int magFilter, final int wrapMode) throws IOException {
         // NVS-415: Appears to be a bug in JOGL where texture provider for PNG files does not flip the texture.
         final TextureData data = TextureIO.newTextureData(gl.getGLProfile(), in, false, null);
         final Texture tex = TextureIO.newTexture(data);
--- a/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/GLTools.java
+++ b/CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/GLTools.java
@@ -868,7 +868,6 @@ public final class GLTools {
                 //Do Nothing
             }
         }
-        // Do nothing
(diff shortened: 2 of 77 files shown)


CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/GraphTaxonomy.java
```diff
@@ -98,7 +98,7 @@ public GraphTaxonomy(final GraphWriteMethods wg, final Map<Integer, Set<Integer>
     }
 
     public GraphTaxonomy(final GraphWriteMethods wg, final Map<Integer, Set<Integer>> taxa, final Map<Integer, Integer> nodeToTaxa) {
-        this(wg, taxa, null, Graph.NOT_FOUND, Graph.NOT_FOUND);
+        this(wg, taxa, nodeToTaxa, Graph.NOT_FOUND, Graph.NOT_FOUND);
```
antares1470, line 101: just curious to know what this change addressed as I would assume this would alter functionality

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/GLTools.java
```diff
@@ -853,20 +854,21 @@ public static void checkError(final GL3 gl, final String msg) {
...
-        } else {
-            // Do nothing
+        switch (fboStatus) {
+            case GL.GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT -> errtext = "framebuffer incomplete missing attachment";
+            case GL.GL_FRAMEBUFFER_UNSUPPORTED -> errtext = "framebuffer unsupported";
+            default -> {
+                //Do Nothing
+            }
         }
+        // Do nothing
```
antares1470, line 871: this comment can be removed

CoreOpenGLDisplay/src/au/gov/asd/tac/constellation/visual/opengl/utilities/ShaderManager.java
```diff
@@ -195,58 +196,54 @@ public int useStockShader(final GL3 gl, final int shaderId, final Object... args
...
-            int textureUnit = gl.glGetUniformLocation(stockShaders[shaderId], "textureUnit0");
-            int i = (Integer) args[4];
-            gl.glUniform1i(textureUnit, i);
-        } else {
-            throw new RenderException("Unimplemented shader.");
+        switch (shaderId) {
+            case SHADER_IDENTITY -> {
+                    // Just the color.
+                    int colorLoc = gl.glGetUniformLocation(stockShaders[shaderId], V_COLOR);
+                    float[] color = (float[]) args[0];
```
antares1470, line 203: could you add some finals here and in some of the cases below this?

antares1470 (changes requested):
For some of the public functions that have had parameters altered, it might be helpful to have some entries in the CHANGELOG to let people know since some of them are breaking changes (in the parameter removal cases)

CoreArrangementPlugins/src/au/gov/asd/tac/constellation/plugins/arrangements/GraphTaxonomy.java
capricornunicorn123: This constructor is only referenced in two classes TaxFromTrees and TaxFromNeighbours, both classes go through a lot of effort to generate the nodeToTaxa map and then this constructor simply ignores it. It appears to only effect the Tree Arrangement Plugins which are not overly performant anyway. their functionality is effected by this change however it doesn't become any better or worse, just different. So figured it would be better to assign the variable than to begin unpacking the methods that invoke the constructor (these methods only ever add to the map meaning that removing the parameter would result in code smells in the invoking methods where writing to a map that is never read from would need to be addressed). Happy to remove the value from the constructor and chase up subsequent code smells if you prefer

antares1470: Ok I've taken a closer look at `GraphTaxonomy`. Looking at it, it actually makes no sense for that particular constructor to be parsing null and instead and it should be parsing `nodeToTaxa` -> else there is no point in having it as a parameter (so that does seem like a bug). Curious to know exactly how the behaviour changes but worse case if the behaviour needs to be reverted, it would actually be better simply then using the 2 parameter constructor in those 2 cases as that makes `nodeToTaxa` null anyway (so achieves the exact same thing)

antares1470 (approved):

Delphinus8821 (approved):

PR #2031, opened 2024-04-22
Private mutable getters update

antares1470:
### Prerequisites

### Description of the Change

Started addressing a number of hits for the following sonar rule:

(Java) Private mutable members should not be stored or returned directly

Specifically, I've focused on return instances where the return value is only used in a read-context (i.e. it's never written back to from the return value). More specifically:
- Collection returns now returning unmodifiable instances
- One-dimensional primitive arrays are returning copies of the original

Improves code management of mutable class use

### Benefits

Improves code management of mutable class use

### Possible Drawbacks

Nil

### Applicable Issues

#2022


Delphinus8821 (approved):

antares1470:
Note I've now also include some instances of setters as well

Quasar985 (approved):

PR #2036, opened 2024-04-26
Enhancement/sonar code smell removal 4

Quasar985:
### Prerequisites

### Description of the Change

Sonar fixes for:
(Java) "public static" fields should be constant
(Java) Boolean expressions should not be gratuitous
(Java) Boolean literals should not be redundant
(Java) Execution of the Garbage Collector should be triggered only by the JVM
(Java) Standard outputs should not be used directly to log anything

### Alternate Designs

N/A

### Benefits

Code smell fix.

### Possible Drawbacks

May break existing code.

### Verification Process

Automated tests.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/2022

### DIFF
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/locking/LockingTarget.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/locking/LockingTarget.java
@@ -39,7 +39,7 @@ public abstract class LockingTarget implements ReadingInterface {
         lock.unlock();
 
         if (VERBOSE) {
-            LOGGER.log(Level.SEVERE, "Read lock released by  {0}", Thread.currentThread());
+            LOGGER.log(Level.INFO, "Read lock released by  {0}", Thread.currentThread());
         }
     }
 
--- a/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/locking/LockingTarget.java
+++ b/CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/locking/LockingTarget.java
@@ -19,6 +19,8 @@ import au.gov.asd.tac.constellation.graph.DuplicateKeyException;
 import static au.gov.asd.tac.constellation.graph.locking.LockingManager.VERBOSE;
 import au.gov.asd.tac.constellation.graph.undo.GraphEdit;
 import java.util.concurrent.locks.Lock;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * manages the locking of a graph
@@ -30,12 +32,14 @@ public abstract class LockingTarget implements ReadingInterface {
     Lock lock;
     protected GraphOperationMode operationMode = GraphOperationMode.EXECUTE;
 
+    private static final Logger LOGGER = Logger.getLogger(LockingTarget.class.getName());
+
     @Override
     public void release() {
         lock.unlock();
 
         if (VERBOSE) {
-            System.out.println("Read lock released by " + Thread.currentThread());
+            LOGGER.log(Level.SEVERE, "Read lock released by  {0}", Thread.currentThread());
         }
     }
 
--- a/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/gui/SimpleGraphOpener.java
+++ b/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/gui/SimpleGraphOpener.java
@@ -89,7 +89,7 @@ public final class SimpleGraphOpener extends GraphOpener {
                     time = System.currentTimeMillis() - t0;
                 } catch (final GraphParseException ex) {
                     gex = ex;
-                    LOGGER.log(Level.WARNING, "{0}", ex.getStackTrace());
+                    LOGGER.log(Level.SEVERE, ex.getLocalizedMessage(), ex);
                 }
             }
 
--- a/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/gui/SimpleGraphOpener.java
+++ b/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/gui/SimpleGraphOpener.java
@@ -70,6 +70,8 @@ public final class SimpleGraphOpener extends GraphOpener {
         private final Runnable doAfter;
         private Exception gex;
         private long time;
+        private static final Logger LOGGER = Logger.getLogger(SimpleGraphFileOpener.class.getName());
+        
 
         SimpleGraphFileOpener(final GraphDataObject gdo, final Graph graph, Runnable doAfter) {
             this.gdo = gdo;
@@ -87,7 +89,7 @@ public final class SimpleGraphOpener extends GraphOpener {
                     time = System.currentTimeMillis() - t0;
                 } catch (final GraphParseException ex) {
                     gex = ex;
-                    ex.printStackTrace(System.out);
+                    LOGGER.log(Level.WARNING, "{0}", ex.getStackTrace());
                 }
             }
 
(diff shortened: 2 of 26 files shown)


CoreGraphFramework/src/au/gov/asd/tac/constellation/graph/locking/LockingTarget.java
```diff
@@ -30,12 +32,14 @@ public abstract class LockingTarget implements ReadingInterface {
...
 
+    private static final Logger LOGGER = Logger.getLogger(LockingTarget.class.getName());
+
     @Override
     public void release() {
         lock.unlock();
 
         if (VERBOSE) {
-            System.out.println("Read lock released by " + Thread.currentThread());
+            LOGGER.log(Level.SEVERE, "Read lock released by  {0}", Thread.currentThread());
```
antares1470, line 42: I think the log level should be either `INFO` or even possibly `FINE`

CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/gui/SimpleGraphOpener.java
```diff
@@ -87,7 +89,7 @@ protected Object doInBackground() throws Exception {
                     time = System.currentTimeMillis() - t0;
                 } catch (final GraphParseException ex) {
                     gex = ex;
-                    ex.printStackTrace(System.out);
+                    LOGGER.log(Level.WARNING, "{0}", ex.getStackTrace());
```
antares1470, line 92: Possibly should be `SEVERE`. Also can you change to the following:
` LOGGER.log(Level.SEVERE, ex.getLocalisedMessage(), ex);`

CoreHelp/src/au/gov/asd/tac/constellation/help/ConstellationHelpDisplayer.java
```diff
@@ -60,8 +60,9 @@ public static void copy(final String filePath, final OutputStream out) throws IO
         final String sep = File.separator;
         final InputStream pageInput = getInputStream(filePath);
         final InputStream tocInput = getInputStream(Generator.getBaseDirectory() + sep + Generator.getTOCDirectory());
-
-        if (pageInput == null || tocInput == null) {
+        
+        // Sonar mistakenly thinks pageInput and tocInput can never be null, but they can be null
+        if (pageInput == null && tocInput == null) { //NOSONAR
```
antares1470, line 65: I think sonar might actually be correct here. If you look at the `getInputStream` function which those variables use to generate their values, it never returns null (always a new `FileInputStream`) and so therefore those values won't ever return null

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/DateTimeRange.java
```diff
@@ -185,7 +189,7 @@ public static DateTimeRange parse(final String s) {
 
                 return new DateTimeRange(period, zi);
             } catch (final DateTimeParseException ex) {
-                ex.printStackTrace(System.err);
+                LOGGER.log(Level.WARNING, "{0}", ex.getStackTrace());
```
antares1470, line 192: similar case as previous logger comment

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/DateTimeRange.java
```diff
@@ -206,7 +210,7 @@ public static DateTimeRange parse(final String s) {
                     return new DateTimeRange(zstart, zend);
                 }
             } catch (final DateTimeParseException ex) {
-                ex.printStackTrace(System.err);
+                LOGGER.log(Level.WARNING, "{0}", ex.getStackTrace());
```
antares1470, line 213: same here

CorePreferences/src/au/gov/asd/tac/constellation/preferences/ApplicationPreferenceKeys.java
```diff
@@ -188,9 +188,25 @@ public static String getRESTDir(final Preferences prefs) {
      */
     public static final String FONT_PREFERENCES = "au/gov/asd/tac/constellation/preferences";
     public static final String FONT_FAMILY = "fontFamily";
-    public static String FONT_FAMILY_DEFAULT = "Arial";
     public static final String FONT_SIZE = "fontSize";
-    public static String FONT_SIZE_DEFAULT = "12";
+    private static String fontFamilyDefault = "Arial";
+    private static String fontSizeDefault = "12";
```
antares1470, line 193: I think when I look at this change, my gut questions why the default values need to be changed at all? I can see in code that they get changed (hence they're not final) but given all other preference keys don't ever modify the default value I'm  curious to know the need to alter it for these 2.

What I'm trying to get at is that if the defaults have to be edited then what you have here is fine, but if we can refactor code such that the defaults can be made final, that would probably be my preferred solution here

CorePreferences/src/au/gov/asd/tac/constellation/preferences/GraphPreferenceKeys.java
```diff
@@ -30,7 +30,7 @@ public final class GraphPreferenceKeys {
     public static final String BLAZE_OPACITY = "blazeOpacity";
     public static final int BLAZE_OPACITY_DEFAULT = 100;
     public static final String BLAZE_PRESET_COLORS = "blazePresetColors";
-    public static String BLAZE_PRESET_COLORS_DEFAULT = "#FF0000;#0000FF;#FFFF00;";
+    private static String BLAZE_PRESET_COLORS_DEFAULT = "#FF0000;#0000FF;#FFFF00;";
```
antares1470, line 33: similar case here (from what I can tell it seems the only reason it isn't final is for unit testing purposes)

antares1470 (changes requested):
Regarding the garbage collector smell fixes:
1) Might be worth testing to see how memory is impacted when graph is opened and closed. I note that these preferences were enabled by default so would just want to ensure that the garbage collector still does its job
2) If all is good, given these were preferences in the developer panel, these should probably be removed from there as they no longer serve a purpose

CorePreferences/src/au/gov/asd/tac/constellation/preferences/GraphPreferenceKeys.java
```diff
@@ -49,4 +49,12 @@ public final class GraphPreferenceKeys {
 
     private GraphPreferenceKeys() {
     }
+
+    public static String getBlazePresetsColorsDefault() {
+        return BLAZE_PRESET_COLORS_DEFAULT;
+    }
+
+    public static void setBlazePresetsColorsDefault(String newPreset) {
```
Delphinus8821, line 57: final

Delphinus8821 (changes requested):

CoreHelp/src/au/gov/asd/tac/constellation/help/ConstellationHelpDisplayer.java
Quasar985: I was confused because I managed to get null in some cases. I figured out the problem was one of the automated tests would force the FileInputStream to return null. All fixed now

CorePreferences/src/au/gov/asd/tac/constellation/preferences/ApplicationPreferenceKeys.java
Quasar985: Managed to refactor. For some reason the values were being set on startup...to themselves. So that got removed.

CorePreferences/src/au/gov/asd/tac/constellation/preferences/GraphPreferenceKeys.java
Quasar985: Removed functions

Delphinus8821 (approved):

antares1470 (approved):

Issue #2048, opened 2024-05-17
Overriding open graph files


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Right now, it is possible to save override a graph while it is still open. This really shouldn't be possible and causes a bunch of errors when trying to save the original graph.

### Steps to Reproduce

1. Open a new graph and save with a new name. Keep that graph open
2. Open another new graph and when prompted choose the previously saved graph to trigger override (this is the problematic step)
3. Go back to first graph, make a change and then attempt to save
4. Exception thrown as it can't find the data object (due to step 2 being allowed to happen)

**Expected behaviour:** Constellation should not allow graphs to be overwritten while in use

**Actual behaviour:** Constellation allows overwriting graphs in use, causing errors down the line

**Reproduces how often:** 100%

HolandCrick:
@Quasar985 thanks for fixing the Overriding open graph issue. After triggering override on an open graph, went back to 1st graph, edited the graph and saved successfully  without error. Working as expected.

![image](https://github.com/constellation-app/constellation/assets/157770877/c4c5c239-db99-40f0-ae18-541d840a8a01)

PR #2051, opened 2024-05-21
Fix exception when saving a graph that has been overwritten

Quasar985:
### Prerequisites

### Description of the Change

Added check to see if file being saved is 'valid'. File is invalid if it's been saved over as described in the issue. If file is invalid, the save as menu is opened instead so the user can decide what happens.

Steps to reproduce error (slightly changed wording from original ticket):

1. Open a new graph and save with a new name. Keep that graph open
2. Open another new graph, save it and when prompted choose the previously saved graph to trigger override (this is the problematic step)
3. Go back to first graph, make a change and then attempt to save
4. Exception thrown as it can't find the data object (due to step 2 being allowed to happen)

### Alternate Designs

N/A

Fix bug.

### Benefits

Bug fixed and users have more control over how overwritten files are saved.

### Possible Drawbacks

None.

### Verification Process

Verified with automated testing and manual tests.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/2048


antares1470 (commented):
While this does fix the problem by allowing a save to occur after the fact, I think it would be better, if possible, to address the underlying issue which is the fact that you shouldn't be able to externally modify a star file (e.g. override, delete) while its still in use. If you look at something like a Microsoft Word file, if you try overriding or deleting the file while its open, you will be blocked from doing so. I think something similar for Star files should ideally also occur.

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphTopComponent.java
```diff
@@ -1027,19 +1028,35 @@ private class MySaveAs implements SaveAsCapable {
...
+            }
+
+            final File newFile = new File(folder.getPath(), name + ext);
+
+            // One last check if file were saving over doesn't have it's grpah open UNLESS were saving over the file with the same graph
+            final Path currentFilePath = Paths.get(graphNode.getDataObject().getPrimaryFile().getPath());
+
+            // Check if overwriting open graph
+            final Path filePath = Paths.get(newFile.getPath());
+            for (Graph g : GraphNode.getAllGraphs().values()) {
```
Delphinus8821, line 1043: final

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphTopComponent.java
```diff
@@ -111,19 +109,23 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
```
Delphinus8821, line 121: remove unused import

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphTopComponent.java
```diff
@@ -956,20 +950,27 @@ protected void handleSave() throws IOException {
                 // graph will get saved each time.
                 requestActive();
 
-                SaveAsAction action = new SaveAsAction();               
+                SaveAsAction action = new SaveAsAction();
```
Delphinus8821, line 953: final

Delphinus8821 (changes requested):
Works well

Delphinus8821 (approved):

CoreInteractiveGraph/src/au/gov/asd/tac/constellation/graph/interaction/gui/VisualGraphTopComponent.java
```diff
@@ -956,20 +949,27 @@ protected void handleSave() throws IOException {
...
+                final BackgroundWriter writer = new BackgroundWriter(name, freshGdo, true);
+                writer.execute();
+                isSaved = true;
+            } else {
+                // File might have been saved over, so just open up the 'save as' menu
+                requestActive();
+                final SaveAsAction action = new SaveAsAction();
+                action.actionPerformed(null);
+                isSaved = action.isSaved();
+            }
```
antares1470, line 972: does this check still need to be in place with the updates you made? Looking good in testing

Quasar985: The file being overwritten shouldn't happen anymore, but there might be other cases where a graph is "invalid". Up to you if you think it should go.

antares1470: There is one case I've discovered where the code does come into effect (which I will put into another ticket) but it only kicks in if you make changes afterwards. Can probably leave in for now but we should probably make a note that when all cases of outside manipulation are covered, we probably won't have any reason to retain the code

Issue #2055, opened 2024-05-28
DataAccessView Option Category Changes


capricornunicorn123:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When adjusting the visible categories in the Data Access View using the options panel, changes to the visible categories are not reflected in the Data Access View until an application restart occurs. Adjusting the option does not however prompt the user to restart the application. 

On restarting the application, if there are no categories in the visible column, an error is thrown.

### Behaviour
**Expected behaviour:** update instantly with no errors
**Actual behaviour:** errors thrown or no update

HolandCrick:
@Quasar985 thanks for fixing the DataAccessView Option Catagory issue. Looks like it is working really well. No more errors and very responsive.
- Did not Prompt the user to restart the application.
- On restarting the application, if there are no categories in the visible column, an error is thrown.

![image](https://github.com/constellation-app/constellation/assets/157770877/3b5103ce-0664-4175-be80-f9c205adc699)

PR #2065, opened 2024-06-04
Bug/data access view option category changes

Quasar985:
### Prerequisites

### Description of the Change

Added code to fix exception caused by removing all data access view plugins and added code to automatically refresh visible plugins in data access view.

### Alternate Designs

Could prompt the user to restart consty, similar to changing look and feel. The chosen solution doesn't require a restart, though.

N/A

### Benefits

Fixes exception and saves user time as they no longer need to restart consty to see data access view changes.

### Possible Drawbacks

None.

### Verification Process

Manual and automated testing.

### Applicable Issues

https://github.com/constellation-app/constellation/issues/2055


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -410,57 +436,58 @@ public DataAccessPane getDataAccessPane() {
...
      * @return the data access plugins
-     * @see DataAccessPaneState#getPlugins() 
+     * @see DataAccessPaneState#getPlugins()
      */
     public Map<String, Pair<Integer, List<DataAccessPlugin>>> getPlugins() {
         return plugins;
     }
-    
+
+    public void setPlugins(Map<String, Pair<Integer, List<DataAccessPlugin>>> newPlugins) {
```
Delphinus8821, line 445: final

Delphinus8821 (changes requested):
Looks like it is working really well. No more errors and very responsive.

The warning on the options page to tell the users to restart Constellation for changes to take effect can be removed now.

Delphinus8821 (approved):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/tasks/LookupPluginsTask.java
```diff
@@ -82,7 +83,11 @@ public Map<String, Pair<Integer, List<DataAccessPlugin>>> get() {
                 orderedPlugins.put(category, new Pair<>(position, allPlugins.get(category)));
             });
         }
-
         return orderedPlugins;
     }
+
+    // Only used in testing
+    public static Preferences getPreferences() {
```
antares1470, line 90: can this be made protected if only being used for testing?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/components/DataAccessTabPane.java
```diff
@@ -77,41 +81,69 @@ public class DataAccessTabPane {
...
         // Update the button when the user adds/removes tabs.
         tabPane.getTabs().addListener(
                 (final ListChangeListener.Change<? extends Tab> change) -> this.dataAccessPane.update()
         );
+
+        NbPreferences.forModule(DataAccessViewPreferenceKeys.class).addPreferenceChangeListener(this);
+    }
+
+    @Override
+    public void preferenceChange(PreferenceChangeEvent evt) {
```
antares1470, line 117: parameter final

antares1470 (changes requested):

capricornunicorn123:
Looks we have a functionality issue.

Selecting the name of a plugin, the name of a category or a collapse arrow should be able to collapse or expand the plugin or the category. In both of these cases, the plugins that are selected should persist and remain selected as indicated by the tick and the green box. However doing these actions appears to reset all plugins to unselected. Also note some formatting anomalies in the alignment of the `Select All` plugin when the data access view is refreshing.

DAV being weird

Also, might be worth noting here as this is what I was trying to test. Are we happy with the functionality that when a plugin category containing selected plugins is removed and re added to the data access view, then these re-introduced, previously-selected plugins are deselected by default? I would assume this functionality is intuitive and easiest to implement however I'm unsure if users would appreciate the persistence of plugin selection through the hiding and unhiding process. 

i.e. `Select Top N` is selected > `Utility` is hidden > `Utility` is unhidden > the selection of `Select Top N` has not persisted through the hiding process and is now unselected.

antares1470:
> Also, might be worth noting here as this is what I was trying to test. Are we happy with the functionality that when a plugin category containing selected plugins is removed and re added to the data access view, then these re-introduced, previously-selected plugins are deselected by default? I would assume this functionality is intuitive and easiest to implement however I'm unsure if users would appreciate the persistence of plugin selection through the hiding and unhiding process.
> 
> i.e. `Select Top N` is selected > `Utility` is hidden > `Utility` is unhidden > the selection of `Select Top N` has not persisted through the hiding process and is now unselected.

@capricornunicorn123 I think plugins being unselected after being hidden then unhidden is ok. The main reason I would suggest is the main use of the functionality. I would guess that normally if someone is moving a category to hidden, its because they don't have an interest in the plugins of that category and in that case, whether any of those plugins are selected or not is probably not important. If they do want to see it again, I suspect that will be dictated by use case, but probably not immediately after hiding.

Similarly for general use, I imagine people using Data Access View will adjust categories either before or after using the View, but probably not in the middle of use. 

The only case then (testing aside) where it could pop up would be following the steps you outlined with the caveat that hiding the category was accidental. But that seems very niche and highly unlikely. In fact in the general case, users may actually prefer plugins come back unselected because it may result in plugins being accidentally run if they didn't realise they were already selected (and weren't intending to run)

capricornunicorn123:
> I think plugins being unselected after being hidden then unhidden is ok.

Brilliant, thanks for clarifying

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/panes/QueryPhasePane.java
```diff
@@ -98,23 +101,37 @@ public final void updatePlugins(final Map<String, Pair<Integer, List<DataAccessP
...
 
+            final List<String> listOfPluginNames = new ArrayList<>();
             for (final Pair<Integer, HeadingPane> plugin : orderedPlugins) {
+                listOfPluginNames.add(plugin.getValue().getText());
                 dataSourceList.getChildren().add(plugin.getValue());
                 dataSources.addAll(plugin.getValue().getDataSources());
             }
 
+            // Iterate over map and remove hidden plugins
+            for (var entry : new HashMap<>(currentPlugins).entrySet()) {
```
antares1470, line 128: `Map.Entry<String, HeadingPane>` instead of `var` (and final)

antares1470 (changes requested):

antares1470 (approved):

Issue #2054, opened 2024-05-22
Schema View throwing errors


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When trying to open Schema View, it freezes for a moment and then periodically throws a bunch of exceptions (see comments for some of them).

### Steps to Reproduce

1. Open new graph
2. Open Schema View
3. Wait

**Expected behaviour:** Schema View loads and runs without issues

**Actual behaviour:** Schema View doesn't load properly and throws a bunch of errors

**Reproduces how often:** \[What percentage of the time does it reproduce?\]

### Additional Information

See comments for sample of exceptions thrown. It seems they are all related but possibly come from the use of the use of the indriya library

antares1470:
```
java.lang.ExceptionInInitializerError: Exception java.lang.IncompatibleClassChangeError: Method 'tech.units.indriya.function.MultiplyConverter tech.units.indriya.function.MultiplyConverter.ofPiExponent(int)' must be Methodref constant [in thread "JavaFX Application Thread"]
	at si.uom.NonSI.<clinit>(NonSI.java:129)
	at org.geotools.referencing.cs.DefaultCoordinateSystemAxis.<clinit>(DefaultCoordinateSystemAxis.java:107)
	at org.geotools.referencing.cs.AbstractCS.<clinit>(AbstractCS.java:79)
	at org.geotools.referencing.crs.DefaultEngineeringCRS.<clinit>(DefaultEngineeringCRS.java:78)
	at org.geotools.referencing.factory.epsg.CartesianAuthorityFactory.<clinit>(CartesianAuthorityFactory.java:54)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1160)
[...]

PR #2069, opened 2024-06-07
Fix Indriya dependency conflict

capricornunicorn123:
### Prerequisites

### Description of the Change

This change aims to fix the issues that occur when having both Indriya 1.3 and Indriya 2.0.2 in the core dependencies module. 

The inclusion of the tag `<dependency  org="org.languagetool" name="language-en" rev="6.0" conf="languagetoolconf->default" />` drags in a number of jar files which are previous versions of existing Jars in core dependencies. To resolve the issues that these duplications cause, the required dependencies for `language-en` have been statically included in a new Core Spell Check Module. All spell-checking code has been moved from Utilities to this module. 

### Alternate Designs

Finding a way for the project to update these Core Spell Check dependencies dynamically would be ideal however I am unsure of how to go about it. I'm not sure if we can have another ivy.xml file in the Core Spell Check module or how we get the build script to use it to update those specific dependencies.

If and when languagetools updates to reference more recent packages, spell check and its dependencies can likely be reintegrated into Core Dependencies and Core Utilities. 

Is a core feature. Separating it into a separate project to be layered ontop of core would resolve this dependency issue however would be a bold move for the sake of one feature. 

### Benefits

It works now ! - yay

### Possible Drawbacks

It is a pretty stinky solution 💩

### Verification Process

Open the Schema view and note no errors
Open the Data Access View and enter misspelt text in the extract words from text field and note the spell check works. 

### Applicable Issues

#2054


antares1470 (commented):
I really don't feel comfortable with this solution. While it might work, as you point out its pretty messy and doesn't look at all maintainable which I think is an important consideration.

For what its worth, I had run a test locally where I removed the conf from the language-en dependency and spellcheck still worked fine (schema view still had troubles of course) which to me indicated that we don't need any duplicates of libraries we already have besides indriya (i.e. it worked with any newer versions of libraries that we already had). I also note the jars copied into the spellcheck module just appears to be the entire ivy cache which spellcheck shouldn't need the majority of (should still be able to make use Core Dependencies).

Ultimately, I'd prefer to look at getting it all working with dependencies still stored in Core Dependencies first, and only look at storing dependencies elsewhere in Core if there are no other options (this solution to me feels like a last resort)

Issue #2050, opened 2024-05-20
Add Spelling Check in Notes View, currently only added in "Data Access View"


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Add Spelling Check in Notes View, currently only added in "Data Access View" But the feature should add in any other view or wherever users might find it useful.

### Steps to Reproduce

1.   Open Consty
2. Open any graph
3. Open Notes View
4.  Type free text in different text box. see the spell check is working or not 

**Expected behaviour:** Spelling Check should be added in Notes View or wherever users might find it useful.

**Actual behaviour:** Spelling Check added only works in Data Access View Text Fields.

**Reproduces how often: ** 100% times

### Additional Information
Spelling Check should be added Notes View.

HolandCrick:
Spellcheck in Note View is not working. Whereas, it works nicely in DAV. This ticket #2025 is for Spellcheck in Note View.

HolandCrick:
@spica198 - Thanks for adding spellcheck in NoteView. Now Notes View and , "Data Access View both area spellcheck working.

PR #1910, opened 2023-10-22
Spell checking in Constellation

Auriga2:
### Prerequisites

- [ ] Reviewed the [checklist](CHECKLIST.md)

- [ ] Reviewed feedback from the "Sonar Cloud" bot. Note that you have to wait
    for the "CI / Unit Tests") to complete first. Failed Unit tests can be 
    debugged by adding the label "verbose logging" to the GitHub PR.

### Description of the Change
Spell checking in now available in Constellation Data access view and Database importer. It can be added in other areas of Constellation as required.

In the right click context menu, `Turn On Spell Checking` check menu item can be used to toggle on/off spell checking of that field. It is set to `On` by default.

Incorrect spellings and some grammar errors are highlighted and underlined as the user types, except when any alpha numeric character or apostrophe is typed at the end. Left click on a highlighted word will pop up the suggestions list to choose from.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/870

Issue #870, opened 2020-10-22
Spellcheck in Constellation


antares1470:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

For text fields in Data Access View plugins and elsewhere, spell check underlining misspelled words would help prevent typos in what is being submitted.

Auriga2:
Here is the summary of a few options we can use. I am inclined to use `Languagetool` as it has active updates.
[Add spelling check.docx](https://github.com/constellation-app/constellation/files/8337532/Add.spelling.check.docx)

Auriga2:
Based on below comparison we chose JLanguagetool as it has ongoing releases, good documentation and support.
![image](https://user-images.githubusercontent.com/64826928/162897410-f0064512-abdc-42ab-8553-57795527469b.png)

But currently having issues getting it to work when the library is downloaded with ivy. It works in a new Maven project. 

**Performance** 
Regarding the slowness, it was only observed when debugging. When running it was relatively faster. When dealing with multiple threads,

> Create one instance of JLanguageTool per thread, but create the language only once (e.g. new BritishEnglish()) and use that for all instances of JLanguageTool

Also consider [suggestions here](https://forum.languagetool.org/t/why-is-jlanguagetool-not-thread-safe/2525/4): 

> You’ll see that only the very first initializing is slow. After that, it will be fast, no matter from how many threads it’s called. That’s what we do on languagetool.org and it works fine to serve many hundred thousand requests per day from a single machine.

For better performance we can also use [MultiThreadedJLanguageTool](https://languagetool.org/development/api/org/languagetool/MultiThreadedJLanguageTool.html) 


**Ignore Words**
`Ignore` should just leave the current text the user has typed while an `Add to Dictionary` option can be added to permanently ignore it, and consider it as a valid word. In JLanguagetool [`addIgnoreTokens` or `acceptPhrases`](https://dev.languagetool.org/java-api#spell-checking) can be used for that. 
[...]

PR #2080, opened 2024-06-17
removed spellcheck while its causing issues

antares1470:
### Prerequisites

### Description of the Change

A few issues have been found in the application recently and they seem to center on the addition of the extra indriya library version which is resulting in some conflicts with other parts of the application trying to use the newer version (but was getting stuck using the older version).

Rather than try address this with this all in master, this PR pulls the spellcheck feature out of master (which was the feature requiring the older indriya version, see #1910) with the intention of creating a new PR after this has been merged in order to look at the issues in an isolated PR (and not affecting the performance of master)

### Alternate Designs

Could try resolve these while retaining spellcheck in master but it seems the solution may not be so simple to resolve neatly. This change is recognising that the feature may not have been as ready as originally thought

Pulls a feature out that was causing a number of runtime issues and exceptions

### Benefits

Bugs caused by the features introduction should be gone

### Possible Drawbacks

This sadly means spellcheck is gone for the meantime, but the intention is to hopefully to able to re-introduce it once these issues have been resolved

### Applicable Issues

#870 
#2054 
(Will also affect #2050)


Delphinus8821 (approved):

capricornunicorn123 (approved):
Looks good to me. The only change that I can see has not been converted back to a pre Spellcheck state is the integration of text inputs and password inputs into a single ValueInputPane. I'd argue that keeping the PasswordInputPane separate is a good move.

Auriga2 (approved):

Issue #2058, opened 2024-05-31
Timeline View error ... ConcurrentModificationException


OrionsGuardian:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Testing in V3 dev build:
Create a simple graph with a few nodes in sequence and link each node to the next node with a single transaction, so you have, for example, 5 nodes and 4 transactions.
Set the datetime field on the transactions in sequence so that they occur one hour after the previous transaction.

Open the Timeline View.
It should show the 4 transactions with 1 hour spacing between them.

Try to select an area in the Timeline View which encompasses 2 or more of the transactions.
Exception occurs:

![image](https://github.com/constellation-app/constellation/assets/104604488/a08cd00e-9c9f-40f7-8f23-806244237aa8)

**Expected behaviour:**

The entries selected in the Timeline View will be selected in the Graph.

**Actual behaviour:**

ConcurrentModificationException

**Reproduces how often:**

100%

### Additional Information

Following the same steps in V2.11 works correctly.

HolandCrick:
@OrionsGuardian - thanks for your time to fix the Timeline View Error. It's working nicely now.

![image](https://github.com/constellation-app/constellation/assets/157770877/6d0cdb0b-44b3-4e41-bff4-4c3bd2608cb3)

PR #2097, opened 2024-06-27
Fix issue in Timeline View causing ConcurrentModificationException

OrionsGuardian:
### Prerequisites

### Description of the Change

Changed the code in the Timeline View which iterates and selects entries in the graph, 
and also changed the condition controlling when that code is run.

Restore pre-existing functionality in the Timeline View.

### Benefits

Allow selecting entries in the Timeline View to be reflected in the graph.

### Verification Process

Create a simple graph with a few nodes in sequence and link each node to the next node with a single transaction, so you have, for example, 5 nodes and 4 transactions.
Set the datetime field on the transactions in sequence so that they occur one hour after the previous transaction.
Open the Timeline View.
It should show the 4 transactions with 1 hour spacing between them.
Try to select an area in the Timeline View which encompasses 2 or more of the transactions.
Confirm that the graph entities are correctly selected, and no Exceptions occurred.

### Applicable Issues

#2058


CoreTimelineView/src/au/gov/asd/tac/constellation/views/timeline/TimelineSelectionPlugin.java
```diff
@@ -50,16 +54,32 @@ public TimelineSelectionPlugin(final Set<Integer> vertices, final Set<Integer> t
     @Override
     public void edit(final GraphWriteMethods graph, final PluginInteraction interaction,
             final PluginParameters parameters) throws InterruptedException {
+        final boolean vertexAttributeUndefined = Graph.NOT_FOUND == graph.getAttribute(GraphElementType.VERTEX,
+                VisualConcept.VertexAttribute.SELECTED.getName());
+        final boolean transactionAttributeUndefined = Graph.NOT_FOUND == graph.getAttribute(GraphElementType.TRANSACTION,
+                VisualConcept.TransactionAttribute.SELECTED.getName());
         final int selectedVertexAttrID = VisualConcept.VertexAttribute.SELECTED.ensure(graph);
         final int selectedTransactionAttrID = VisualConcept.TransactionAttribute.SELECTED.ensure(graph);
```
antares1470, line 62: I think I need some help understanding this one. It seems a bit odd to me that we're adding a boolean variable for whether the Selected attributes exist and then creating those attributes immediately after

OrionsGuardian: This was done to replicate the functionality it was using previously, but I suspect with the old code it was never actually entering that code segment otherwise we would have been getting the ConcurrentModificationException more often.
I was considering fully removing the code block which removes vertices and transaction, as I was not able to replicate a situation that would trigger that code.

antares1470 (approved):

Delphinus8821 (approved):

Issue #2107, opened 2024-07-02
java.lang.NullPointerException - Error when delete any Template


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

java.lang.NullPointerException - Error when delete any template. 

### Steps to Reproduce

1.  File Menu -> Select Manage Template

2.  Select any saved template from the dropdown box

3.  Click Delete

**Expected behaviour:** Delete Should perform with a confirmation warning. 

**Actual behaviour:** \[java.lang.NullPointerException - Exception Error appears
![image](https://github.com/constellation-app/constellation/assets/157770877/615c6ea3-d3cd-4b8f-851d-26b9532c0c1e)

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

Auriga2:
Present in v2.11 too.

Auriga2:
The exception is thrown in v2.11 but the template gets deleted. In v3 they don't get deleted after the exception, so has to be fixed.

HolandCrick:
@Auriga2 thanks for fixing the bug when delete any Template. It's working nicely now.

PR #2156, opened 2024-07-24
Fix the issue when deleting templates

Auriga2:
### Prerequisites

### Description of the Change
Templates can now be deleted (although the null pointer exception is thrown, which will be handled in a separate PR)

The cause was using `File.pathSeparator ( ; )` in file path, which doesn't work with java.io.FILE 
E.g. `C:\Users\user1\.CONSTELLATION\Graph Templates\au.gov.asd.tac.constellation.graph.schema.AnalyticSchemaFactory;New Template 1` doesn't exist to delete. Reverted back to use `/` instead.

No way to delete templates otherwise.

### Verification Process

1. File Menu -> Select Manage Template
2. Select a saved template from the dropdown list
3. Click Delete button. 
4. Click Ok Button 
5. Open File -> Select Manage Template again and verify the above template is no  longer available in the dropdown list

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2107

### DIFF
--- a/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/create/NewSchemaGraphAction.java
+++ b/CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/create/NewSchemaGraphAction.java
@@ -68,7 +68,6 @@ public class NewSchemaGraphAction extends AbstractAction implements DynamicMenuC
     private static Map<String, String> templates;
     private static JMenu menu;
     private static final Map<String, Icon> ICON_CACHE = new HashMap<>();
-    public static final String FORWARD_SLASH = "/";
 
     public static File getTemplateDirectory() {
         return templateDirectory;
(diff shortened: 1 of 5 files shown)


CoreGraphNode/src/au/gov/asd/tac/constellation/graph/node/create/NewSchemaGraphAction.java
```diff
@@ -68,6 +68,7 @@ public class NewSchemaGraphAction extends AbstractAction implements DynamicMenuC
     private static Map<String, String> templates;
     private static JMenu menu;
     private static final Map<String, Icon> ICON_CACHE = new HashMap<>();
+    public static final String FORWARD_SLASH = "/";
```
antares1470, line 71: I'm not sure it makes sense to have this constant sitting in this class. Maybe it would fit better in `SeparatorConstants`?

Auriga2:
Fixed the exception also in this PR, as it ended up being a few lined of code change.

Delphinus8821 (approved):

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/text/SeparatorConstants.java
```diff
@@ -36,4 +36,5 @@ public class SeparatorConstants {
     public static final String ASTERISK = "*";
     public static final String BLANKSPACE = " ";
     public static final String QUOTE = "\"";
+    public static final String FORWARD_SLASH = "/";
```
OrionsGuardian, line 39: It seems like this constant is unnecessary.
The original code was using `File.pathSeparator`, which was wrong because we're not defining separate paths.
It should work just fine with `File.separator` in each of the places referring to `FORWARD_SLASH`

OrionsGuardian (changes requested):

OrionsGuardian (approved):

antares1470 (approved):

Issue #2139, opened 2024-07-15
Test Notifications OK button throws exception


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

An Unexpected exception is thrown on Test Notifications when OK clicked:
![image](https://github.com/user-attachments/assets/787a6c5a-b25d-45e8-8a3e-54313f437507)

### Steps to Reproduce

1.  Create a new Graph.

2.  Select menu item Experimental -> Developer -> Test Notifications

3.  Click OK. An Unexpected Exception is thrown.

**Expected behaviour:** Expected a series of notification dialogs. This can be seen if you click Cancel instead of OK.

**Actual behaviour:** 

- An exception is thrown. 
- Also, in Display Large Alert dialogs, the "Show Details" and "Hide Details" is difficult to see in FlatLaf Dark Mode. Probably related to https://github.com/constellation-app/constellation/issues/2137.
- If app is not on primary monitor, the some of the dialogs will show up on a different screen.

**Reproduces how often:** 100%

### Additional Information

Happening on v.3.0.0 - rc1.  No exception thrown in v02.11.0.

HolandCrick:
problem is not only with  ok button but also in  cancel and X all three giving error.

andromeda-224:
The notification dialogs (plain, warning, info, error) displayed are expected. The exception on the Ok button click is not expected and should be fixed.

andromeda-224:
The issue with "Show Details" and "Hide Details" being difficult to see in FlatLaf Dark Mode. is not related to https://github.com/constellation-app/constellation/issues/2137. This is fixed with stylesheet. New ticket created for Display alert dialogs appearing on different screen #2159.

HolandCrick:
@andromeda-224 Thanks for addressing the bug. It's generating 13 Notifications without exception error.

![image](https://github.com/user-attachments/assets/6f7adf14-42c4-489d-868c-6565877bed58)

PR #2161, opened 2024-07-25
Fix Test Notifications Ok button exception

andromeda-224:
### Prerequisites

### Description of the Change
Exception fixed by returning set of non-null PluginParameters in CreateParameters().
Dark themed Show/Hide Details text fixed with stylesheet.

### Alternate Designs
None.

To fix the unexpected exception on the Ok button for Test Notifications.

### Benefits
To fix the unexpected exception on the Ok button for Test Notifications.

### Possible Drawbacks
None.

### Verification Process

1. Create a new Graph.
2. Select menu item Experimental -> Developer -> Test Notifications. 
3. Click OK.

Notification dialogs pop up and no exception is displayed. All text in the dialog boxes are readable.

### Applicable Issues
#2139


antares1470 (commented):
I've had a look at the issue, and I wonder there may be a better way to address the problem.

Adding a plugin parameters object does solve the issue as but upon closer inspection, I actually question why there is a need to display a PluginParameterSwingDialog at all given there are no parameters (generally, it wouldn't make any sense to display one unless there were parameters to tinker with before executing). It seems like its adding an unnecessary extra layer. 

Acknowledging this is just a developer testing action but I wonder whether it would be better to simply remove the need for the dialog and instead just execute the plugin from the get go? This would also remove the need to have an empty parameters object.

Delphinus8821 (approved):
Agree with @antares1470 about removing the empty parameter object, otherwise looking good

Quasar985 (approved):

Delphinus8821 (approved):

CoreTesting/src/au/gov/asd/tac/constellation/testing/TestNotificationsAction.java
```diff
@@ -64,14 +64,8 @@ public TestNotificationsAction(final GraphNode context) {
     public void actionPerformed(final ActionEvent e) {
         // plugin notifications
         final Plugin plugin = new TestNotificationsPlugin();
-        final PluginParameters pp = plugin.createParameters();
-
-        final PluginParametersSwingDialog dialog = new PluginParametersSwingDialog(Bundle.CTL_TestNotificationsAction(), pp);
-        dialog.showAndWait();
-        if (PluginParametersDialog.OK.equals(dialog.getResult())) {
-            PluginExecution.withPlugin(plugin).withParameters(pp).executeLater(context.getGraph());
-        }
```
OrionsGuardian, line 73: With the removal of this code, there are now some unused imports that should also be removed.

OrionsGuardian (changes requested):

CoreTesting/src/au/gov/asd/tac/constellation/testing/TestNotificationsAction.java
andromeda-224: Done.

antares1470 (approved):

Issue #2149, opened 2024-07-19
Conversation View showing results from a different graph


OrionsGuardian:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

It appears that the Conversation data that is used in one graph will be re-used in the next graph that is activated when that graph does not contain conversation data of its own.

![image](https://github.com/user-attachments/assets/07912e9f-8a2d-492f-8b99-f212a7f2ea51)

The Conversation data should be cleared when a new graph becomes active.

Auriga2:
Not present in V 2.11, so need fixed in v3

HolandCrick:
@Auriga2 thanks a lot for fixing the bug. It's working now; does not need to select a different node to see any Conversation data again. All good now.
![image](https://github.com/user-attachments/assets/a9b40762-5b95-42a4-8c83-6227530dc2d3)

PR #2169, opened 2024-08-01
Fix the conversation view to not show conversations from other graphs.

Auriga2:
### Prerequisites

### Description of the Change
Conversation view no longer shows conversations from the last viewed graph when the current graph does not contain conversation data of its own.

Bug fix

### Benefits
Bug fix

### Verification Process

1. Open a graph with conversation data
2. Create a new graph and add 2 nodes and a transaction, which would not have conversation data to display
3. Open Conversation View and verify conversations are shown when the first graph is selected
4. Verify conversations clear when the nodes are selected on the second graph

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2149

### DIFF
--- a/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/Conversation.java
+++ b/CoreConversationView/src/au/gov/asd/tac/constellation/views/conversationview/Conversation.java
@@ -507,9 +507,9 @@ public class Conversation {
                 final Thread thread = new Thread(CONVERSATION_VIEW_UPDATE_SENDER_THREAD_NAME) {
                     @Override
                     public void run() {
+                        senderMessages.clear();
                         if (!temporalMessages.isEmpty()) {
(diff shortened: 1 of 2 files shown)


Delphinus8821 (approved):

OrionsGuardian (changes requested):
The code change does work to address the described issue, but I found another related issue that should also be addressed at the same time.

Test the case where you have one graph (A) with Conversation data and another graph (B) without Conversation data.
Select a node in graph A, see the Conversations displayed in the view.
Switch to graph B.
Deselect all nodes there.
Switch back to graph A.
The currently selected node in graph A is now not showing its Conversation data.
You need to select a different node to see any Conversation data again.

antares1470 (approved):

Delphinus8821 (approved):

antares1470 (approved):

Issue #2172, opened 2024-08-02
Welcome page allows multiple copies of the same graph to open, throwing exceptions


Auriga2:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Recent graphs in the Welcome page can be clicked multiple times, which opens the same graph in multiple tabs. 

![image](https://github.com/user-attachments/assets/b816d602-9452-42fb-a2ce-3cb0fa1ed442)

If you do this on a larger graph, constellation can run out of memory and stuck. 

![image](https://github.com/user-attachments/assets/ca8da986-13c7-488c-8aed-dd3b331618db)

Clicking on these graphs/tabs while partially loading throws this exception.

![image](https://github.com/user-attachments/assets/9a76ae04-25e7-427b-8542-9496140ea65a)

### Steps to Reproduce

1.  Open constellation, open a larger graph and close the app which will add it in the `Recent` section in the Welcome page.

2.  Open constellation and double click on the graph above in the Welcome page

3.  Observe it creating 2 graphs. 
4. Clicking on the graph above in the Welcome page multiple times can cause the 1st exception 
5. Clicking on the graph above in the Welcome page a few times and clicking on these graphs/tabs that are partially loading can cause the 1st exception 

**Expected behaviour:** Welcome page not allowing to open multiples of the same graph and no exceptions thrown

**Actual behaviour:** Welcome page allows to open multiples of the same graph and exceptions are thrown

**Reproduces how often:** 90% with a larger graph

### Additional Information
Observed in v3.0.0.-rc1
Same behavior in v2.11

HolandCrick:
@Quasar985 thanks for fixing the - Welcome page allows multiple copies of the same graph to open, throwing exceptions. It not throwing exception anymore. Working as expected.

PR #2183, opened 2024-08-13
Stop the Welcome View from Opening the Same Graph Multiple Times

Delphinus8821:
### Prerequisites

### Description of the Change
The Welcome View currently allows you to open the graphs stored in the Recent Graphs as many times as you want and this can sometimes cause exceptions.
This change stops users from clicking the graph in Recent Graphs multiple times and having the graph open multiple times, instead it will swap to the tab with the already open graph.

This change still allows for multiple graphs of the same name to be opened as they would have to be saved in different locations. 

Bugfix to stop an exception from occurring in Core.

### Verification Process
1. Have a graph saved in the Recent graphs on the Welcome View
2. Open the graph
3. Attempt to open the graph again, Constellation should swap tabs to the already open graph
4. Attempt to open a graph with the same name which has been saved in a different location. It should still open.

### Applicable Issues
#2172


CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
```diff
@@ -230,12 +237,24 @@ private FlowPane recentGraphsSetup() {
...
+                //If the graph is already open then it will swap to the open graph instead of opening it again
                 final String path = fileDetails.get(i).getPath();
                 recentGraphButtons[i].setOnAction(e -> {
-                    OpenFile.open(RecentFiles.convertPath2File(path), -1);
-                    saveCurrentDirectory(path);
+                    final Set<TopComponent> topComponents = WindowManager.getDefault().getRegistry().getOpened();
+                    if (topComponents != null) {
+                        fileOpen = false;
+                        topComponents.forEach(component -> {
+                            if ((component instanceof VisualGraphTopComponent) && ((VisualGraphTopComponent) component).getGraphNode().getDisplayName().equals(fileName)) {
```
antares1470, line 248: you can declare a variable with the instanceof so you don't have to cast `component` afterwards

antares1470 (approved):

OrionsGuardian (changes requested):
Can still get multiple copies of the same graph opened.
When you first start Constellation, the first graph you try to open takes some time before it opens.
During that time if you keep clicking on the Recent Files entry on the Welcome Page it will queue up more instances to open the same file.

Delphinus8821:
> Can still get multiple copies of the same graph opened. When you first start Constellation, the first graph you try to open takes some time before it opens. During that time if you keep clicking on the Recent Files entry on the Welcome Page it will queue up more instances to open the same file.

I think the only way to stop this from occurring would be to only allow one graph to be opened from the Recent Graphs at a time. So you wouldn't be able to open multiple different graphs at the same time which I think could be more annoying

Quasar985 (approved):

CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
```diff
@@ -230,12 +237,24 @@ private FlowPane recentGraphsSetup() {
...
-                    OpenFile.open(RecentFiles.convertPath2File(path), -1);
-                    saveCurrentDirectory(path);
+                    final Set<TopComponent> topComponents = WindowManager.getDefault().getRegistry().getOpened();
+                    if (topComponents != null) {
+                        fileOpen = false;
+                        topComponents.forEach(component -> {
+                            if ((component instanceof VisualGraphTopComponent vgtc) && (vgtc.getGraphNode().getDisplayName().equals(fileName))) {
+                                fileOpen = true;
+                            }
+                        });
```
OrionsGuardian, line 251: This code doesn't seem to actually match with the open graph ... it appears that the ".star" suffix is removed from the graph name.
So, it goes into the next block of code, tries to open the same file again, finds it is already open and activates the tab with that graph. So that functionality is fine as is.

OrionsGuardian (changes requested):

Issue #2205, opened 2024-09-10
Password Obfuscator/Deobfuscator throwing error


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

In recent times, a change was made in `PasswordUtilities` to change the algorithm spec from AES/CBC/PKCS5Padding to AES/GCM/NoPadding. This has inadvertently broken the obfuscate algorithm of `PasswordObfuscator` and deobfuscate algorithm of `PasswordDeobfuscator`. It throws the following exception when calling the init function of the Cipher object:
```
java.security.InvalidAlgorithmParameterException: AlgorithmParameterSpec not of GCMParameterSpec
```
It turns that when we changed the algorithm spec to what it is currently, we should have also changed the obfuscate and deobfuscate functions to create a `GCMParameterSpec` object instead of an `IvParameterSpec` (as it is currently doing)

### Steps to Reproduce

Note this will only be possible to fix in the local instance of Constellation (run in Netbeans)

1. Create an action which you can easily access that calls one of `PasswordObfuscator.obfuscate` or `PasswordDeobfuscator.deobfuscate`
2. Compile and launch Constellation
3. Run this action

**Expected behaviour:** No exception thrown

**Actual behaviour:** Exception thrown

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@Quasar985 Thanks for fixing this bug. Looks like it's working now.

PR #2208, opened 2024-09-19
Replace password obfuscator and deobfuscator

Quasar985:
### Prerequisites

### Description of the Change

Fix exception being thrown because of the wrong type of obfuscator and deobfuscator being used.

The constant `T_LEN` has also been added in `PasswordUtilities` as it's required for the new `GCMParameterSpec` obfuscator/deobfuscator. `T_LEN` can only have the values {128, 120, 112, 104, 96}, or {64, 32} for certain applications, so I have chosen 128 (info from: https://docs.oracle.com/javase/8/docs/api/javax/crypto/spec/GCMParameterSpec.html).

Feel free to request any changes.

### Alternate Designs

N/A

N/A

### Benefits

Exception no longer thrown

### Possible Drawbacks

May mess with password related systems. There didn't seem to be any unit testing for the security module until now

### Applicable Issues

https://github.com/constellation-app/constellation/issues/2205

### DIFF
--- a/CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
+++ b/CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
@@ -24,7 +24,6 @@ import static org.mockito.Mockito.when;
 import org.openide.util.Lookup;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
-import static org.testng.Assert.fail;
 import org.testng.annotations.Test;
 
 /**
@@ -103,16 +102,14 @@ public class PasswordUtilitiesNGTest {
 
     /**
      * Test of generateKey method, of class PasswordUtilities.
+     * @throws java.security.NoSuchAlgorithmException
      */
     @Test
-    public void testGenerateKey() {
+    public void testGenerateKey() throws NoSuchAlgorithmException {
         System.out.println("generateKey");
-        try {
-            final byte[] result = PasswordUtilities.generateKey();
-            assertNotNull(result);
-        } catch (NoSuchAlgorithmException e) {
-            fail("NoSuchAlgorithmException exception");
-        }
+
+        final byte[] result = PasswordUtilities.generateKey();
+        assertNotNull(result);
 
     }
 }
--- a/CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
+++ b/CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
@@ -39,6 +39,21 @@ public class PasswordUtilitiesNGTest {
     @Test
     public void testGetIV() {
         System.out.println("getIV");
+        PasswordUtilities.reset();
+
+        final byte[] firstResult = PasswordUtilities.getIV();
+        assertNotNull(firstResult);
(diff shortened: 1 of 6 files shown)


Delphinus8821 (approved):

CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
```diff
@@ -0,0 +1,124 @@
...
+        final Lookup mockLookup = mock(Lookup.class);
+        when(mockLookup.lookup(PasswordSecret.class)).thenReturn(null);
+        try (final MockedStatic<Lookup> mockLookupStatic = Mockito.mockStatic(Lookup.class)) {
+            // Setup static mock
+            mockLookupStatic.when(Lookup::getDefault).thenReturn(mockLookup);
+
+            PasswordUtilities.getIV();
+            // If exception was not thrown, fail
+            fail("NoSuchElementException should have been thrown");
+        } catch (NoSuchElementException e) {
```
antares1470, line 68: instead of doing this, you can add the `expectedExceptions` parameter to the Test annotation, which is a cleaner way of doing the same thing:
`@Test(expectedExceptions = NoSuchElementException.class)`

CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
```diff
@@ -0,0 +1,124 @@
...
+        final Lookup mockLookup = mock(Lookup.class);
+        when(mockLookup.lookup(PasswordSecret.class)).thenReturn(null);
+        try (final MockedStatic<Lookup> mockLookupStatic = Mockito.mockStatic(Lookup.class)) {
+            // Setup static mock
+            mockLookupStatic.when(Lookup::getDefault).thenReturn(mockLookup);
+
+            PasswordUtilities.getKey();
+            // If exception was not thrown, fail
+            fail("NoSuchElementException should have been thrown");
+        } catch (NoSuchElementException e) {
```
antares1470, line 105: same here

CoreSecurity/test/unit/src/au/gov/asd/tac/constellation/security/password/PasswordUtilitiesNGTest.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    @Test
+    public void testGenerateKey() {
+        System.out.println("generateKey");
+        try {
+            final byte[] result = PasswordUtilities.generateKey();
+            assertNotNull(result);
+        } catch (NoSuchAlgorithmException e) {
+            fail("NoSuchAlgorithmException exception");
+        }
```
antares1470, line 121: this could be changed to a throw instead

Quasar985: Rather than catching and re-throwing the same exception. I've removed the try catch block and added `throws` to the test function itself. Do you think this would cause any issues?

antares1470: My apologies, when I meant change to a throw, I meant what you ended up doing (i.e. throw the exception rather than catch it) as appose to throwing an exception in the catch block. It shouldn't cause any unintended issues, as the test will just fail if it encounters an exception. The only reason I would use a catch block in a test is if you do something with the test afterwards that is more than just failing

antares1470 (approved):

Issue #2119, opened 2024-07-08
Info Map options dialog look and feel should be consistent with the rest of the application


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Info Map clustering can be run a couple of different ways, one of which brings up a dialog box to allow you to tinker with the parameters before running the algorithm. Looking closer at the dialog though, it seems that the way its been created is slightly different to other plugins, and so the look and feel of the dialog also feels different. The most notable thing is that it is using radio buttons which should actually be check boxes given you can select more than one at a time in each group.

This should be altered to ensure a more consistent look and feel across the application

### Steps to Reproduce

1. Open new graph
2. Tools -> Cluster -> Info Map...

**Expected behaviour:** Resulting dialog should have a look and feel that is consistent with other dialogs across the application

**Actual behaviour:** Dialog is uniquely created and stands out as much

**Reproduces how often:** 100%

### Additional Information

At the minimum, the radio boxes should become checkboxes.

Given Info Map is a plugin though, there's an argument that this should be refactored to be able to use `PluginParametersSwingDialog` rather than having a custom dialog. This though will require some refactoring of the plugin itself to make some of the options in the dialog parameters themselves (currently the only parameter for the plugin is a config which consists of those parameters found in the dialog

HolandCrick:
@Quasar985 Thanks for re-arranging the Info Map Dialog from radio buttons to drop-down.  The Info Map Dialog is stands out now.

PR #2216, opened 2024-10-15
Refactor info map options dialog box

Quasar985:
### Prerequisites

### Description of the Change

Refactor info map dialog box to have styling consistent with the rest of consty. Uses the `PluginParametersSwingDialog` code instead of swing dialog.
Also, the original issue suggests changing the "Connection Type" and "Dynamic Type" parameters be changed to checkboxes as multiple options can be selected at once. From what I understand this wasn't intentional, so I kept these as single choice parameters.

### Alternate Designs

N/A

N/A

### Benefits

Makes the dialog box style more consistent with the rest of consty.

### Possible Drawbacks

None

### Verification Process

1. Open new graph
2. Tools -> Cluster -> Info Map...
3. Observe styling
4. Use info map with different configurations and observe same results as info map from master branch

### Applicable Issues

https://github.com/constellation-app/constellation/issues/2119


Quasar985:
This is what the new dialog looks like now:
infoMapOptionsPanel

OrionsGuardian (changes requested):
Can you re-implement the Help button which appeared on the old Info Map dialog.

Quasar985:
> Can you re-implement the Help button which appeared on the old Info Map dialog.

Done. Looks like this now:
infoMapOptionsPanel

CoreAlgorithmPlugins/test/unit/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/infomap/InfoMapPluginNGTest.java
```diff
@@ -0,0 +1,234 @@
...
+            case CONNECTION_TYPE_EDGES ->
+                Config.ConnectionType.EDGES;
+            case CONNECTION_TYPE_LINKS ->
+                Config.ConnectionType.LINKS;
+            default ->
+                Config.ConnectionType.TRANSACTIONS;
+        };
+    }
+
+    private PluginParameters createParametersHelper(final String connectionType, final String dynamicType) {
```
antares1470, line 195: this code in this function can probably be simplified by calling instance.createParameters() and then just updating the parameter values with the desired values, rather than recreating them all from scratch

OrionsGuardian (changes requested):
This is likely a pre-existing issue ...
There are a large number of LOG entries being created for mid-large graphs.
The method dumpActiveNetwork() in the InfomapGreedy class should be in an `if (DEBUG) {}` block.
There may be other LOG messages that should also be controlled by the DEBUG setting.

Quasar985:
> This is likely a pre-existing issue ... There are a large number of LOG entries being created for mid-large graphs. The method dumpActiveNetwork() in the InfomapGreedy class should be in an `if (DEBUG) {}` block. There may be other LOG messages that should also be controlled by the DEBUG setting.

Done. I've added these if statements in a few places (`InfomapGreedy.java` and `Lcg.java`), so let me know what you think. I found it sped up the info map significantly on a graph with 10,000 nodes and 5000 transactions.

CoreAlgorithmPlugins/test/unit/src/au/gov/asd/tac/constellation/plugins/algorithms/clustering/infomap/InfoMapPluginNGTest.java
```diff
@@ -192,42 +153,25 @@ private Config.ConnectionType connectionTypeHelper(final String connectionType)
...
+        final PluginParameters parameters = infoMapInstance.createParameters();
+        final PluginParameters updatedParameters = new PluginParameters();
 
         // Connection type
         final PluginParameter<SingleChoiceParameterType.SingleChoiceParameterValue> connectionParam = SingleChoiceParameterType.build(CONNECTION_TYPE_PARAMETER_ID);
         connectionParam.setName(CONNECTION_TYPE_PARAMETER_ID_NAME);
         SingleChoiceParameterType.setOptions(connectionParam, CONNECTION_TYPE_PARAM_VALUES);
         SingleChoiceParameterType.setChoice(connectionParam, connectionType);
-        parameters.addParameter(connectionParam);
+        updatedParameters.addParameter(connectionParam);
```
antares1470, line 165: is this different to the parameter for the original plugin? If not, the correct thing to do here would be to simply get the parameter and update the value rather than creating a new instance of it entirely. e.g.
```
connectionParam = parameters.getParameters().get(CONNECTION_TYPE_PARAMETER)
SingleChoiceParameterType.setChoice(connectionParam, connectionType)
```
If the value you're setting is the same as the default, you don't even need to change the value after calling instance.createParameters() (in which case you can probably remove the helper function here)

Quasar985: Done. I originally thought it couldn't be done this way as getParameters returns an unmodifiable map, but I was mistaken about how that works. I've removed the helper function and streamlined that part of testing

antares1470 (approved):

Delphinus8821 (approved):

Issue #2275, opened 2025-03-13
Find View -> Delete Results From Graphs(s) - ConcurrentModificationException


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

There is a ConcurrentModificationException thrown when attempting to "Delete Results From Graph(s)" for multiple results across multiple graphs from the Find View.

### Steps to Reproduce

1. Open multiple graphs. It would be good if they have nodes or transactions with similar attribute identifiers that can be searched on. Eg. You can create multiple Complete Graphs that will have same identifiers.

2. Open the Find View and search for Standard Text "Node", Search Attributes: "Identifier", Search In: "All Open Graphs". Multiple nodes across graphs are selected.

3. Click Delete Results From Graph(s) button and Yes to confirm deletion.

4. A ConcurrentModificationException pops up. 

This happens with transactions too.

**Expected behaviour:** \[What you expect to happen\]

Expect the selected results to be deleted from graphs with no exceptions.

**Actual behaviour:** \[What actually happens\]

ConcurrentModificationException occurs.

**Reproduces how often:** \[What percentage of the time does it reproduce?\] 100%

### Additional Information

![Image](https://github.com/user-attachments/assets/0c3b9970-dfdf-4a62-8e19-7290e509b111)
Occurs in v3, v3.1-rc1 but not in v2.11

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@andromeda-224  - Thanks for resolving the bug. The selected results can be deleted from graphs with no exceptions.

PR #2279, opened 2025-03-19
Fix for FindView ConcurrentModificationException

andromeda-224:
### Prerequisites

### Description of the Change
There is a ConcurrentModificationException thrown when attempting to "Delete Results From Graph(s)" for multiple results across multiple graphs from the Find View.

This was because the deletion of the elements was done while running through the list of found results, which changes the list (hence the exception). The change was to create and store the DeleteResultsPlugin for each result and execute them later.

This change would prevent this exception from happening in Core.

### Benefits

This change would prevent this exception from happening in Core.
### Verification Process
1. Open multiple graphs. It would be good if they have nodes or transactions with similar attribute identifiers that can be searched on. Eg. You can create multiple Complete Graphs that will have same identifiers.

2. Open the Find View and search for Standard Text "Node", Search Attributes: "Identifier", Search In: "All Open Graphs". Multiple nodes across graphs are selected.

3. Click Delete Results From Graph(s) button and Yes to confirm deletion.

There would be no exception and the results would be deleted from the graphs.

### Applicable Issues
#2275

### DIFF
--- a/CoreFindView/src/au/gov/asd/tac/constellation/views/find/FindViewController.java
+++ b/CoreFindView/src/au/gov/asd/tac/constellation/views/find/FindViewController.java
@@ -34,8 +34,11 @@ import au.gov.asd.tac.constellation.views.find.utilities.FindResult;
 import au.gov.asd.tac.constellation.views.find.utilities.FindResultsList;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


CoreFindView/src/au/gov/asd/tac/constellation/views/find/FindViewController.java
```diff
@@ -477,21 +480,39 @@ public void deleteResults(final FindResultsList foundResults, final int graphsSe
...
-                }
-            }
+                    final String graphId = result.getGraphId();
+                    ArrayList deleteResultsPluginsList = graphDeletePluginMap.get(graphId);
+                    if (deleteResultsPluginsList == null) {
+                        deleteResultsPluginsList = new ArrayList<>();
+                    }
+                    // store in map against graph id                    
+                    deleteResultsPluginsList.add(deleteResultsPlugin);
+                    graphDeletePluginMap.put(graphId, deleteResultsPluginsList);
```
OrionsGuardian, line 496: When deleteResultsPluginsList is null, do we still need to put an empty list in the deletion map?

OrionsGuardian (changes requested):

CoreFindView/src/au/gov/asd/tac/constellation/views/find/FindViewController.java
andromeda-224: If the deleteResultsPluginsList is null, it means that the element is on a different graph and it needs to be created in order to put in the current result to delete for that graph. So technically, the list will not be empty and should be populated by the new deleteResultsPlugin for that result for that new graph.  The for loop will iterate through the arraylist, so even if it is empty, it will still run and won't throw a NPE.

Delphinus8821 (changes requested):
Working well, but can the code smells be fixed?

Quasar985 (approved):

Delphinus8821 (approved):

OrionsGuardian (approved):

Issue #2274, opened 2025-03-12
Export -> GeoPackag Error


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Getting the Export -> GeoPackag -> Save -> Replace Y -> Export then error

![Image](https://github.com/user-attachments/assets/8d0cbc9e-9cba-419f-9d78-121bbdb83025)

### Steps to Reproduce

Verification Process:
1. Open consty
2. Open or create a graph
3. Export a graph via Export -> To GeoPackage..., to a file (leaving everything else default) eg. <filename.gpkg>
4. Export the same file Export -> To GeoPackage... to the same filename eg. <filename.gpkg>
5. Click Yes to replace the existing file
6. Click Export

### Behaviour

I'm getting the Export -> GeoPackag -> Save -> Replace Y -> Export then error  every time.

-->

**Expected behaviour:** GeoPackag File should Export to Target Folder without Error

**Actual behaviour:** Error
![Image](https://github.com/user-attachments/assets/a1060294-bd91-4fe6-8d63-13389d264723)

**Reproduces how often:** 100% times

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

andromeda-224:
To consistently reproduce the IOException:
1. Export a graph via Export -> To GeoPackage..., to a file (leaving everything else default) eg. <filename.gpkg>
2. Export the same file Export -> To GeoPackage... to the same filename eg. <filename.gpkg>
3. Click Yes to replace the existing file
4. Click Export

The IOException pops up.

HolandCrick:
I am having issue with "Export Geo Package ticket #2274" - Instead of clicking "Save"; if I Copy Paste existing file name with path in text box> -> Click Export
It does not give the "Overwrite File" warning!
 
Looks like the ticket #2274 needs some further tuning. Positive thing is it's not giving Exception Error anymore.

HolandCrick:
[...]

PR #2292, opened 2025-04-03
Exporting to same geopkg filename throws exception

andromeda-224:
### Prerequisites

### Description of the Change
Should be able to Export to GeoPackage with the same file without Error.
Geopackage is a SQLite database so there were constraints to prevent updates/re-creation of certain elements (eg. tables, index) with the same name.
Delete geopackage with same filename if it already exists, then re-create a new geopackage with the existing filename,

### Alternate Designs
Previous considerations:
 - Had previously thought of using the store finder to get the store and removing existing content before adding the current content, but the API doesn't seem to have methods to remove spatial index or content for existing features from the previous save.
- I could remove the schema from the previous datastore/geopackage, then add the new feature collection; and I could check if it had an existing spatial index and skip if it already exists, but that won't remove existing content already stored. This might cause the file/database to keep growing and have content that may not be relevant to the current geopackage.
- The graph name might be changed. Even if you overwrite the existing geopackage, and a new spatial index would be generated, but the previous spatial index would still hang around in the geopackage. 
 
So in order to have a clean geopackage, it seems best to delete the old file and create a new one.

To prevent the IOException.

### Benefits
To prevent the IOException when overwriting an existing geopackage.

### Possible Drawbacks
The existing file will be completely deleted.

### Verification Process

1. Open consty
2. Open or create a graph
3. Export a graph via Export -> To GeoPackage..., to a file (leaving everything else default) eg. <filename.gpkg>
4. Export the same file Export -> To GeoPackage... to the same filename eg. <filename.gpkg>
5. Click Yes to replace the existing file
6. Click Export

The file will have been generated without exceptions.

### Applicable Issues
#2274


CoreUtilities/src/au/gov/asd/tac/constellation/utilities/geospatial/Shape.java
```diff
@@ -526,6 +527,10 @@ public static void generateGeoPackage(final String uuid, final Map<String, Strin
         featureEntry.setBounds(ReferencedEnvelope.EVERYTHING);
         featureEntry.setSrid(spatialReference.getSrid());
 
+        // Remove geopackage file if exists as it stores existing tables/indexes
+        if (output.isFile()) {
+            Files.delete(output.toPath());
```
Auriga2, line 532: Wouldn't a simple output.delete(); work here?

Delphinus8821:
The exception is no longer occurring but it now takes a lot longer for the export window to open which is not ideal

CoreUtilities/test/unit/src/au/gov/asd/tac/constellation/utilities/geospatial/ShapeNGTest.java
```diff
@@ -1109,6 +1108,13 @@ public void testGenerateGeoPackage() throws IOException {
                 store.dispose();
             }
         }
+        // try to generateGeoPackage to the same filename
+        try {
+            Shape.generateGeoPackage(id, getTestShapes(), getTestAttributes(), f, Shape.SpatialReference.WGS84);
+            assertTrue(true);  //successful if no exception 
+        } catch (IOException ex) {
```
OrionsGuardian, line 1115: `final`

OrionsGuardian (changes requested):

CoreUtilities/src/au/gov/asd/tac/constellation/utilities/geospatial/Shape.java
OrionsGuardian: While it does work to use `output.delete()` here, the recommendation from SonarQube is to use `Files.delete()`, apparently for better error reporting, reliability and maintainability.

andromeda-224:
With regards to the slowness of the dialog appearing, the deletion of the file occurs after the user has chosen to save the file to an existing geopackage with the same name in the Export window; the Export window taking longer to appear in the first instance might be a separate issue.

OrionsGuardian (approved):

Delphinus8821 (approved):

PR #2298, opened 2025-05-01
Check if output file exists when exporting

andromeda-224:
### Prerequisites

### Description of the Change
When exporting, add a warning to the user when the file in the output textfield already exists in the following scenarios:
1) When file path is pasted into the output
2) Whenever the filepath in output is manually modified

### Alternate Designs
Other considerations were:
- If the warning was displayed when the user clicks on the Export button. This works fine if the user clicks Yes to overwrite, but the issue is if the user clicks No, the Export dialog has already closed (due to the set up of the framework).
- The user could manually modify the output file in the textfield and the file exist check doesn't get validated straight away.
- If we use the isValidPath to invalidate the output (so that the output field background goes red), this disables Export button. This means that the user cannot overwrite the file.

This will be more user friendly and apply all plugins extending AbstractGeoExportPlugin.

### Benefits
This will be more user friendly and apply all plugins extending AbstractGeoExportPlugin.

### Verification Process
Copy and paste an existing valid filepath in the output textfield when exporting to GeoPackage.
A warning notification that the output file already exist should be displayed. Close the notification.
This will allow you to continue to either change the output file or click on the Export button.

### Applicable Issues
#2274

### DIFF
--- a/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/AbstractGeoExportPlugin.java
+++ b/CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/AbstractGeoExportPlugin.java
@@ -120,6 +120,16 @@ public abstract class AbstractGeoExportPlugin extends SimpleReadPlugin {
         FileParameterType.setWarnOverwrite(outputParameter, true);
         parameters.addParameter(outputParameter);
 
+        // Check if file exists when output parameter changes, including error
+        parameters.addController(OUTPUT_PARAMETER_ID, (master, params, change) -> {
+            if (change == ParameterChange.VALUE || change == ParameterChange.ERROR) {
+                final String output = params.get(master.getId()).getStringValue();
+                if (doesFileExist(new File(output))) {
+                    NotifyDisplayer.display("Output file already exists", NotifyDescriptor.WARNING_MESSAGE);
+                }
+            }
+        });       
+
         if (includeSpatialReference()) {
             final PluginParameter<SingleChoiceParameterValue> spatialReferenceParameter = SingleChoiceParameterType.build(SPATIAL_REFERENCE_PARAMETER_ID, SpatialReferenceParameterValue.class);
             spatialReferenceParameter.setName("Spatial Reference");
@@ -161,7 +171,7 @@ public abstract class AbstractGeoExportPlugin extends SimpleReadPlugin {
                     final ReadableGraph readableGraph = activeGraph.getReadableGraph();
                     try {
                         final ParameterValue pv = params.get(master.getId()).getSingleChoice();
-                        if (pv instanceof ElementTypeParameterValue elementTypeParameterValue){
+                        if (pv instanceof ElementTypeParameterValue elementTypeParameterValue) {
                             final GraphElementType elementType = elementTypeParameterValue.getGraphElementType();
                             switch (elementType) {
                                 case TRANSACTION:
@@ -462,11 +472,11 @@ public abstract class AbstractGeoExportPlugin extends SimpleReadPlugin {
             default -> throw new PluginException(PluginNotificationLevel.ERROR, "Invalid element type");
         }
 
-        try {            
+        try {
             //Check for valid path
             if (isValidPath(output)) {
-                exportGeo(parameters, GraphNode.getGraphNode(graph.getId()).getDisplayName(), shapes, attributes, output);  
-            }            
+                exportGeo(parameters, GraphNode.getGraphNode(graph.getId()).getDisplayName(), shapes, attributes, output);
+            }
         } catch (final IOException ex) {
             throw new PluginException(PluginNotificationLevel.ERROR, ex);
         }
@@ -478,18 +488,26 @@ public abstract class AbstractGeoExportPlugin extends SimpleReadPlugin {
                 ConstellationLoggerHelper.SUCCESS
         );
     }
-    
+
     private boolean isValidPath(File output) {
-        if(StringUtils.isEmpty(output.getPath())) {
+        if (StringUtils.isEmpty(output.getPath())) {
             NotifyDisplayer.display("Invalid output file provided, cannot be empty", NotifyDescriptor.ERROR_MESSAGE);
             return false;
         }
-        if(output.isDirectory() || (!output.isDirectory() 
+        if (output.isDirectory() || (!output.isDirectory()
                 && output.getParentFile() != null && output.getParentFile().exists())) {
             return true;
         } else {
             NotifyDisplayer.display("Invalid file path", NotifyDescriptor.ERROR_MESSAGE);
             return false;
-        }        
+        }
     }
-} 
+
+    protected boolean doesFileExist(final File output) {
+        if (Boolean.TRUE.toString().equalsIgnoreCase(System.getProperty("java.awt.headless"))) {
+            return false;
+        }
+        // if file exists
+        return output.isFile();
+    }    
+}


mimosa2 (changes requested):
I tested this by copying the file path. The existing filename warning popup appeared but only option available was 'Ok' which then overwritten the file. There should a question asking 'Do you want to overwrite?' with options 'Yes', 'No' and 'Cancel'.

Delphinus8821:
This solution can cause the popup to appear twice if you overwrite a file from the file chooser and then hit export. A way around this could be to put a flag on the file chooser button that if that button has been pressed and the check box hasn't been edited since, then don't show the popup on export

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,36 @@
 @NbBundle.Messages("ExportToGeoPackagePlugin=Export to GeoPackage")
 public class ExportToGeoPackagePlugin extends AbstractGeoExportPlugin {
 
+    final PluginParameters parameters = new PluginParameters();
```
Delphinus8821, line 49: This parameters variable is unused

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,36 @@
...
+
+    @Override
+    public PluginParameters createParameters() {
+
+        PluginParameters parametersCreated = super.createParameters();
```
Delphinus8821, line 54: final

Delphinus8821 (changes requested):

andromeda-224:
The solution has been updated such that a flag is set when the filechooser's overwrite dialog Yes has been clicked and so that there won't be another dialog popping up after that. If the filechooser dialog has not been clicked on to choose the file and a duplicate file is manually typed (or pasted) into the inputfield, an Overwrite dialog will pop up. In this dialog, if you choose No or Cancel Overwrite, the Export button will be disabled. If you type in a valid filename, the Export button will be enabled.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,36 @@
...
+    @Override
+    public PluginParameters createParameters() {
+
+        final PluginParameters parametersCreated = super.createParameters();
+
+        // Add "listener" on output parameter to check file overwrite
+        parametersCreated.addController(OUTPUT_PARAMETER_ID, (master, params, change) -> {
+            final String output = params.get(master.getId()).getStringValue();
+            final File file = new File(output);
+            FileParameterType.FileParameterValue fileParamValue = (FileParameterType.FileParameterValue) params.get(master.getId()).getParameterValue();
```
OrionsGuardian: `final`

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,36 @@
...
+            // do a doesFileExist check if value changed, path is valid and filechooser dialog not previously opened
+            if (change == ParameterChange.VALUE && super.isValidPath(file) && !fileParamValue.isFileChooserSelected()) {
+                if (doesFileExist(new File(output))) {
+                    String msg = String.format("The file %s already exists. Do you want to replace the existing file?", file.getName());
+                    final NotifyDescriptor nd = new NotifyDescriptor.Confirmation(msg, "Overwrite file", NotifyDescriptor.YES_NO_CANCEL_OPTION, NotifyDescriptor.QUESTION_MESSAGE);
```
OrionsGuardian: NO and CANCEL do the same thing. 
Could just have YES and NO as the options, no need to include a CANCEL option.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,36 @@
...
+                    if (DialogDisplayer.getDefault().notify(nd) == NotifyDescriptor.YES_OPTION) {
+                        return;
+                    } else {
+                        params.get(master.getId()).setError(String.format("The file %s already exists.", file.getName()));                        
+                    }
```
OrionsGuardian: This block can be simplified to:
```
if (DialogDisplayer.getDefault().notify(nd) == NotifyDescriptor.NO_OPTION) {
    params.get(master.getId()).setError(String.format("The file %s already exists.", file.getName()));                        
}
```

OrionsGuardian (changes requested):

CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/FileParameterType.java
```diff
@@ -241,6 +241,7 @@ public static class FileParameterValue extends ParameterValue {
         private ExtensionFilter filter;
         private boolean acceptAllFileFilterUsed;
         private boolean warnOverwrite;
+        private boolean fileChooserSelected;
```
OrionsGuardian, line 244: Instead of a boolean, could we instead hold a list of filenames that the user has clicked "Yes" to overwrite the file.
Using the file chooser would always add an entry to the list (even if it's not overwriting an existing file).
Manually modifying the filename afterwards should check if the new filename is in the list ... if not, check if the filename matches an existing file on the machine and if so then display the overwrite warning. 
If the user clicks on "yes" to overwrite, then add that filename to the list.
Note it should not clear the list if there's an error in the path while the user is typing a new filename.

OrionsGuardian (changes requested):

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/AbstractGeoExportPlugin.java
```diff
@@ -478,18 +478,26 @@ public void read(final GraphReadMethods graph, final PluginInteraction interacti
                 ConstellationLoggerHelper.SUCCESS
         );
     }
-    
-    private boolean isValidPath(File output) {
-        if(StringUtils.isEmpty(output.getPath())) {
+
+    protected boolean isValidPath(File output) {
```
Delphinus8821: final on parameter

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,38 @@
...
+
+        // Add "listener" on output parameter to check file overwrite
+        parametersCreated.addController(OUTPUT_PARAMETER_ID, (master, params, change) -> {
+            final String output = params.get(master.getId()).getStringValue();
+            final File file = new File(output);
+            final FileParameterType.FileParameterValue fileParamValue = (FileParameterType.FileParameterValue) params.get(master.getId()).getParameterValue();
+            // do a doesFileExist check if value changed, path is valid and filechooser dialog not previously opened
+            if (change == ParameterChange.VALUE && super.isValidPath(file) && !fileParamValue.isFileChooserSelected()) {
+                if (doesFileExist(new File(output))) {
+                    String msg = String.format("The file %s already exists. Do you want to replace the existing file?", file.getName());
```
Delphinus8821: final

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,38 @@
...
+                    final NotifyDescriptor nd = new NotifyDescriptor.Confirmation(msg, "Overwrite file", NotifyDescriptor.YES_NO_OPTION, NotifyDescriptor.QUESTION_MESSAGE);
+                    if (DialogDisplayer.getDefault().notify(nd) == NotifyDescriptor.NO_OPTION) {
+                        params.get(master.getId()).setError(String.format("The file %s already exists.", file.getName()));                        
+                    }
+                }
+            } else if (change == ParameterChange.ERROR) {
+                params.get(master.getId()).getParameterValue().setStringValue(output);
+                // If an error occurs, assume the user has typed in the inputbox
+                fileParamValue.setFileChooserSelected(false);
```
OrionsGuardian: This is now redundant with the code change in the next if block.

CoreImportExportPlugins/src/au/gov/asd/tac/constellation/plugins/importexport/geospatial/ExportToGeoPackagePlugin.java
```diff
@@ -41,6 +46,38 @@
...
+            }
+            // reset the FileChooserSelected flag after processing
+            if (fileParamValue.isFileChooserSelected()) {
+                fileParamValue.setFileChooserSelected(false);
+            }
```
OrionsGuardian: Unnecessary if condition. Can just set the value to false every time.

Delphinus8821 (approved):

OrionsGuardian (approved):

Issue #2301, opened 2025-05-13
Find and Replace Enhancements


Delphinus8821:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Would it be possible to include a simple replace with nothing option in the Find and Replace?

I might want to remove the + from some phone numbers such as +6140000000 and replace the + with nothing. This'd be similar to something you might do when cleaning data in Notepad++ for example. Currently, Consty seems to require a value in the replacement field.

It'd also be nice to search within a wider variety of attributes under Node and transaction, too. E.g. Type, source and user created custom attributes.

### Steps to Reproduce

1.  Open Find and Replace

2. Attempt to replace an existing value with nothing, e.g. attempt to remove the '+' from all phone numbers

3. Observe that the replace button does not function unless there is a value in the replace with textbox

HolandCrick:
@Delphinus8821 - Thanks for your help to fix the bug. It's working nicely now.

PR #2315, opened 2025-06-11
Add Ability to Replace with Empty String

Delphinus8821:
### Prerequisites

### Description of the Change
Added the ability to use the Find View to replace the value of an attribute with an empty string. 

Adds additional functionality to the Find View.

### Benefits
Allows users to replace the value or parts of the value of an attribute with an empty string. E.g. replacing the + in a phone number with an empty string.

### Verification Process
1. Create a graph where the identifiers of the nodes are phone numbers
2. Open the Find View and navigate to the Replace tab
3. Type + into the Find textbox and check the With Nothing? checkbox
4. Click Replace Next or Replace All and observe the values of the attributes have been updated

### Applicable Issues
#2301

### DIFF
--- a/CoreFindView/src/au/gov/asd/tac/constellation/views/find/components/ReplaceTab.java
+++ b/CoreFindView/src/au/gov/asd/tac/constellation/views/find/components/ReplaceTab.java
@@ -24,6 +24,7 @@ import au.gov.asd.tac.constellation.views.find.utilities.BasicFindReplaceParamet
 import java.util.ArrayList;
 import java.util.List;
 import javafx.scene.control.Button;
+import javafx.scene.control.CheckBox;
 import javafx.scene.control.Label;
 import javafx.scene.control.TextField;
 import javafx.scene.image.ImageView;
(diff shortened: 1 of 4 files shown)


OrionsGuardian (changes requested):
image

Can we do this without a "With Nothing?" checkbox ?
I would expect having an empty replacement text field would result in using an empty value.

CoreFindView/src/au/gov/asd/tac/constellation/views/find/components/ReplaceTab.java
```diff
@@ -24,6 +24,7 @@
 import java.util.ArrayList;
 import java.util.List;
 import javafx.scene.control.Button;
+import javafx.scene.control.CheckBox;
```
antares1470: this import doesn't appear to be used

CoreFindView/src/au/gov/asd/tac/constellation/views/find/plugins/ReplacePlugin.java
```diff
@@ -44,7 +44,7 @@ public class ReplacePlugin extends SimpleEditPlugin {
     private final GraphElementType elementType;
     private final List<Attribute> selectedAttributes;
     private String findString;
-    private final String replaceString;
+    private String replaceString;
```
antares1470: Given there are no other changes in this class, this can probably remain final

antares1470 (approved):

mimosa2 (approved):

Issue #853, opened 2020-10-12
Issue in the graph created using RecordStoreQueryPlugin when identical nodes present.


Auriga2:
### Description

Cannot split SOME of the nodes created by `Import From RDF` plugin which uses `RecordStoreQueryPlugin`. It throws the error `Attempt to create transaction from source vertex that does not exist`

### Steps to Reproduce
1. Open Constellation from the Constellation Adaptors project. In Data Access View Run the `Import From RDF` plugin with below parameters:
Input File: `https://dbpedia.org/sparql?default-graph-uri=http%3A%2F%2Fdbpedia.org&query=DESCRIBE+%3Chttp%3A%2F%2Fdbpedia.org%2Fresource%2FNicholas_Kyrgios%3E&format=text%2Fplain&CXML_redir_for_subjs=121&CXML_redir_for_hrefs=&timeout=30000&debug=on&run=+Run+Query+`
File Format: `N-Triples`
![image](https://user-images.githubusercontent.com/64826928/95695959-bbbcec80-0c80-11eb-91e2-509f83b7d573.png)

2. Select the below node and run the `Split Nodes Based on Identifier` plugin with `_` as the `split character`
![image](https://user-images.githubusercontent.com/64826928/95696003-f1fa6c00-0c80-11eb-8dc4-788e3b294112.png)

The error below is thrown :
![image](https://user-images.githubusercontent.com/64826928/95696243-e6f40b80-0c81-11eb-8f3b-11a0b2015e7d.png)
Because:
![image](https://user-images.githubusercontent.com/64826928/95697491-700d4180-0c86-11eb-8f15-40beb9050fc9.png)

But if you save the graph, reopen and try the same, all nodes split without any error, because the id2position[] is now rearranged:
![image](https://user-images.githubusercontent.com/64826928/95695913-9334f280-0c80-11eb-8c8c-e71dbb70735b.png)

**Expected behaviour:** 
No error is thrown when splitting.
**Actual behaviour:** 
An error is thrown when splitting.
**Reproduces how often:**
100% but only on some nodes, based on the arrangement.

### Additional Information
n/a

Auriga2:
The same error can be recreated on a graph created by the File importer. Also on a graph manually created: On nodes last added after deleting some previously added nodes. (As explained in the PR)

HolandCrick:
@Auriga2 - Thank you for fixing the Pajek file import issue. It's working nicely now.

PR #2323, opened 2025-06-19
Split nodes exception on certain edited graphs

Auriga2:
### Prerequisites

### Description of the Change
An Illegal Argument Exception was thrown when `Split nodes Based on Identifier` was run on certain nodes of graphs created using File importer or using RecordStoreQueryPlugin. The same issue can be recreated on a graph manually created, as explained below. 

The issue was caused by the confusion between the 2 arrays in ElementStore.java (node position ID and Element ID). The code had used the node position ID instead of the Vertex Id when creating the new node. A usual graph before editing would have identical arrays, so the splitting was still working. 

### Alternate Designs
n/a

Bug Fix

### Benefits
bug fix

### Possible Drawbacks
n/a

### Verification Process

- Create a new graph and add 4 nodes. 
- Remove the Vertex#1 and Vertex#2
- Open DAV. Run the Spit notes plugin on each remaining node, with `#` as the split  character.
- 3 nodes `3`,`4`,`Vertex` should be created without any exception.
- Also verify no exceptions are thrown on graphs created by the File importer or using RecordStoreQueryPlugin, when splitting nodes.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/853

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
@@ -197,20 +197,20 @@ public class SplitNodesPlugin extends SimpleEditPlugin implements DataAccessPlug
         );
         
         //Determine the positions of the selected nodes
-        final List<Integer> selectedVerticies = new ArrayList<>();    
+        final List<Integer> selectedVertexPositions = new ArrayList<>();
         final int graphVertexCount = graph.getVertexCount();
         for (int vertexPosition = 0; vertexPosition < graphVertexCount; vertexPosition++) {
             final int currentVertexId = graph.getVertex(vertexPosition);
             if (graph.getBooleanValue(vertexSelectedAttributeId, currentVertexId)) {
-                selectedVerticies.add(vertexPosition);
+                selectedVertexPositions.add(vertexPosition);
             }
         }
         
-        totalProcessSteps = selectedVerticies.size();
+        totalProcessSteps = selectedVertexPositions.size();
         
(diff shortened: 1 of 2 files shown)


CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/SplitNodesPlugin.java
```diff
@@ -197,20 +197,20 @@ public void edit(final GraphWriteMethods graph, final PluginInteraction interact
...
         
         //Determine the positions of the selected nodes
-        final List<Integer> selectedVerticies = new ArrayList<>();    
+        final List<Integer> selectedVertexPositions = new ArrayList<>();
         final int graphVertexCount = graph.getVertexCount();
         for (int vertexPosition = 0; vertexPosition < graphVertexCount; vertexPosition++) {
             final int currentVertexId = graph.getVertex(vertexPosition);
             if (graph.getBooleanValue(vertexSelectedAttributeId, currentVertexId)) {
-                selectedVerticies.add(vertexPosition);
+                selectedVertexPositions.add(vertexPosition);
```
antares1470: I wonder if the optimal fix here actually just stores the vertex id here instead of the position. Looking at the subsequent for loop code (iterating over the list), the positions are only then used for:
1) Getting the vertex id (line 221)
2) Incorrectly being used where the vertex id should've been used (line 242 and 254, which you've correctly fixed)

You've already addressed 2), but then by storing a list of ids instead of positions, there is also no need for 1)

Auriga2: Agree, removes one line of code and avoids confusion.

Issue #2326, opened 2025-06-22
It shouldn't create a new node using the same Identifier/Type combo that already exists in the graph.


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Constellation is creating new node using the same Identifier/Type combo that already exists in the graph.

### Steps to Reproduce

1.  Open a new graph
2.  Create 3 Nodes
3.  Open Attribute Editor
4. Change one of the Node's Identifier Name to "Vertex #3 "
5. The Consty will creating the 4th Node using the same Identifier and through exception

**Expected behaviour:**  It shouldn't create a new node using the same Identifier and no exception

**Actual behaviour:** Creates new node using the existing Identifier and shown exception

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@Auriga2 - Thanks for fixing the bug. The Consty is not creating  any additional node, while using the same Identifier and not through exception anymore.

PR #2336, opened 2025-07-21
Fix the DuplicateKeyException in Analytic graphs when creating a new element with existing key attributes

Auriga2:
### Prerequisites

### Description of the Change
Analytic graphs now set the key attributes other than the Identifier, before creating the new element, to avoid duplicate key exceptions thrown. For Vertices it's Type and for transactions Type and Datetime. 

When creating a new element, the code uses the key attributes to determine whether it already exists, before incrementing the unique id in the identifier. Prior to these changes, any new element added was always identified as non existing, hence didn't impact new graphs unless edited in between.

### Alternate Designs
N/A, but perhaps we should change the code in the VisualSchemaFactory which uses the DuplicateKeyException thrown by StoreGraph.validateKey, to identify whether a given ID is in use.  An "exists" function would be a better option rather than throw/catch of the exception.

Bug fix

### Benefits
Bug fix

### Possible Drawbacks
N/A

### Verification Process

1. Open a new graph
2. Create 2 Nodes
3. Open Attribute Editor
4. Change one of the Node's Identifier to "Vertex 2 "
5. Add another node. No exceptions should throw and the new node should be added as "Vertex 3"
6. Test the transactions using same steps. 

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2326

### DIFF
--- a/CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/AnalyticSchemaFactory.java
+++ b/CoreAnalyticSchema/src/au/gov/asd/tac/constellation/graph/schema/analytic/AnalyticSchemaFactory.java
@@ -138,15 +138,15 @@ public class AnalyticSchemaFactory extends VisualSchemaFactory {
 
         @Override
         public void newVertex(final GraphWriteMethods graph, final int vertexId) {
-            super.newVertex(graph, vertexId);
-
+            // Other key attributes (TYPE here) have to be set before creating the new vertex in super.newVertex()
             final int vertexTypeAttribute = AnalyticConcept.VertexAttribute.TYPE.ensure(graph);
             final int vertexSourceAttribute = AnalyticConcept.VertexAttribute.SOURCE.ensure(graph);
 
             graph.setObjectValue(vertexTypeAttribute, vertexId, SchemaConceptUtilities.getDefaultVertexType());
             graph.setStringValue(vertexSourceAttribute, vertexId, "Manually Created");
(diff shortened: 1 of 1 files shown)


andromeda-224:
With the transaction testing for step 6, the transaction that is added after the renaming of the identifier must have the same source (as well as datetime value) in order for the identifier to be bumped up. Just wondering if that was intended as I could get the same identifier name with Identifier, type and datetime the same:
image

Auriga2:
Looks like that's the expected behaviour. StoreGraph.ElementKeySet only added primaryKeys in the hash and equals functions.
Then the StoreGraph.TransactionKeySet  adds `Transaction Direction`, `Source Vertex` and `Destination Vertex` also. 

For that reason, I get the same exception when adding a transaction between the same nodes, in the same direction, in v3.2.0-rc1. (Fixed with this PR)

It's the mismatching type (Unknown vs null) which caused the exception, and type needed to be set before creating the transaction using super.newTransaction() in AnalyticSchemaFactory.newTransaction() (And equivalent functions with new node)

If we ever set a default value in any other primaryKeys field, those should also be done before calling super.newVertex() or  super.newTransaction().  

`Transaction Direction`, `Source Vertex` and `Destination Vertex` are already set when adding the transaction in CreateTransactionPlugin.

Quasar985 (approved):

andromeda-224 (approved):

Delphinus8821 (approved):

Issue #2341, opened 2025-07-24
Map View - Default Map Throwing Index Out of Bounds Exception


Delphinus8821:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
If you swap from one of the other maps in the Map View to the default map but you are zoomed in further than you are able to zoom in on the default map it will throw an index out of bounds exception.
<img width="499" height="352" alt="Image" src="https://github.com/user-attachments/assets/2f85c432-baf6-4841-8f5c-7374f6dd5e5e" />

### Steps to Reproduce

1.  Open the Map View and zoom in all the way on one of the maps that isn't the default map 

2.  Swap to the default map 

3.  Observe the exception thrown

**Expected behaviour:** Reset to the highest zoom level for the default map 

**Actual behaviour:** Throws exception

**Reproduces how often:** 100%

HolandCrick:
@Auriga2 - Thanks for fixing the bug. I have tested against 3.2-rc1 dt 04/08/2025 and new code. Looks like job well done. No exception coming anymore.

PR #2374, opened 2025-08-18
Fix the exception when switching from another map to default map

Auriga2:
### Prerequisites

### Description of the Change
When switching from another map with a higher zoom level than the Maximum zoom of default map threw an exception. It has been fixed, and additionally when switching the maps, the zoom range is now reset to the max zoom level of the current map.

### Verification Process
- Open the Map View and zoom in all the way on one of the maps that isn't the default map

- Swap to the default map

- Observe no exception is thrown and verify the zoom bar displays the correct range of the new map.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2341

### DIFF
--- a/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
+++ b/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
@@ -142,6 +142,14 @@ public class MapViewTileRenderer extends PApplet {
 
         currentProvider = provider;
         map.mapDisplay.setMapProvider(provider);
+
+        // Adjust zoom level to fit the new provider
+        if (map.getZoomLevel() > provider.zoomLevels()) {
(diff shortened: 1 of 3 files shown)


andromeda-224 (commented):
I don't get the original exception when testing switching maps, so the changes fixed that exception.
But I'm getting an exception when I first open Constellation and open Map View.
This happens when it opens on default map and ArcGIS Online.
image

Delphinus8821 (approved):
Looks like it is working well. I do get the same exception that @andromeda-224 got but I think that is a separate issue as I have seen it before

Auriga2:
Agree because the exception is thrown after resetContent in MapViewTopComponent, which doesn't seem to be related to new code changes. 
I created https://github.com/constellation-app/constellation/issues/2376 to fix the map view reset issue when resizing, which requires work in that area. I'll add notes there, hopefully we can fix them together.

Issue #2372, opened 2025-08-11
Search textbox doesn't display full view of text when large fontsize set


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When running Constellation at a larger fontsize, the text in the Search textbox does not fit. It looks like the textbox does not expand corresponding to the fontsize.

<img width="552" height="108" alt="Image" src="https://github.com/user-attachments/assets/5c7f393a-5021-4d92-af79-d50605f7ea64" />

### Steps to Reproduce

1.  Run Constellation with --fontsize 26

2.  View the Search textbox on the top right hand of Constellation. Text cannot be fully seen.

The large fontsize also affects some other UI displays such as the Application font panel (cannot be fully seen) Tools -> options -> Constellation -> Application -> Appliation Font/Autosave every 5 minutes/Listen port

<img width="549" height="643" alt="Image" src="https://github.com/user-attachments/assets/b5e10e90-43d9-4322-a0eb-58f92b963716" />

### Behaviour

**Expected behaviour:** To be able to view text when fontsize has changed
**Actual behaviour:**Can't view some text when fontsize is large

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@andromeda-224 - Thank you for fixing the font size bug. It's working now, excellent job mate.

PR #2375, opened 2025-08-19
Modified elements to accommodate text when fontsize is large

andromeda-224:
### Prerequisites

### Description of the Change
Changes were:

1. Default horizontal and vertical width and height for ApplicationOptionsPanel so that text will not be restricted by set height and width expecially when the font size is changed.
2. Modify the JScrollPane of the QuickSearch component to accommodate the new height of the text should the font size be changed.

### Alternate Designs
There didn't seem to be any other way. Had to import netbeans quicksearch module.

### Verification Process
Devs:
1. If running from netbeans, in Constellation project.properties, add --fontsize 20 after the line "-J--add-exports=javafx.base/com.sun.javafx.event\=ALL-UNNAMED" (don't forget to add a backslash after this line)
2. Restart netbeans
3. Run Constellation. Text should be fontsize 20.
4. Tools -> options -> Constellation -> Application and observe that all text are visible.
5. Click in  thet QuickSearch bar on the top right hand side and type something (eg. "Test"). The text should be completely visible.

### Applicable Issues
#2372


CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/startup/Startup.java
```diff
@@ -84,4 +104,30 @@ public void run() {
...
+     * @param source component to traverse.
+     */
+    public void processComponentTree(final Component source) {
+
+        if (source instanceof JScrollPane jsp) {
+            final Dimension origSize = jsp.getSize();
+            // only resize if there is a custom font set
+            if (UIManager.get("customFontSize") != null) {                
+                Integer customFontSize = (Integer) UIManager.get("customFontSize");
+                final Dimension newDimension = new Dimension(origSize.width, 18 * customFontSize / 12);
```
OrionsGuardian: Minor tweak ... change the 18 to 19 for a slightly larger text box.

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/startup/Startup.java
```diff
@@ -84,4 +104,30 @@ public void run() {
...
+    /**
+     * Traverse the component tree to find the right source to set the size.
+     * @param source component to traverse.
+     */
+    public void processComponentTree(final Component source) {
+
+        if (source instanceof JScrollPane jsp) {
+            final Dimension origSize = jsp.getSize();
+            // only resize if there is a custom font set
+            if (UIManager.get("customFontSize") != null) {
```
OrionsGuardian: This if condition should be used to surround the code block on lines 74 to 83, since the rendering issue only occurs when Constellation is launched with a custom font size.

Auriga2:
Noticed the same issue in some other places as below. Should we fix here, or need a  separate ticket?

DAV:
image

Error report:
image

Quasar985 (approved):

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/startup/Startup.java
```diff
@@ -61,7 +70,19 @@ public void run() {
         }
 
         ConstellationSecurityManager.startSecurityLater(null);
-
+        
+        final List<? extends Action> actions = Utilities.actionsForPath("Actions/Edit");
+        for (final Action action : actions) {
+            // only process for quicksearch and if fontsize is customised
+            if (action.getClass().getName().toLowerCase().contains("quicksearchaction") && UIManager.get("customFontSize") != null) {
```
antares1470: The 2nd condition can probably its own if case around the outer for loop since we only really care to loop through actions at all if there is a custom font size (and the case isn't dependent on the value of `action`). This will also prevent unnecessary looping (e.g. currently once you find quick search and you don't have a custom font size, it will continue to loop through subsequent actions, even though no other action will ever satisfy the condition).

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/startup/Startup.java
```diff
@@ -84,4 +105,27 @@ public void run() {
 
         ProxyUtilities.setProxySelector(null);
     }
+
+    /**
+     * Traverse the component tree to find the right source to set the size.
+     * @param source component to traverse.
+     */
+    public void processComponentTree(final Component source) {
+
```
antares1470: remove whitespace line here

CoreFunctionality/src/au/gov/asd/tac/constellation/functionality/startup/Startup.java
```diff
@@ -84,4 +105,27 @@ public void run() {
...
+        if (source instanceof JScrollPane jsp) {
+            final Dimension origSize = jsp.getSize();
+            final Integer customFontSize = (Integer) UIManager.get("customFontSize");
```
antares1470: can the primitive `int` be used here instead? Technically that field can be null, but the only call of this function only calls if the value definitely isn't null

andromeda-224:
> Noticed the same issue in some other places as below. Should we fix here, or need a separate ticket?
> 
> DAV: image
> 
> Error report: image

Created a new ticket as this is the font size set in Constellation app itself.
#2387

nbproject/project.properties
```diff
@@ -125,7 +125,8 @@ run.args.extra=-J-Dconstellation.environment\=IDE(CORE) \
                -J--add-exports=java.base/java.lang\=ALL-UNNAMED \
                -J--add-exports=java.desktop/sun.awt\=ALL-UNNAMED \
                -J--add-exports=java.desktop/sun.java2d\=ALL-UNNAMED \
-               -J--add-exports=javafx.base/com.sun.javafx.event\=ALL-UNNAMED
+               -J--add-exports=javafx.base/com.sun.javafx.event\=ALL-UNNAMED \
+               -J-Dnb.useSwingHtmlRendering=true
```
OrionsGuardian, line 129: Note that this setting will need to also be added to the constellation-app/constellation-applications repo in the constellation.conf file

Issue #2353, opened 2025-07-27
Views->Quality Control View - Delete From Graph Exception


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Views ->Quality Control View - "Delete From Graph" Exception. 

### Steps to Reproduce

1.  Open a new Graph
2.  Open View ->Quality Control View
3.  Select all Nodes
4. Select All Identifier from the list
5. Delete From Graph few time until all nodes deleted

**Expected behaviour:** Should Give Warning or appropriate Error Message

**Actual behaviour:** Throwing Exception

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@andromeda-224 - Thank you for fixing the bug. It's working nicely without exception now.

PR #2378, opened 2025-08-21
Quality Control View Delete From Graph Exception - table not updating

andromeda-224:
### Prerequisites

### Description of the Change

While in minimized view, the Quality Control View was throwing an exception when the nodes were deleted from the Quality Control View table when using Delete From Graph . This occurs after the nodes previously been deleted from the table while the QCV is in minimized mode (but activated) - the reason being that the table had not been updated after the first Delete From Graph.

This is an issue with isVisible field (in AbstractTopComponent) being set to false while in "minimized" mode.
The solution was to override "componentActivated" and set isVisible to true; This method is activated when the Quality Control View is minimized but made visible. This will allow qualityControlViewPane to refresh the view when qualityControlChanged is triggered.

### Benefits
This will update the view immediately.

### Possible Drawbacks
This will update the table even when it's not visible.

### Verification Process

1. Open a new Graph. Add some nodes.
2. Open View ->Quality Control View and minimise the view (If the view is docked normally it will not work)
3. Select all Nodes on the graph.
4. Select all rows from the Quality Control View table list.
5. Select Delete From Graph button at the bottom of the view.
Observe that the table is updated and the nodes have been deleted.

### Applicable Issues
#2353

### DIFF
--- a/CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewTopComponent.java
+++ b/CoreQualityControlView/src/au/gov/asd/tac/constellation/views/qualitycontrol/QualityControlViewTopComponent.java
@@ -106,6 +106,11 @@ public final class QualityControlViewTopComponent extends JavaFxTopComponent<Qua
     protected void componentActivated() {
         setVisible(true);
     }
+    
+    @Override
+    protected void componentDeactivated() {
+        setVisible(false);
+    }
         
     @Override
     protected void handleGraphClosed(final Graph graph) {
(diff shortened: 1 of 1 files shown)


antares1470 (commented):
Did we potentially need to add the inverse operation for this? i.e. override componentDeactivated and set Visible to false (so that it doesn't continue updating when that happens)

mimosa2 (approved):

andromeda-224:
> Did we potentially need to add the inverse operation for this? i.e. override componentDeactivated and set Visible to false (so that it doesn't continue updating when that happens)

Done!

antares1470 (commented):
Never mind. Having just tested with that, it definitely doesn't do what I expected it would, so we might need to change that back. Sorry

andromeda-224:
> Never mind. Having just tested with that, it definitely doesn't do what I expected it would, so we might need to change that back. Sorry

Reverted

antares1470 (approved):

PR #2367, opened 2025-08-01
Add input validation for the Notes View time range fields

sol695510:
### Prerequisites

- [ ] Reviewed the [checklist](CHECKLIST.md)

- [ ] Reviewed feedback from the "Sonar Cloud" bot. Note that you have to wait
    for the "CI / Unit Tests") to complete first. Failed Unit tests can be 
    debugged by adding the label "verbose logging" to the GitHub PR.

### Description of the Change

Added input validation to the Notes View's time range fields to avoid non-numerical character to be entered or the field being left blank.

### Alternate Designs

N/A

### Why Should This Be In Core?

Part of core functionality.

### Benefits

Avoids exceptions being thrown from bad input from a user.

### Possible Drawbacks

N/A

### Verification Process

Attempted to input non-numerical character, such as letters, new changes stops them from being entered.

Cleared the time range field, new changes now restore the value to 0 instead of a blank field.

### Applicable Issues

#2363

PR #2393, opened 2025-09-15
Attribute Editor Update

antares1470:
### Prerequisites

### Description of the Change

A first pass for a code cleanup for the Attribute Editor. In theory, functionality should remain the same (with minor exceptions) but it should now be easier to maintain the code for it.

Notable Changes include:
- Added several unit tests
- Removed the use `DefaultGetter` which appeared to add more complexity to getting a default value than was necessary
- Refactored the way the need for the No Value checkbox is determined by incorporating into the `AbstractEditor` constructor, rather than defined in a separate function
- Re-ordered some of the parameters in the `AbstractEditor` creation, aiming for a more logical order
- Fixed a bug around the date attribute type where the editor type and the native type for the attribute being different caused some conversion issues (especially when restoring default value)
- the bug with the Datetime editor that was previously identified and fixed (#2160) was also occurring for the date and local datetime editors (fixed in the same manor as with the datetime editor).
- Have updated the Decorators Editor UI to make more intuitive

decoratorsUI

This change refactors code in Core, aiming to make it more maintainable.

### Benefits

Attribute editor code should be easier to follow and maintain. Bugs fixed


Delphinus8821 (commented):
Looks good overall.
Playing with the date time editor, it has the same issues that the Notes View one did that were fixed here https://github.com/constellation-app/constellation/pull/2367 
Not sure if its worth fixing it as part of this or as a separate ticket 
image

antares1470:
> Looks good overall.
> Playing with the date time editor, it has the same issues that the Notes View one did that were fixed here #2367
> Not sure if its worth fixing it as part of this or as a separate ticket

My preference would probably be a separate ticket given it also occurs in master (and this PR is already pretty big as is)

Delphinus8821:
> > Looks good overall.
> > Playing with the date time editor, it has the same issues that the Notes View one did that were fixed here #2367
> > Not sure if its worth fixing it as part of this or as a separate ticket
> 
> My preference would probably be a separate ticket given it also occurs in master (and this PR is already pretty big as is)

Created a new ticket for this issue https://github.com/constellation-app/constellation/issues/2396

Quasar985 (approved):

PR #2423, opened 2025-11-10
Welcome page add file exists check and warning

andromeda-224:
### Prerequisites

### Description of the Change
Before attempting to open a file from the Recent Files button, add a file exists check and display warning if it does not exist.

### Alternate Designs
Could display a popup dialog but by displaying the warning notification there's no need to have extra popups and button clicks to close them.

### Benefits
The user will not see an ugly exception dialog popup.

### Verification Process

1. Open Consty and create a new graph. Save and close the graph  (note location and name).
2. Delete the graph created in step 1 outside of Consty (delete from folder).
3. Return to Welcome Page and click on that deleted graph. There should be a file does not exist warning notification at the bottom right hand of Constellation.

### DIFF
--- a/CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
+++ b/CoreWelcomeView/src/au/gov/asd/tac/constellation/views/welcome/WelcomeViewPane.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.prefs.Preferences;
 import javafx.application.Platform;
-import javafx.event.EventType;
 import javafx.geometry.Insets;
 import javafx.geometry.Orientation;
 import javafx.geometry.Pos;
@@ -56,7 +55,6 @@ import javafx.scene.layout.VBox;
 import javafx.scene.paint.Color;
 import javafx.stage.Screen;
 import org.openide.awt.NotificationDisplayer;
-import org.openide.filesystems.FileObject;
 import org.openide.util.Lookup;
 import org.openide.util.NbPreferences;
 


antares1470 (commented):
I thought I would compare what is showing in the welcome page vs Open Recent (since you would presume these to be comparable). I note when you delete the file externally that the file is also removed from Open Recent. I just wonder then whether it is possible to also remove it from the Welcome page? If it is possible, I think that will be a far better user experience than putting up a notification when they try open up a graph that no longer exists

andromeda-224:
> I thought I would compare what is showing in the welcome page vs Open Recent (since you would presume these to be comparable). I note when you delete the file externally that the file is also removed from Open Recent. I just wonder then whether it is possible to also remove it from the Welcome page? If it is possible, I think that will be a far better user experience than putting up a notification when they try open up a graph that no longer exists

The Welcome does refresh when you are away from the Welcome Page and click back to it. It will not refresh if you stay on the Welcome Page ( and do some modification like externally rename/delete the file) and then click on the file from the Recent button. That is when the notification will appear. I don't know of a way to detect if a file has been modified externally bar using a timer or similar mechanism to check continuously, but I'd be worried about performance.

antares1470:
> The Welcome does refresh when you are away from the Welcome Page and click back to it.

That was the issue as I had the welcome page in focus the entire time. Can confirm after another test that deletion did occur after re-focusing.

> I don't know of a way to detect if a file has been modified externally bar using a timer or similar mechanism to check continuously, but I'd be worried about performance.

Agreed, I think a timer-based solution might not be the best option in this case. If there is a good solution that would allow deletion without re-focusing, I think that would be ideal. But perhaps such a solution doesn't exist.

antares1470 (approved):

Delphinus8821 (approved):

Issue #2468, opened 2026-02-11
New Sphere Graph "blinking" rapidly before settling


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When creating a new sphere graph, the graph appears to be "blinking" rapidly before the view is reset and it stops blinking.

### Steps to Reproduce

1.  Open consty and create a sphere graph (leave the defaults)

2.  Observe that the graph "blinks" before the final view is set.

### Behaviour

**Expected behaviour:** Expect the new graph to be created without the strobe effect.

**Actual behaviour:** The new graph image "blinks" rapidly before it settles.

**Reproduces how often:** 100%

### Additional Information

Occurred since the merge of #2459

HolandCrick:
@andromeda-224  - Thank you for fixing the Sphere Graph blinking issue. Looks like it is settled now.

PR #2459, opened 2025-12-23
Refresh the Graph Status line and Graph visuals

andromeda-224:
### Prerequisites

- [ ] Reviewed the [checklist](CHECKLIST.md)

- [ ] Reviewed feedback from the "Sonar Cloud" bot. Note that you have to wait
    for the "CI / Unit Tests") to complete first. Failed Unit tests can be 
    debugged by adding the label "verbose logging" to the GitHub PR.

### Description of the Change
Update/refresh the graph straightaway whenever it is activated (graph is made active in componentActivated).
Update the visuals of the graph whenever graphChanged is triggered.

### Verification Process

1. Open an existing graph with lots of transactions, links and edges. Take note of the numbers on the Graph Status line at the bottom right hand of the app.
2. Select a bunch of transactions and select Tools > Merge Nodes
3. Take note that the numbers on the Graph Status line have changed. Select Undo.
4. The numbers on the Graph Status line should be what it was in step 1 and the configuration on the graph should return to selected transactions in step 2.
5. Select Redo. The numbers on the Graph status line should be what it was in step 3.
6. Select Undo and it should return to what it was.
7. Select the Welcome Page and select the opened graph tab to bring it to the front again. The graph should be immediately displayed on the screen.

### Applicable Issues
#2439, #2447

PR #2469, opened 2026-02-11
Remove unneeded swapBuffers which causes strobe effect

andromeda-224:
### Prerequisites

### Description of the Change
Remove unneeded swapBuffers in VisualGraphTopComponent which causes rapid swapping of the graphics which causes the strobing effect.

So that the strobe effect will not occur.

### Benefits
Stop the strobe effect when graphics are updated rapidly.

### Verification Process
Open consty and create a sphere graph (leave the defaults)

Observe that the graph does not strobe before the final view is set.

### Applicable Issues

#2468 
#2459


Auriga2 (approved):
There's still a flicker, but noticed that it's the same in v3.3.0 as well.

antares1470 (approved):

Issue #2376, opened 2025-08-19
Map view resets to default map when the window or pane is resized


Auriga2:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Map view resets to the default map, default location and default zoom level (4) when the window or pane is resized.

### Steps to Reproduce

1.  Open Map view and switch to another map from the map that opened by default (Not necessarily the default map). Zoom in or out to any level except 4, in a different location.
2.  Resize the Map view pane or the constellation window and notice that it resets to the  map that opened by default.
3.  Notice the same behavior after you zoom in/out to a different location in the map that opens by default as well.

### Behaviour

**Expected behaviour:** The same map resets to the same location/zoom when resizing

**Actual behaviour:** Map resets to a default map/location/zoom when resizing

**Reproduces how often:** 100%

### Additional Information

Should work on this, depending on the state of https://github.com/constellation-app/constellation/issues/1746

Auriga2:
Quoting this exception from https://github.com/constellation-app/constellation/pull/2374#pullrequestreview-3130531710. Hopefully we can fix together.

`I'm getting an exception when I first open Constellation and open Map View.
This happens when it opens on default map and ArcGIS Online.`
<img width="1911" height="843" alt="Image" src="https://github.com/user-attachments/assets/2bb7cf63-bd81-46da-81bd-97d0d0e48d0a" />

antares1470:
This is probably due to the Map View not storing state. Similar to #618

Auriga2:
I couldn't really replicate it- resetting map view after doing analysis in histogram. Perhaps fixed now? 
This is slightly different as this doesn't need to store the state.

HolandCrick:
@Auriga2 - thank you for fixing the map view reset issue. Now the map resets to the same location/zoom when resizing. Excellent work.

Issue #618, opened 2020-07-02
Map view resets to NW Australia when map is docked in graph window


forestfairy:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Map Resets to NW Australia SE Indo after every refresh of data manaipulation rather than remaining over geolocations being analysed

### Steps to Reproduce

1. Open Constellation v2beta5 
2. Import .star data
3. Open map view (this opens docked in graph window)
4. Do analysis in histogram
5. Return to map view to view analysis, centre relevant points in map view
6. Return to histogram to do more analysis
7. Return to Map view..... 

**Expected behaviour:** 
Expect map to remain centred over relevant geolocation located in step 5, with updated analysis points refreshed

**Actual behaviour:** 
When Map view is docked in graph window, the map returns to Indo-Australia each time it refreshes. This does not seem to occur when Map view is docked in Histogram window

**Reproduces how often:** 
100% when map docked in graph window (map view opens in graph window when following tutorials)

### Additional Information

Likelihood 100% when mapview opened as per tutorials
Consequence low
Annoyance very high

aldebaran30701:
This sounds like it could benefit from having the map view maintain state with the position it is centered over the map. Currently as far as I know it does not keep a state, but may benefit from the addition of this and maybe some other features could be tracked in a state attribute for the Map View

formalhaut69:
If I recall correctly there were discussions to completely rework the map view from the ground up. Perhaps a new ticket should be created to re work the map view which includes these features.

HolandCrick:
@arcturus2 thanks for fixing the bug. Now the map to remain centered over relevant geolocation located and , the map does not return to Indo-Australia each time it refreshes. Good work mate.

Issue #1746, opened 2022-08-03
Rewrite Map View


antares1470:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Map View is currently implemented with Swing. Given the majority of other views are implemented in JavaFX, a rewrite of the Map View in JavaFX should be done to closer align it with the other views (both from a visual and code perspective).

### Additional Information

Should ensure that JavaFX version of the Map View uses the View Framework (i.e. it's top component should extend JavaFxTopComponent)

PR #2490, opened 2026-04-23
Map view save zoom, location and provider state

Auriga2:
### Prerequisites

### Description of the Change
Map view no longer resets to the default map, default location and default zoom level (4) when the window or pane is resized. This fix disposes the renderer and rebuilds it for each resize. 

The Alternate Design eliminates that but it can have issues when moving constellation window  between displays with different resolutions, and when reducing the size of the map on certain displays.

### Alternate Designs
https://github.com/constellation-app/constellation/pull/2491

User doesn't have to reselect the map from the drop down again and manually zoom to the required location after resizing.

### Benefits
Better UX

### Possible Drawbacks
Performance limitations if resizing frequently. 

### Verification Process

1. Open Map view and switch to another map from the map that opened by default (Not necessarily the default map). Zoom in or out to any level except 4, in a different location.
2. Resize the Map view pane or the constellation window and notice that it no longer resets to the map that opened by default.
3. Verify the same behavior after you zoom in/out to a different location in the map that opens by default as well.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2376
https://github.com/constellation-app/constellation/issues/618

### DIFF
--- a/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewState.java
+++ b/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewState.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2010-2026 Australian Signals Directorate
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package au.gov.asd.tac.constellation.views.mapview;
+
+import au.gov.asd.tac.constellation.views.mapview.providers.MapProvider;
+import de.fhpotsdam.unfolding.geo.Location;
+
+public final class MapViewState {
+    private static final int DEFAULT_ZOOM = 4;
+    private static final Location DEFAULT_LOCATION = new Location(3.0, 126.0);
+    private final MapProvider defaultProvider;
+    private MapProvider provider;
+    private Location center;
+    private int zoomLevel = -1;
+    private boolean valid = false;
+
+    public MapViewState(MapProvider defaultProvider) {
+        this.defaultProvider = defaultProvider;
+    }
+
+    public void update(MapProvider provider, Location center, int zoomLevel) {
+        this.provider = provider;
+        this.center = center;
+        this.zoomLevel = zoomLevel;
+        this.valid = true;
+    }
+
+    public MapProvider getProvider() {
+        return provider;
+    }
+
+    public Location getCenter() {
+        return center;
+    }
+
+    public int getZoomLevel() {
+        return zoomLevel;
+    }
+
+    public Location getCenterOrDefault() {
+        return center != null ? center : DEFAULT_LOCATION;
+    }
+
+    public int getZoomOrDefault() {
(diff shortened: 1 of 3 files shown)


CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewState.java
```diff
@@ -0,0 +1,68 @@
...
+    private MapProvider provider;
+    private Location center;
+    private int zoomLevel = -1;
+    private boolean valid = false;
+
+    public MapViewState(MapProvider defaultProvider) {
+        this.defaultProvider = defaultProvider;
+    }
+
+    public void update(MapProvider provider, Location center, int zoomLevel) {
```
Quasar985: The function parameters can be final

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewState.java
```diff
@@ -0,0 +1,68 @@
...
+public final class MapViewState {
+    private static final int DEFAULT_ZOOM = 4;
+    private static final Location DEFAULT_LOCATION = new Location(3.0, 126.0);
+    private final MapProvider defaultProvider;
+    private MapProvider provider;
+    private Location center;
+    private int zoomLevel = -1;
+    private boolean valid = false;
+
+    public MapViewState(MapProvider defaultProvider) {
```
Quasar985: The function parameter can be final

Quasar985 (changes requested):

OrionsGuardian:
Comparing the functionality between this PR and the other PR for the Map View reset issue, I prefer this version, even if it does rebuild the view more often. The end result is cleaner and consistent, and doesn't have the same side-effect issues that are occurring in the other PR.

Auriga2:
I agree too. Probably we could implement the other option or similar, after [#1746](https://github.com/constellation-app/constellation/issues/1746)

antares1470:
> Comparing the functionality between this PR and the other PR for the Map View reset issue, I prefer this version, even if it does rebuild the view more often. The end result is cleaner and consistent, and doesn't have the same side-effect issues that are occurring in the other PR.

I agree. My only concern will be if the performance hit becomes significant enough to impact usability of the application. From testing I didn't experience that, but its probably something to note (e.g. does a significant number of markers on the map cause notable issues with redrawing every time)

Auriga2:
> I agree. My only concern will be if the performance hit becomes significant enough to impact usability of the application. From testing I didn't experience that, but its probably something to note (e.g. does a significant number of markers on the map cause notable issues with redrawing every time)

In terms of performance, this is same as v3.3 which disposes the renderer and rebuilds it with the default map for each resize. So hopefully there won't be issues.

Quasar985 (approved):

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewState.java
```diff
@@ -0,0 +1,68 @@
...
+    public Location getCenter() {
+        return center;
+    }
+
+    public int getZoomLevel() {
+        return zoomLevel;
+    }
+
+    public Location getCenterOrDefault() {
+        return center != null ? center : DEFAULT_LOCATION;
```
antares1470: Are there situations where we're ok with `provider` or `center` being null, or `zoomLevel` not being at least 0? 

I note there are the regular getters above this but I'm thinking that if the answer to the above is no (i.e. there aren't any situations where we want that), then perhaps we should applying those logic checks at the point of setting the values (in the `update` function) rather than at the get stage. Means you could also set those fields to the defaults in the constructor (which in turn ensures you avoid any errors from calling getters before `update` is called, and potentially means you don't require the `valid` boolean).

Auriga2: Makes sense. Fixed now.

antares1470 (approved):

PR #2491, opened 2026-04-24
Mapview resize without reinitializing a new renderer each time

Auriga2:
### Prerequisites

### Description of the Change
This works perfectly only on certain displays. It works on my monitor but not the laptop screen:

On the smaller screen it doesn't identify the entire space as available for the map, causing a bigger background. When resizing to a smaller size it doesn't clear, leaving parts of the previous map in the background. This could still happen in the Alternate Design but less likely.

Sometimes this can cause issues when moving constellation window between displays with different resolutions. This can move the map to appear displaced and floating when the constellation window is floating. It fixes when the constellation window is docked. 

This appears to be a limitations with AWT/Swing using floating top‑level native windows that can move between monitors with different DPI/scale, and JOGL/NEWT embedding a native OpenGL child window (GLWindow) via native reparenting (NewtCanvasAWT).

Probably we could implement this or similar after https://github.com/constellation-app/constellation/issues/1746

### Alternate Designs
https://github.com/constellation-app/constellation/pull/2490

### Benefits
Better performance when resizing.

### Verification Process
Same as in https://github.com/constellation-app/constellation/pull/2490

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2376
https://github.com/constellation-app/constellation/issues/618

### DIFF
--- a/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
+++ b/CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
@@ -40,6 +40,7 @@ import de.fhpotsdam.unfolding.events.ZoomMapEvent;
 import de.fhpotsdam.unfolding.geo.Location;
 import de.fhpotsdam.unfolding.ui.BarScaleUI;
 import de.fhpotsdam.unfolding.utils.MapUtils;
+import de.fhpotsdam.unfolding.utils.ScreenPosition;
 import java.applet.Applet;
 import java.awt.Component;
 import java.io.File;
@@ -97,6 +98,10 @@ public class MapViewTileRenderer extends PApplet {
     private int boxDeltaX;
     private int boxDeltaY;
     private boolean updating = false;
+    private boolean resizePending = false;
+    private int cachedWidth;
+    private int cachedHeight;
+    private float cachedZoom;
 
     public MapViewTileRenderer(final MapViewTopComponent parent) {
         this.parent = parent;
@@ -336,6 +341,11 @@ public class MapViewTileRenderer extends PApplet {
         }
     }
 
+    public void resizeContent(final int width, final int height) {
+        resizePending = true;
+        surface.setSize(width, height);
+    }
+
     @Override
     public void settings() {
         size(parent.getContentWidth(), parent.getContentHeight(), PConstants.P2D);
@@ -353,6 +363,7 @@ public class MapViewTileRenderer extends PApplet {
         markerFactory.setDefaultStrokeColor(MarkerUtilities.DEFAULT_STROKE_COLOR);
         markerFactory.setDefaultStrokeWeight(MarkerUtilities.DEFAULT_STROKE_WEIGHT);
 
+        surface.setResizable(true);
         map = new UnfoldingMap(this, "Map View", currentProvider);
         map.setTweening(true);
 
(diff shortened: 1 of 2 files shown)


OrionsGuardian (changes requested):
Image

Auriga2:
Thanks @OrionsGuardian, setting the laptop display `Scale` to 100% fixed most of the issues I listed, (except the occasional scenario you noted above.) Is it still acceptable given the recommended scale for this display is 125%?

The above issue can easily be fixed with a zoom in/out or a window maximizing/minimizing. It happens if you move the vertical divider very fast.

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTileRenderer.java
```diff
@@ -371,11 +382,32 @@ public void setup() {
...
         zoomToLocation(null);
+
+        cachedWidth = width;
+        cachedHeight = height;
     }
 
     @Override
     public void draw() {
         assert !SwingUtilities.isEventDispatchThread();
+        float currentZoom = map.getZoomLevel();
```
OrionsGuardian, line 393: `final`

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTopComponent.java
```diff
@@ -322,23 +320,16 @@ public MapViewTopComponent() {
...
-                scheduledFuture = ConstellationGlobalThreadPool.getThreadPool().getScheduledExecutorService().schedule(() -> {
-                    if (event.getComponent().getWidth() != cachedWidth
-                            || event.getComponent().getHeight() != cachedHeight) {
-                        cachedWidth = event.getComponent().getWidth();
-                        cachedHeight = event.getComponent().getHeight();
-                        resetContent();
-                    }
-                    return null;
-                }, 500, TimeUnit.MILLISECONDS);
+            public void componentResized(ComponentEvent e) {
```
OrionsGuardian, line 331: `final`

CoreMapView/src/au/gov/asd/tac/constellation/views/mapview/MapViewTopComponent.java
OrionsGuardian, line 152: `cachedWidth` and `cachedHeight` don't appear to be used anywhere.
They have values assigned to them, but never used.

OrionsGuardian (changes requested):
Please also address the sonar code smells.

Delphinus8821:
MapViewResizing

Giving this a test, found that when you zoom out on the default map it expands the map over the menu bar

antares1470 (changes requested):
Can confirm on a laptop screen that the map doesn't scale. Also came across a bug where if I open the map view, switch to another view in the same place, and then go back to the map view, it still renders a portion of the previous view in the gaps where the map isn't rendering.

Image


Edit: I'm actually getting the same issue on master so it might not be specific to this change

Auriga2:
It's the same root cause on the smaller screen (screen with a different scale) where it doesn't identify the entire space as available for the map, causing a bigger background. When resizing to a smaller size it doesn't clear, leaving parts of the previous map/view in the background. This could still happen in the `Alternate Design`, but less often. The easy workaround is to  zoom in/out or a window maximizing/minimizing.

I can fix the issue that Delphinus8821 noted above.

Issue #2488, opened 2026-04-08
Boolean or Null Attributes - with a duplicate label - show exception. Need to manage with proper error message


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Boolean or Null Attributes - with a duplicate label - show exception. Need to manage with proper error message.

### Steps to Reproduce

1.  Create a graph with a node and open the Attribute Editor
2. Create a custom node attribute with the attribute type 'boolean or null'
3. Attribute Editor -> Transaction Section - Select "+" button
4. Select "Custom" from dropdown list
5. On the new dialog box give attribute name; select attribute type
6. Click OK
7. Repeat 5 and 6 above; make sure you give "same attribute name" you will end up  exception error

<img width="1163" height="1080" alt="Image" src="https://github.com/user-attachments/assets/1f4516af-aa00-4758-8538-0e7683c32876" />

### Behaviour

**Expected behaviour:**  Error Message showing Duplicate Attribute Name, please check...

**Actual behaviour:** Exception error

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

Auriga2:
The issue happens whenever the user attempts to add another custom attribute with the same `Name` and a different `Type`, not just `boolean or null`

HolandCrick:
@Auriga2 thanks for fixing the bug. The duplicate label - is now identified and controlled. Works fine aa expected.

PR #2492, opened 2026-04-29
Fix the duplicate name exception when adding custom attributes

Auriga2:
### Prerequisites

### Description of the Change
The exception thrown when adding a custom attribute with an existing name but a different type has now been fixed.

### Verification Process
- Open/create a graph and open Attribute Editor
- Create a custom Node/Transaction or Graph  attribute. Note the attribute name and type used.
- Create another custom attribute with the same name and a different type.
- An error message should be displayed instead of an exception.
- Create another custom attribute with the same name and same type. A message is displayed now to inform the user that it already exists.

### Applicable Issues
https://github.com/constellation-app/constellation/issues/2488

### DIFF
--- a/CoreGraphFramework/test/unit/src/au/gov/asd/tac/constellation/graph/AttributeDefinitionNGTest.java
+++ b/CoreGraphFramework/test/unit/src/au/gov/asd/tac/constellation/graph/AttributeDefinitionNGTest.java
@@ -127,10 +127,10 @@ public class AttributeDefinitionNGTest {
         final String name = "node1";
         final String attributeType = "string";
         final GraphWriteMethods graph = new StoreGraph();
-        int attrId1 = graph.addAttribute(elementType, attributeType, name, null, null, null);
-        int attrId2 = graph.addAttribute(elementType, attributeType, name, null, null, null);
+        final int attrId1 = graph.addAttribute(elementType, attributeType, name, null, null, null);
+        final int attrId2 = graph.addAttribute(elementType, attributeType, name, null, null, null);
         assertEquals("Add duplicate Attribute with same Type:", attrId1, attrId2);
-        int attrId3 = graph.addAttribute(elementType, IntegerAttributeDescription.ATTRIBUTE_NAME, name, null, null, null);
+        final int attrId3 = graph.addAttribute(elementType, IntegerAttributeDescription.ATTRIBUTE_NAME, name, null, null, null);
         assertEquals("Add duplicate Attribute with different Type:", attrId1, attrId3);
     }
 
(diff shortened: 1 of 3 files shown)


OrionsGuardian (changes requested):
Some issues found during testing:
Firstly, the dialog pops up on the wrong monitor (always on the default monitor). 
It should be displayed on the same screen where Constellation is Running.
Secondly, when creating a duplicate Transaction attribute, we get this popup:
Image
The "(no merging)" part of the warning shouldn't be included in either the message or the title of the dialog.
I would suggest rewording the error message:
"Unable to create a new custom attribute: test1"
"An attribute with that name already exists"

CoreGraphFramework/test/unit/src/au/gov/asd/tac/constellation/graph/AttributeDefinitionNGTest.java
```diff
@@ -122,14 +122,16 @@ public void changeAttributeDefault() {
...
      */
-    @Test(expectedExceptions = IllegalArgumentException.class)
     public void addAttributeTwice() {
         final GraphElementType elementType = GraphElementType.VERTEX;
         final String name = "node1";
         final String attributeType = "string";
         final GraphWriteMethods graph = new StoreGraph();
-        graph.addAttribute(elementType, attributeType, name, null, null, null);
-        graph.addAttribute(elementType, IntegerAttributeDescription.ATTRIBUTE_NAME, name, null, null, null);
+        int attrId1 = graph.addAttribute(elementType, attributeType, name, null, null, null);
```
Quasar985: `attrId1`, `attrId2` and `attrId3` could be final

Quasar985 (changes requested):

Delphinus8821 (approved):

antares1470 (commented):
For a better user experience, I wonder if it would be possible, instead of simply exiting out after displaying the dialog, to instead reload the custom attribute dialog with the old values already filled in. That way if someone didn't realise a name was already taken, they can quickly adjust it and re-submit, rather than having to go through the process of opening and filling it in again from scratch (which might not be fun if you had included a decent description or default value)

There also appears to still be some unit tests failing

Quasar985 (approved):

OrionsGuardian (changes requested):
Image

It would read better as "A custom Node attribute ..."
However, this is technically not correct.
It will also trigger if there's a default/schema attribute with the same name as well (not necessarily a custom attribute).

Delphinus8821 (approved):

antares1470 (approved):

OrionsGuardian (approved):

Issue #2498, opened 2026-05-07
Cannot extract multiple types from text in Extract Types From Text in DAV


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Attempting to extract multiple types from text only ever returns one.

### Steps to Reproduce

1.  Open Constellation > Data Access View
2. Create a new graph.
3. In Import > Extract Types from Text, type the following in the Text input box: 172.16.0.0, 192.168.0.0, 70.0.0.1
4. Click Go. Only one node is displayed on the graph. There should be 3.

### Behaviour

**Expected behaviour:** Expecting 3 nodes for 3 types in the graph.

**Actual behaviour:** Only one node is displayed in the graph.

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
@andromeda-224 - Thanks for fixing the multi type text extraction issue. It working nicely now.

<img width="1965" height="717" alt="Image" src="https://github.com/user-attachments/assets/60740cb7-128f-4103-983c-1e8a7778b439" />

Issue #2499, opened 2026-05-12
Extract types from Text returns 3 nodes from email address a.b.c@d.e.f.com


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When extracting types from an email (eg. a.b.c@def.com). three nodes were returned: an email type and two host names. Only an email node is expected. 

### Steps to Reproduce

1.  Open Constellation > Data Access View
2. Under Import > Extract Types from Text, type "a.b.c@def.com" in the text box. Click Go button.
3.  Three types are extracted, where there should only be one a.b.c@def.com<Email>, but def.com<Host Name>, a.b.c<Host Name> were also returned

### Behaviour

**Expected behaviour:** Only one type to be created in the graph.

**Actual behaviour:** Three nodes were created in the graph.

**Reproduces how often:** 100%

### Additional Information
Could maybe use the text position to move to the next match after the first match

HolandCrick:
@andromeda-224 - Thanks for fixing the issue relating to 3 nodes creation from email.  Now it's creating only one in the graph.

Issue #596, opened 2020-06-21
Inconsistency Type resolution


arcturus2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Running the latest version of Constellation
	* Attached the ***Support Package*** via `Help` > `Support Package`
    * Checked the FAQs: https://github.com/constellation-app/constellation/wiki/FAQ
    * Checked that your issue isn't already filed: https://github.com/constellation-app/constellation/issues
    * Checked that there is not already a module that provides the described functionality: https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

For the IP 192.168.1.1 there are 3 possible types and using the `AnalyticDominanceCalculator` the one with highest priority is `HOST_NAME` so that's why it's getting picked. Below is the output of the possible types when I run the unit test:

```
Jun 17, 2020 4:34:22 PM au.gov.asd.tac.constellation.graph.schema.type.SchemaVertexTypeUtilities matchVertexTypes
WARNING: Multiple types matched identifier 192.168.1.1: [Network Identifier.IP Address.IPv4 Address, Online Location.Host Name]
```

Note that if I run the 'Extract Types from Text" plugin and put in 192.168.1.1 I get "IPv4 Address" for the type which is inconsistent to Hostname here. The issue there looks like different methods are used. Split nodes uses `SchemaVertexTypeUtilities.matchVertexTypes()` and the applies the `VertexDominantCalculator` whereas "Extract Types from Text" plugin uses `SchemaVertexTypeUtilities.extractVertexTypes()`.

See https://github.com/constellation-app/constellation/pull/568/files/e21df02894d567fd829e0e6839a54b5a58ef4bcd#r441318148 for the original correspondence.

### Steps to Reproduce

TDA

**Expected behaviour:** [What you expect to happen]

**Actual behaviour:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

cygnus-x-1:
This could be the result of those schema regex changes from a while back.

PR #2500, opened 2026-05-12
Allow multiple types to be extracted with Extract Types from Text

andromeda-224:
### Prerequisites

### Description of the Change
If the matching text/type does not match the extractedType identifier in the list of extractedTypes, it should continue to check the rest of the extractedTypes and be added to the list of extractedTypes. Removed the else block so that it does not set isDeficientResult = true and break out of it. 

Also updated the comparison logic so that the higher priority is kept as the extractedType.

This is to fix the bug so that the function returns the expected result.

### Benefits

The expected number of types will be returned.
### Verification Process

1. Open Constellation > Data Access View
2. Create a new graph.
3. In Import > Extract Types from Text, type the following in the Text input box: 172.16.0.0, 192.168.0.0, 70.0.0.1
4. Click Go. Only one node is displayed on the graph. There should be 3. And they should be IPv4 types.

### Applicable Issues
#2498

### DIFF
--- a/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractTypesFromTextPluginNGTest.java
+++ b/CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractTypesFromTextPluginNGTest.java
@@ -155,6 +155,10 @@ public class ExtractTypesFromTextPluginNGTest {
         expResult.set(GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.IDENTIFIER, "abc@def.ghi");
         expResult.set(GraphRecordStoreUtilities.SOURCE + AnalyticConcept.VertexAttribute.TYPE, AnalyticConcept.VertexType.EMAIL_ADDRESS);
         expResult.set(GraphRecordStoreUtilities.SOURCE + AnalyticConcept.VertexAttribute.SEED, "true");
+        expResult.add();
+        expResult.set(GraphRecordStoreUtilities.SOURCE + VisualConcept.VertexAttribute.IDENTIFIER, "def.ghi");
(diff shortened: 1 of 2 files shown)


OrionsGuardian (changes requested):
Image

Testing with the above example ...
The front part of the email address is incorrectly being seen as a hostname.

I'm also wondering if each line of text should generate multiple matches ?
As shown in the above image, each of the HostName entries were sub-components of a bigger match.
Is this the expected/desired behaviour ?

andromeda-224:
> Image
> Testing with the above example ... The front part of the email address is incorrectly being seen as a hostname.
> 
> I'm also wondering if each line of text should generate multiple matches ? As shown in the above image, each of the HostName entries were sub-components of a bigger match. Is this the expected/desired behaviour ?

Yes, I have raised this as a separate issue: https://github.com/constellation-app/constellation/issues/2499

antares1470 (approved):

Auriga2:
> Yes, I have raised this as a separate issue: https://github.com/constellation-app/constellation/issues/2499

In v3.3 an email address creates only one type. Feels like this should be fixed together as this fix introduced the bug?

Probably related to https://github.com/constellation-app/constellation/issues/596 too?

antares1470:
> In v3.3 an email address creates only one type. Feels like this should be fixed together as this fix introduced the bug?
> 
> Probably related to #596 too?

v3.3 only creates one type because as the original issue highlights, it was only returning one result even if there were multiple strings that hit (it ignored substrings, but it also ignored other completely different strings). One could then argue that this fix merely exposes an existing issue, rather than technically introducing it. That being said, there is definitely a case to resolve the other issue in tandem, given if a string returns a result, we're probably not interested in any result for a substring (which has a good chance of not actually being what we're trying to extract).

596 is a slightly different issue as that relates to the which type a string resolves to when it gets a hit on more than one type's detection regex (as appose to this one where the issue is there are multiple strings with hits).

Quasar985 (approved):

Auriga2 (approved):

Issue #2496, opened 2026-05-06
Update help tooltips in Data Access View


andromeda-224:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Add more information in tooltips for DAV plugins. Extend tooltip display duration for longer tooltip text.

### Steps to Reproduce

1.  Open Constellation > Views > Data Access Views

2.  Hover over each plugin - a helpful tooltip hint for that plugin should be displayed

### Behaviour

**Expected behaviour:** Add more info to tooltips. For longer tooltip text, tooltip should be displayed a little longer.

**Actual behaviour:** 

**Reproduces how often:** 

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

HolandCrick:
The tooltips in Data Access View works fine. Thanks for the enhancement. But larger text box is overlapping the graph, would be nice to limit within left panel. Also, the font size is too small; consider 1 size bigger font.

HolandCrick:
@andromeda-224 Thank for updating the tooltips format. It's now word warping nicely.

PR #2501, opened 2026-05-12
Update dav tooltips

andromeda-224:
### Prerequisites

### Description of the Change
Format the text of long tooltips in DAV plugins so that they are a shorter width.

### Alternate Designs
The Font Size of the tooltips is set to the application's font size. You can set the font size to a larger font and all the application fonts will automatically be set to that size.

Also tried to set the tooltip to be displayed at a certain X,Y point with the following code in DataSourceTitledPane constructor :
 

```
getTooltip().setOnShowing(s -> {

            //Get button current bounds on computer screen
            Bounds bounds = this.localToScreen(label.getBoundsInLocal());
            if (bounds != null) {                
                getTooltip().setX(bounds.getMaxX()/2);
                getTooltip().setY(bounds.getMinY());
            }
        });
```
But this had some side effects such as the tooltip display duration did not seem to apply (very short duration display) and if you move the application to a different screen/monitor, it would still be displayed on the primary screen.

### Verification Process

1. Open Constellation > Views > Data Access Views
2. Hover over each plugin - a helpful tooltip hint for that plugin should be displayed. For longer tooltips, text should be displayed for at least 10 seconds.

### Applicable Issues
#2496

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
@@ -94,7 +94,7 @@ public class MergeNodesPlugin extends SimpleQueryPlugin implements DataAccessPlu
 
     @Override
     public String getDescription() {
-        return "Merge nodes in your graph together";
+        return "Merge nodes in your graph together based on the defined criteria.";
     }
 
     protected static final Comparator<String> LONGEST_VERTEX_CHOOSER = (o1, o2) -> {
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
@@ -94,7 +94,9 @@ public class MergeNodesPlugin extends SimpleQueryPlugin implements DataAccessPlu
 
     @Override
     public String getDescription() {
-        return "Merge nodes in your graph together based on the defined criteria.";
+        return """
+               Merge nodes in your graph together based on the
+               defined criteria.""";
     }
 
     protected static final Comparator<String> LONGEST_VERTEX_CHOOSER = (o1, o2) -> {
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeTransactionsPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeTransactionsPlugin.java
@@ -101,7 +101,7 @@ public class MergeTransactionsPlugin extends SimpleQueryPlugin implements DataAc
     @Override
     public String getDescription() {
         return """
-               Merge groups of transactions (of the same type"
+               Merge groups of transactions (of the same type
                and activity) between two nodes together.""";
     }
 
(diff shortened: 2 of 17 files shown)


CoreDataAccessView/test/unit/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeTransactionsPluginNGTest.java
```diff
@@ -131,7 +131,9 @@ public void testGetPosition() {
     @Test
     public void testGetDescription() {
         MergeTransactionsPlugin instance = new MergeTransactionsPlugin();
-        String expResult = "Merge groups of transactions (of the same type and activity) between two nodes together.";
+        String expResult = """
+               Merge groups of transactions (of the same type"
```
Auriga2: Accidental extra " after type?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeTransactionsPlugin.java
```diff
@@ -100,7 +100,9 @@ public int getPosition() {
 
     @Override
     public String getDescription() {
-        return "Merge groups of transactions (of the same type and activity) between two nodes together.";
+        return """
+               Merge groups of transactions (of the same type"
```
Auriga2: Probably an accidental extra " after type?

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
```diff
@@ -94,7 +94,9 @@ public int getPosition() {
 
     @Override
     public String getDescription() {
-        return "Merge nodes in your graph together based on the defined criteria.";
+        return """
```
Auriga2, line 97: Rather than hardcoding, would it be possible to determine the width programmatically based on the DAV width?

Auriga2 (changes requested):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
andromeda-224: Thinking that it might be good to be the same size each time as the panel could be narrowed too much. That way, you can still read the tooltips.

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/RemoveUnusedAttributesPlugin.java
```diff
@@ -56,7 +56,10 @@ public String getName() {
 
     @Override
     public String getDescription() {
-        return "Removes all (non-key) node and transaction attributes on the graph that are not being used (i.e. contain no values anywhere on the graph).";
+        return """
+               Removes all (non-key) node and transaction attributes
+               on the graph that are not being used (i.e. contain no
+                values anywhere on the graph).""";
```
antares1470: extra space at the beginning here could probably be removed

Quasar985 (approved):

antares1470 (approved):

CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/clean/MergeNodesPlugin.java
Auriga2: My only concern is then we have to remember and update the length manually for every new plugin. 

I was wondering whether we can do it programmatically, then you don't need to bother changing each plugin, and it can be the DAV width and set the above width as min, if we want to avoid it being too narrow. Probably not worth the effort now that you've fixed all existing plugins, and if new plugins are rare.

Auriga2 (approved):

PR #2505, opened 2026-05-25
Extract Types from Text Update to handle substrings of identifier

andromeda-224:
### Prerequisites

### Description of the Change
Update Extract types from Text to return the longest/highest dominant identifier node from email address a.b.c@d.e.f.com instead of 3 nodes.

The condition to compare dominance is:
If the new identifier picked up by detectionRegex is the same identifier as previously extracted type.
OR if the new identifier picked up by detectionRegex is a shorter substring of previously extracted type.

Else if the previously existing extractedType is higher dominance than the currentExtractedType, then remove the existing extractedType and add the current extractedType to the list.

### Alternate Designs
Alternate design was to obtain all matching identifiers and do a compare afterward. This would be similar to the current schemaType loop, but would also require another for loop to iterate through and compare again again.

### Benefits
This will return the highest dominant schema type for a matching identifier that was picked up with regex, rather than 3 types for the same text.

### Verification Process

1. Open Consty and a new graph.
2. Open Data Access View and under Extract Types from text, type in: friend@gmail.com , a.b.c@def.com, 172.16.0.0, 192.168.0.0, 70.0.0.1
3. There should be 5 nodes returns (2 emails, and 3 host names)

### Applicable Issues
#2499


antares1470 (approved):

Quasar985 (approved):

antares1470:
Just noting there are some unit tests failing so we won't be able to merge this until those are fixed

Issue #2510, opened 2026-06-10
Word Cloud Attribute dropdown does not update when Element Type is changed


Delphinus8821:
### Prerequisites

- [x] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

When using the Word Cloud View in v3.4.0-rc1, the Attribute dropdown list does not update if the Element Type is changed from the default

### Steps to Reproduce

1. Open a graph and the Word Cloud View

2.  Click on the Attribute drop down list, confirm some options load

3.  Change the Element Type to node and then try to change the value in the Attribute drop down, the options no longer load 

### Behaviour

**Expected behaviour:** Attribute drop down displays values based on the Element Type 

**Actual behaviour:** Attribute drop down clears when the Element Type is changed

**Reproduces how often:** 100%

### Additional Information

Introduced in v3.4.0-rc1

HolandCrick:
@Delphinus8821  - Thanks for addressing the issue.  I am having issue in Word Cloud Attribute Dropdown List; once select any option, all the options running non-stop and selected option not settling in the field. Can you please look into this again.

HolandCrick:
@Delphinus8821 - Thanks for your help to fix the bug. Working nicely now.

PR #2524, opened 2026-06-16
Fix Attribute dropdown in Word Cloud View 

Delphinus8821:
### Prerequisites

### Description of the Change
Updated the order that options are added to the dropdowns in the Word Cloud View to ensure that the options are set before the string value is updated.

### Alternate Designs
An alternative solution would be to add a fireChangeEvent to setOptions in the SingleChoiceParameterType which would catch if this happens anywhere else.

Fixes a bug in core

### Verification Process
1. Open or create a sphere graph
2. Open the Word Cloud View 
3. Change the Element Type dropdown to node
4. Select a value in the Attribute dropdown, the list should be populated with options

### Applicable Issues
#2510

### DIFF
--- a/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/SingleChoiceParameterType.java
+++ b/CorePluginFramework/src/au/gov/asd/tac/constellation/plugins/parameters/types/SingleChoiceParameterType.java
@@ -140,6 +140,9 @@ public class SingleChoiceParameterType extends PluginParameterType<SingleChoiceP
 
             parameterValue.setOptions(options);
             parameter.setProperty(CHOICES, new Object());
+            
+            // TODO: find out why this is needed after the single choice input was updated 
+            parameter.fireChangeEvent(ParameterChange.VALUE);
         }
     }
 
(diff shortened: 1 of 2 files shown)


andromeda-224:
I'm wondering if the issue could be because the options are set after the setStringValue? I'm thinking maybe in the params.addController code for ParameterChange.VALUE in WordCloudParametersPane, if you move the setOptions to somewhere before the setStringValue(), it would populate with the options?

image

Delphinus8821:
> I'm wondering if the issue could be because the options are set after the setStringValue? I'm thinking maybe in the params.addController code for ParameterChange.VALUE in WordCloudParametersPane, if you move the setOptions to somewhere before the setStringValue(), it would populate with the options?
> 
> image

This does work so I will change the PR to this but it doesn't really make any sense as to why this worked before the single choice inputs were updated.

andromeda-224 (approved):

OrionsGuardian (approved):

Quasar985 (approved):

Auriga2 (approved):

antares1470 (approved):

Issue #2547, opened 2026-06-29
Word Cloud Plugin does not load the Attribute field after creating a new graph while docked.


HolandCrick:
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

Word Cloud Plugin does not load the Attribute field after creating a new graph while docked. 

### Steps to Reproduce

1. Open the Word Cloud plugin and keep it docked. 
2. Create a new graph. 
3. Open or switch back to the Word Cloud plugin.
4. Word Cloud Plugin does not load the Attribute field after creating a new graph while docked
5. Changing the Element Type from Node to Transaction correctly updates the Attribute list, though there a blank row. However, switching it back to Node does not refresh the list; it still displays the transaction attributes.

### Behaviour

**Expected behaviour:** Attribute field for Node should display

**Actual behaviour:** Word Cloud Plugin does not load the Attribute field after creating a new graph while docked.

**Reproduces how often:** 100%

### Additional Information

Any additional information, configuration or data that might be
necessary to reproduce the issue.

Auriga2:
Copying more details on recreating in Dev: 

> @OrionsGuardian :Can confirm, I'm seeing the same issue in the Dev environment with the Attribute field not being populated ... specifically when running from the Adaptors module. The issue doesn't occur in dev when we run only the core module.

OrionsGuardian:
Testing in 3.4rc2 showed some unusual default behaviour when a graph has a Content attribute.
Consider the case where you have 2 graphs: graph A and graph B, but only graph B has a Content attribute.
On graph A, set Word Cloud attribute to **Label**, switch to graph B and notice that the selected attribute gets cleared
(it tries to change to **Content** even though the Label attribute would have been valid in graph B)
Now switch back to graph A and it will retain the empty attribute setting since graph A doesn't have a Content attribute.
Switch again to graph B and it will now auto-select the **Content** attribute.

The whole behaviour sequence is unintuitive.

[...]

Issue #2546, opened 2026-06-26
Data Access View - Extract Words from Text - the Content Attribute field has no values


GammaVel:
### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description

The Data Access View has the option to run "Extract Words from Text". Here you must first pick one of the Content Attributes to run this on, as it is a mandatory field. This field currently has no values shown. This was working in the previous versions. 

### Steps to Reproduce

1.  Ensure you have the latest available version of Constellation (v 3.4 RC1)
2.  Open any graph
3.  Open the Data Access View
4.  Go to the "Extract words from Text"
5.  Click on the Content Attribute - there are no options to select!
6.  Check in a previous version of Constellation, the options are there.

### Behaviour

**Expected behaviour:** The Content Attribute has options to select

**Actual behaviour:** There are no options

**Reproduces how often:** 100 % 

### Additional Information

<img width="871" height="337" alt="Image" src="https://github.com/user-attachments/assets/2a7ae360-6ac4-4144-8ce3-7fce327b93d6" />

Here are the options I am expecting to be able to use from v3.33

<img width="631" height="168" alt="Image" src="https://github.com/user-attachments/assets/5cb70469-b14d-4220-aca1-b572d18c51b6" />

OrionsGuardian:
Testing in rc2, the Content Attribute doesn't get updated when switching between graphs, where each graph has a different set of attributes.

HolandCrick:
@andromeda-224 - Thanks for fixing the bug. Now the Content Attribute get updated when switching between graphs, where each graph has a different set of attributes.

PR #2554, opened 2026-07-02
Fix for extact words trigger and content being set to empty string on…

andromeda-224:
… selection

### Prerequisites

### Description of the Change
1) Extract Words from Text suppressEvent was suppressing the fireChangeEvent in the Choices property which sets the choice in setFieldValue. Remove the SuppressEvent to allow the choice to be set in the dropdown list.

2) When switching between graphs where Content attribute does and does not exist on graph, the choice selected is set to empty string/null for when the Content attribute is available. This is because the options for that graph had not been set before setStringValue is called. Moved the setOptions to above setStringValue call.

### Verification Process

1. Open Consty. 
2. Open a graph with no content attribute in any of the transactions.
3. Open a graph with content attribute in any of the transactions.

For Extract Words from Text verification:

1. Open Data Access View -> Extract Words from Text
2. Content Attribute should default to Activity (first in the list)
3. Switch to graph with content. Activity should still be selected in Content Attribute.
4. Check the Content Attribute drop down list, drop down list should list Content. 

For Word Cloud View verification:

1. Open Word Cloud view
2. Select Element Type = transaction. Select Attribute = Activity.
3. Switch to graph with content attribute. Attribute selection should not change. Dropdown list should list content.
4. Switch to graph with no content. Attribute selection should not change.  Drop down list does not list content.
5. Clear the Attribute selection (select the blank option). Switch to graph with Content. The Word Cloud View should auto-select Content in the Attribute dropdown.
6. Switch back to the graph with no content and the Word Cloud Attribute should be cleared.

### Applicable Issues
#2546 #2547

### DIFF
--- a/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractWordsFromTextPlugin.java
+++ b/CoreDataAccessView/src/au/gov/asd/tac/constellation/views/dataaccess/plugins/importing/ExtractWordsFromTextPlugin.java
@@ -251,9 +251,7 @@ public class ExtractWordsFromTextPlugin extends SimpleQueryPlugin implements Dat
 
             @SuppressWarnings("unchecked") //ATTRIBUTE_PARAMETER will always be of type SingleChoiceParameter
             final PluginParameter<SingleChoiceParameterValue> contentAttribute = (PluginParameter<SingleChoiceParameterValue>) parameters.getParameters().get(ATTRIBUTE_PARAMETER_ID);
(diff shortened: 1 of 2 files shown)


Quasar985 (commented):
~~For the word cloud verification, in step 5 the word cloud didn't automatically select the content attribute when swapping to the graph with that attribute. The selected attribute stayed blank, which is what I would expect to happen anyway~~

Nevermind, I was using Content.Author, if it is just Content as an attribute, then it works as expected

Quasar985 (approved):

antares1470 (approved):

Delphinus8821 (approved):

PR #2565, opened 2026-07-31
Fix Notes View Null Pointer Exception

### LINKED ISSUE #1528: Notes View Null Pointer Exception
### Prerequisites

- [ ] Put an X between the brackets on this line if you have done all of
    the following:

    -   Running the latest version of Constellation

    -   Attached the ***Support Package*** via `Help` &gt;
        `Support Package`

    -   Checked the FAQs:
        https://github.com/constellation-app/constellation/wiki/FAQ

    -   Checked that your issue isn’t already filed:
        https://github.com/constellation-app/constellation/issues

    -   Checked that there is not already a module that provides the
        described functionality:
        https://github.com/constellation-app/constellation/wiki/Catalogue-of-Repositories

### Description
Linux OS.
Current Master.
Running via NetBeans.
When working on the graph with Notes View opened, an exception occurs.
```
	at au.gov.asd.tac.constellation.views.notes.NotesViewPane.createNote(NotesViewPane.java:622)
	at au.gov.asd.tac.constellation.views.notes.NotesViewPane.lambda$updateNotesUI$13(NotesViewPane.java:517)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1540)
	at au.gov.asd.tac.constellation.views.notes.NotesViewPane.lambda$updateNotesUI$14(NotesViewPane.java:516)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
[catch] at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)

```

### Steps to Reproduce

1. Open Constellation
2. Open New Graph
3. Add a few vertices
4. Enable the Notes View filter to show all note types
5. Create one user note
6. Use select in box to make some auto notes
7. Experimental -> build graph -> sphere graph builder

**Expected behaviour:** \[What you expect to happen\]
No exception to occur
**Actual behaviour:** \[What actually happens\]
Exception occurs.
**Reproduces how often:** \[What percentage of the time does it reproduce?\]
20%
### Additional Information
The opening of the Sphere Graph Builder Pane may not be a prerequisite of this issue being replicated as it seems to stem from the "createNotes()" method which appears to be updating the notes when a new plugin report occurs.

arimu1:
### Prerequisites

### Description of the Change

When the Notes View rebuilt its UI (for example while auto notes were being created from plugin reports such as the Sphere Graph Builder), `createNote` could throw a `NullPointerException`.

The stack trace from #1528 pointed at `NotesViewPane.createNote`. At the time the failure was `contentTextArea.getText().length()` when a note had null content (common for in-progress plugin reports). The TextArea path has since been replaced, but the same null sources still exist and can NPE elsewhere in `createNote` and related selection helpers:

- `NotesViewEntry` could store a null title/content/colour (constructor, setters, and `pluginReportChanged`)
- selection lists (`nodesSelected` / `transactionsSelected`) were dereferenced without null checks, while other call sites already guarded them
- `ConstellationColor.fromHtmlColor(...).getJavaFXColor()` NPE'd when the colour string was invalid

This change:

1. Normalises null title/content to empty strings in `NotesViewEntry` (construction, setters, plugin report listener)
2. Ignores null colour on `setNodeColour` (same approach as the constructor)
3. Null-guards selection lists in `createNote`, "Select on Graph", and add/remove selected element helpers, matching the existing style used elsewhere in the pane
4. Uses `StringUtils.isBlank` for colour and falls back to the default user colour when `fromHtmlColor` returns null
5. Adds unit tests for the null-safe `NotesViewEntry` behaviour
6. Updates `CHANGELOG.md`

### Alternate Designs

Hardening only the UI layer was considered, but normalising nulls at the model (`NotesViewEntry`) prevents the same values from causing NPEs in future callers (state IO, listeners, markdown rendering).

Bug fix for an existing Core Notes View feature.

### Benefits

- Notes View no longer crashes when rendering auto notes or notes with incomplete selection/colour data
- Behaviour is consistent with existing null checks already used for selection lists in the same class

### Possible Drawbacks

None expected; null values are replaced with empty/default values rather than changing valid note data.

### Verification Process

- Reviewed the original #1528 stack against the historical `createNote` implementation and current code
- Confirmed no open prior PR targets this issue
- Added `NotesViewEntryNGTest` covering null title/content/colour and null plugin report messages
- Full NetBeans/ant suite not run locally (ant not available in this environment); CI unit tests should cover the module

### Applicable Issues

Fixes #1528

### DIFF
--- a/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
+++ b/CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
@@ -843,9 +843,8 @@ public class NotesViewPane extends BorderPane {
                 selectionLabelText = "Note linked to: the graph.";
             } else {
                 selectionLabelText = "Note linked to: ";
-                // Null-guard: selection lists can be null for some persisted/user notes.
-                final int nodeCount = newNote.getNodesSelected() != null ? newNote.getNodesSelected().size() : 0;
-                final int transactionCount = newNote.getTransactionsSelected() != null ? newNote.getTransactionsSelected().size() : 0;
+                final int nodeCount = newNote.getNodesSelected().size();
+                final int transactionCount = newNote.getTransactionsSelected().size();
                 if (nodeCount == 1) {
                     selectionLabelText += nodeCount + " node, ";
                 } else {
@@ -1008,20 +1007,16 @@ public class NotesViewPane extends BorderPane {
                     // Select the specific nodes and/or transactions applied to the note.
                     // Add nodes that are selected to the note.
                     final List<Integer> selectedNodes = newNote.getNodesSelected();
-                    if (selectedNodes != null) {
-                        final int nodesLength = selectedNodes.size();
-                        for (int i = 0; i < nodesLength; i++) {
-                            elementIdsVx.set(selectedNodes.get(i));
-                        }
+                    final int nodesLength = selectedNodes.size();
+                    for (int i = 0; i < nodesLength; i++) {
+                        elementIdsVx.set(selectedNodes.get(i));
                     }
 
                     // Add transactions that are selected to the note.
                     final List<Integer> selectedTransactions = newNote.getTransactionsSelected();
-                    if (selectedTransactions != null) {
-                        final int transactionsLength = selectedTransactions.size();
-                        for (int i = 0; i < transactionsLength; i++) {
-                            elementIdsTx.set(selectedTransactions.get(i));
-                        }
+                    final int transactionsLength = selectedTransactions.size();
+                    for (int i = 0; i < transactionsLength; i++) {
+                        elementIdsTx.set(selectedTransactions.get(i));
                     }
                 }
 
@@ -1057,7 +1052,7 @@ public class NotesViewPane extends BorderPane {
                 }
             });
 
-            if (newNote.getNodesSelected() != null && newNote.getTransactionsSelected() != null && newNote.getNodesSelected().isEmpty() && newNote.getTransactionsSelected().isEmpty()) {
+            if (newNote.getNodesSelected().isEmpty() && newNote.getTransactionsSelected().isEmpty()) {
                 addOnGraphMenuItem.disableProperty().set(true);
                 removeOnGraphMenuItem.disableProperty().set(true);
             }
@@ -1156,8 +1151,7 @@ public class NotesViewPane extends BorderPane {
             if (noteToEdit.isGraphAttribute()) {
                 noteToEdit.setGraphAttribute(false);
             }
-            final List<Integer> originalNodes = noteToEdit.getNodesSelected() != null
-                    ? noteToEdit.getNodesSelected() : new ArrayList<>();
+            final List<Integer> originalNodes = noteToEdit.getNodesSelected();
             nodesSelected.forEach(node -> {
                 if (!originalNodes.contains(node)) {
                     originalNodes.add(node);
@@ -1170,8 +1164,7 @@ public class NotesViewPane extends BorderPane {
             if (noteToEdit.isGraphAttribute()) {
                 noteToEdit.setGraphAttribute(false);
             }
-            final List<Integer> originalTransactions = noteToEdit.getTransactionsSelected() != null
-                    ? noteToEdit.getTransactionsSelected() : new ArrayList<>();
+            final List<Integer> originalTransactions = noteToEdit.getTransactionsSelected();
             transactionsSelected.forEach(transaction -> {
                 if (!originalTransactions.contains(transaction)) {
                     originalTransactions.add(transaction);
@@ -1187,7 +1180,7 @@ public class NotesViewPane extends BorderPane {
     public void removeFromSelectedElements(final NotesViewEntry noteToEdit) {
         updateSelectedElements();
 
-        if (!nodesSelected.isEmpty() && noteToEdit.getNodesSelected() != null) {
+        if (!nodesSelected.isEmpty()) {
             final List<Integer> originalNodes = noteToEdit.getNodesSelected();
             nodesSelected.forEach(node -> {
                 if (originalNodes.contains(node)) {
@@ -1198,7 +1191,7 @@ public class NotesViewPane extends BorderPane {
             noteToEdit.setNodesSelected(originalNodes);
         }
 
-        if (!transactionsSelected.isEmpty() && noteToEdit.getTransactionsSelected() != null) {
+        if (!transactionsSelected.isEmpty()) {
             final List<Integer> originalTransactions = noteToEdit.getTransactionsSelected();
             transactionsSelected.forEach(transaction -> {
                 if (originalTransactions.contains(transaction)) {
@@ -1209,9 +1202,7 @@ public class NotesViewPane extends BorderPane {
             noteToEdit.setTransactionsSelected(originalTransactions);
         }
 
-        final List<Integer> nodes = noteToEdit.getNodesSelected();
-        final List<Integer> transactions = noteToEdit.getTransactionsSelected();
-        if ((nodes == null || nodes.isEmpty()) && (transactions == null || transactions.isEmpty())) {
+        if (noteToEdit.getNodesSelected().isEmpty() && noteToEdit.getTransactionsSelected().isEmpty()) {
             noteToEdit.setGraphAttribute(true);
         }
     }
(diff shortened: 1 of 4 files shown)


CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -896,13 +899,19 @@ private void createNote(final NotesViewEntry newNote) {
         gap.setMinWidth(10);
         gap2.setMinWidth(10);
 
-        if (newNote.getNodeColour().isBlank()) {
+        if (StringUtils.isBlank(newNote.getNodeColour())) {
             newNote.setNodeColour(USER_COLOR);
         }
```
antares1470: I think this check can be removed, as the subsequent null check should also account for blank (since `ConstellationColor.fromHtmlColor` returns null if the color doesn't start with '#'), and it also sets the node color of `newNote` to `USER_COLOR`

CoreNotesView/test/unit/src/au/gov/asd/tac/constellation/views/notes/state/NotesViewEntryNGTest.java
```diff
@@ -0,0 +1,156 @@
...
+                false,
+                false,
+                "#ffffff",
+                false
+        );
+
+        assertEquals(entry.getNoteTitle(), "");
+        assertEquals(entry.getNoteContent(), "");
+        assertFalse(entry.isUserCreated());
+        assertNotNull(entry.getNodeColour());
```
antares1470: since we know what the value of `entry.getNodeColour()` should be in this case, can we do an assertEquals for that instead? That will be a stronger assert than simply checking for not null (given the majority of not null values are also incorrect)

antares1470 (changes requested):

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
arimu1: Done on tip `9c3de91cd` — removed the blank pre-check; the existing `fromHtmlColor` null fallback covers blank/non-`#` colours and still sets `USER_COLOR`.

CoreNotesView/test/unit/src/au/gov/asd/tac/constellation/views/notes/state/NotesViewEntryNGTest.java
arimu1: Done — `assertEquals(entry.getNodeColour(), "#ffffff")` for the colour passed into that constructor case.

arimu1:
@antares1470 Thanks for the review — both points addressed on tip `9c3de91cd`:

1. Dropped the redundant `StringUtils.isBlank` colour pre-check; `ConstellationColor.fromHtmlColor` already returns null for blank/invalid colours, and the subsequent null path sets `USER_COLOR`.
2. Strengthened `testConstructorNullTitleAndContent` to `assertEquals(..., "#ffffff")` for the colour supplied to that constructor.

Happy to adjust further if anything else stands out.

Delphinus8821 (approved):

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -843,15 +843,18 @@ private void createNote(final NotesViewEntry newNote) {
                 selectionLabelText = "Note linked to: the graph.";
             } else {
                 selectionLabelText = "Note linked to: ";
-                if (newNote.getNodesSelected().size() == 1) {
-                    selectionLabelText += newNote.getNodesSelected().size() + " node, ";
+                // Null-guard: selection lists can be null for some persisted/user notes.
+                final int nodeCount = newNote.getNodesSelected() != null ? newNote.getNodesSelected().size() : 0;
+                final int transactionCount = newNote.getTransactionsSelected() != null ? newNote.getTransactionsSelected().size() : 0;
```
antares1470: I think these null checks may not be necessary. Because these calls to `getNodesSelected` and `getTransactionsSelected` only occur if `newNote` is user created (if statement on line 841), these values should already have a non-null value (based on the instantiation in the `NotesViewEntry` constructor, and current calls to the related `setNodesSelected` and `setTransactionsSelected` functions don't appear to set to null unless it was already null).

If we wanted to apply a defensive check, you could ensure in those setter functions (in NotesViewEntry) that null can't be set for user created notes (similar to the constructor) but that might be the extent that is needed

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/NotesViewPane.java
```diff
@@ -1002,17 +1007,21 @@ private void createNote(final NotesViewEntry newNote) {
                 } else {
                     // Select the specific nodes and/or transactions applied to the note.
                     // Add nodes that are selected to the note.
-                    final int nodesLength = newNote.getNodesSelected().size();
                     final List<Integer> selectedNodes = newNote.getNodesSelected();
-                    for (int i = 0; i < nodesLength; i++) {
-                        elementIdsVx.set(selectedNodes.get(i));
+                    if (selectedNodes != null) {
```
antares1470: similar case as above for this one (and the equivalent transactions one below it). This is only called if the note is user created

antares1470 (changes requested):

arimu1:
@antares1470 Thanks — addressed on tip `0409ddec7`:

1. **Selection list null checks removed** in `NotesViewPane` for user-created note paths (selection label counts ~848, context-menu "Select on Graph" ~1011, menu disable guard, and `addToSelectedElements` / `removeFromSelectedElements`).
2. **Setter hardening** in `NotesViewEntry`: `setNodesSelected` / `setTransactionsSelected` now keep empty lists for user-created notes when null is passed (mirrors constructor behaviour).
3. **Test added**: `testUserCreatedSettersRejectNullSelectionLists` covers the setter hardening.

Re Unit Tests CI on the prior run: all TestNG suites reported `Failures: 0`; the job exit was from SonarCloud auth (`SONAR_TOKEN` / project binding), not from test failures. Happy to adjust further if anything else stands out.

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/state/NotesViewEntry.java
```diff
@@ -120,11 +120,19 @@ public List<Integer> getTransactionsSelected() {
     }
 
     public void setNodesSelected(final List<Integer> nodesSelected) {
-        this.nodesSelected = nodesSelected;
+        if (userCreated) {
+            this.nodesSelected = nodesSelected != null ? nodesSelected : new ArrayList<>();
+        } else {
+            this.nodesSelected = nodesSelected;
```
antares1470: this can be simplified to
`this.nodesSelected = userCreated && nodeSelected == null ? new ArrayList<>() : nodesSelected`

CoreNotesView/src/au/gov/asd/tac/constellation/views/notes/state/NotesViewEntry.java
```diff
@@ -120,11 +120,19 @@ public List<Integer> getTransactionsSelected() {
...
+        }
     }
 
     public void setTransactionsSelected(final List<Integer> transactionsSelected) {
-        this.transactionsSelected = transactionsSelected;
+        if (userCreated) {
+            this.transactionsSelected = transactionsSelected != null ? transactionsSelected : new ArrayList<>();
+        } else {
+            this.transactionsSelected = transactionsSelected;
+        }
```
antares1470: similar here

antares1470 (changes requested):
A couple more minor changes. I think it should be good to go after this

arimu1:
Thanks @antares1470 — addressed the setter simplification on tip `39db9fc9c` (`setNodesSelected` / `setTransactionsSelected` now use the ternary form you suggested).

Note on the red **Unit Tests** check: the TestNG suites themselves are green (CoreNotesView 22/0 failures in the last run). The job fails afterward on SonarCloud (`Not authorized or project not found` / missing `SONAR_TOKEN` for fork PRs). Happy to adjust further if anything else stands out.

antares1470 (approved):

PR #2572, opened 2026-09-18
Fix online help search

Delphinus8821:
### Prerequisites

### Description of the Change
Since the domain change, the search function in online help hasn't been working. 
This is a potential fix for the search but it cannot be tested until it has been merged. 

Fix for something in Core


search.json
```diff
@@ -2,972 +2,918 @@
     "id": 0,
     "title": "About Constellation",
     "category": "Overview",
-    "link": "ext\\docs\\CoreFunctionality\\about-constellation.html"
+    "link": "../ext\\docs\\CoreFunctionality\\about-constellation.html"
```
antares1470, line 5: I suspect this needs to be \\\ rather than /

Delphinus8821: I believe the \\\ is just for when it is run from offline help, and it still functions in that case. All of the links in online help are / so I'm not sure whether it matters

Delphinus8821: Actually I've just tested using ..\\\ and that does not work in offline help

antares1470 (approved):

